PR #3, opened 2015-11-30
Logging Upgrade

vjkoskela:
Upgrade to latest size and time based logging via Logback-Steno; removed local flyway-play library


BrandonArp:
LGTM, just waiting for build to complete

PR #5, opened 2015-12-17
Added telemetry proxy endpoint and actor

vjkoskela:


app/actors/ProxyConnection.java
```diff
@@ -0,0 +1,231 @@
...
+
+        @Override
+        public void onOpen(final ServerHandshake handshakedata) {
+            ProxyConnection.this.getSelf().tell(new DestinationConnected(), ProxyConnection.this.getSelf());
+        }
+
+        @Override
+        public void onMessage(final String message) {
+            // Unsure what the underlying threading model in WebSocketClient is; so play it safe
+            synchronized (_isProxied) {
```
BrandonArp: Definitely not the way to go here.  You should be firing messages to the ProxyConnection actor for it to write to the originator.

app/actors/ProxyConnection.java
```diff
@@ -0,0 +1,231 @@
...
+                    .addData("actor", self())
+                    .addData("destination", getURI())
+                    .setThrowable(ex)
+                    .log();
+            ProxyConnection.this.getSelf().tell(PoisonPill.getInstance(), ProxyConnection.this.getSelf());
+        }
+
+        public void establishProxy() {
+            // Unsure what the underlying threading model in WebSocketClient is; so play it safe
+            synchronized (ProxyConnection.this._isProxied) {
```
BrandonArp: Seems like this should also go to the actor.

app/assets/javascripts/classes/ConnectionModel.ts
```diff
@@ -66,20 +66,26 @@ class ConnectionModel {
...
-            {path: "/stream", protocol: new V1Protocol(this)}
-        ];
-    }
-
-    buildWebSocket() {
         var serverNameComponents = this.server.split(":");
         var serverHost = serverNameComponents[0];
         var serverPort = (typeof serverNameComponents[1] === "undefined") ? "7090" : serverNameComponents[1];
 
+        var directRoutePrefix = "ws://" + serverHost + ":" + serverPort;
```
BrandonArp: Since you're in here, you should match the protocol of the window.location so that we won't get ssl vs non-ssl errors.  Chrome is quite strict about that these days.

app/actors/ProxyConnection.java
vjkoskela: Modified to loop-back through the actor.

app/actors/ProxyConnection.java
vjkoskela: By implication of the above, with the queue in the actor the drain on establish is also in the actor now.

BrandonArp:
Much better, thanks.

PR #6, opened 2015-12-29
SQL implementation of expression and alert repositories

misradeepika:


app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,376 @@
...
+    /**
+     * Public constructor.
+     *
+     * @param environment Play's <code>Environment</code> instance.
+     * @param config      Play's <code>Configuration</code> instance.
+     * @throws Exception If the configuration is invalid.
+     */
+    @Inject
+    public DatabaseAlertRepository(final Environment environment, final Configuration config) throws Exception {
+        this(
```
BrandonArp: Why is the query generator being constructed like this.  What cases do you foresee us needing to specialize on through configuration?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,376 @@
...
+                .setMessage("Querying")
+                .addData("query", query)
+                .log();
+
+        // Create the base query
+        final PagedList<models.ebean.Alert> pagedAlerts = _alertQueryGenerator.createAlertQuery(query);
+
+        // Compute the etag
+        // NOTE: Another way to do this would be to use the version field and hash those together.
+        final String etag = Long.toHexString(pagedAlerts.getList().stream()
```
BrandonArp: So, this isn't going to work like you think it is.  Unfortunately, the place that you took this from is also broken.  The gist of the problem is that the etag needs to represent the entire state of the resource, in this case that's the whole table.  Lets talk about alternatives.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,376 @@
...
+        /**
+         * Save the <code>Alert</code> to the database. This needs to be executed in a transaction.
+         *
+         * @param alert The <code>Alert</code> model instance to save.
+         */
+        void saveAlert(models.ebean.Alert alert);
+    }
+
+    /**
+     * RDBMS agnostic query for alerts using 'like'.
```
BrandonArp: I don't think you're using 'like' here.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,376 @@
...
+            ebeanAlert.setContext(alert.getContext());
+            ebeanAlert.setNagiosExtension(getNagiosExtensionFromMap(alert.getExtensions()));
+            ebeanAlert.setName(alert.getName());
+            ebeanAlert.setOperator(alert.getOperator());
+            ebeanAlert.setPeriod(alert.getPeriod().toStandardSeconds().getSeconds());
+            ebeanAlert.setQuantityValue(alert.getValue().getValue());
+            ebeanAlert.setQuantityUnit(alert.getValue().getUnit().isPresent() ? alert.getValue().getUnit().get() : null);
+            ebeanAlert.setStatistic(alert.getStatistic());
+            ebeanAlert.setService(alert.getService());
+            _alertQueryGenerator.saveAlert(ebeanAlert);
```
BrandonArp: What happens if this fails?  There is a race condition here.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,376 @@
...
+        } finally {
+            transaction.end();
+        }
+    }
+
+    //NOTE: Only to be used to clear the tables for testing
+    /**
+     * Package private.
+     */
+    void deleteAll() {
```
BrandonArp: I really don't think this is a good idea.  Even if it's for testing, use of this function means that tests can't run in parallel.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,376 @@
...
+    public void addOrUpdateAlert(final Alert alert) {
+        assertIsOpen();
+        LOGGER.debug()
+                .setMessage("Upserting alert")
+                .addData("alert", alert)
+                .log();
+
+        final Transaction transaction = Ebean.beginTransaction();
+
+        try {
```
BrandonArp: Make the transaction a try-with-resources.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    private NagiosExtension getNagiosExtensionFromMap(final Map<String, Object> extensionsMap) {
+        if (extensionsMap == null || extensionsMap.isEmpty()) {
+            return null;
+        }
+        final NagiosExtension nagiosExtension = new NagiosExtension();
+        nagiosExtension.setSeverity((String) extensionsMap.get("severity"));
+        nagiosExtension.setNotify((String) extensionsMap.get("notify"));
+        nagiosExtension.setMaxCheckAttempts((Integer) extensionsMap.get("maxCheckAttempts"));
```
BrandonArp: Use the constants here.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,376 @@
...
+        return extensionMap;
+    }
+
+    private final AtomicBoolean _isOpen = new AtomicBoolean(false);
+    private final AlertQueryGenerator _alertQueryGenerator;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DatabaseAlertRepository.class);
+    private static final String NAGIOS_EXTENSION_SEVERITY_KEY = "severity";
+    private static final String NAGIOS_EXTENSION_NOTIFY_KEY = "notify";
+    private static final String NAGIOS_EXTENSION_MAX_CHECK_ATTEMPTS_KEY = "maxCheckAttempts";
```
BrandonArp: I don't think this is the right constant.

app/com/arpnetworking/metrics/portal/alerts/impl/ReadOnlyAbstractAlertRepository.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Deepika Misra (deepika at groupon dot com)
+ */
+public abstract class ReadOnlyAbstractAlertRepository implements AlertRepository {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void addOrUpdateAlert(final Alert alert) {
+        LOGGER.info()
```
BrandonArp: We're probably going to want this to throw some sort of exception so the UI can display that some notification that the write didn't happen.

app/com/arpnetworking/metrics/portal/expressions/impl/ReadOnlyAbstractExpressionRepository.java
```diff
@@ -0,0 +1,41 @@
...
+ * @author Deepika Misra (deepika at groupon dot com)
+ */
+public abstract class ReadOnlyAbstractExpressionRepository implements ExpressionRepository {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void addOrUpdateExpression(final Expression expression) {
+        LOGGER.info()
```
BrandonArp: Same as above, lets throw an exception here.

project/Build.scala
```diff
@@ -276,6 +276,8 @@ object ApplicationBuild extends Build {
             <Class name="~controllers\.Reverse.*"/>
           </Match>
         </FindBugsFilter>
-      )
+      ),
+
+      javaOptions in Test += "-Dconfig.file=conf/portal.application.conf"
```
BrandonArp: Can you please move this up by the devSettings := line?

app/com/arpnetworking/metrics/portal/alerts/impl/ReadOnlyAbstractAlertRepository.java
misradeepika: Done

app/com/arpnetworking/metrics/portal/expressions/impl/ReadOnlyAbstractExpressionRepository.java
misradeepika: Done

project/Build.scala
misradeepika: Done

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
misradeepika: not right in what sense? Is the key name incorrect or the constant name?

BrandonArp: The value of the string is incorrect.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
misradeepika: Done

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
misradeepika: For similar reasons as in the host repository. That is, if we ever use a special kind of query like the full text search capability. I dont think we need the full search abilities on alerts but building the query generator this way at least allows the flexibility for that

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
misradeepika: Ebean automatically does an optimistic locking for updates. And now I am also throwing an exception when a query fails for any reason, so the user could be made aware that one of their queries failed

app/com/arpnetworking/database/h2/triggers/AlertsUpdateEtagTrigger.java
```diff
@@ -0,0 +1,67 @@
...
+            final String tableName,
+            final boolean before,
+            final int type) throws SQLException {}
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void fire(final Connection conn, final Object[] oldRow, final Object[] newRow) throws SQLException {
+        final PreparedStatement stmt = conn.prepareStatement("SELECT NEXTVAL('portal.alerts_etag_seq');");
```
BrandonArp: Try with resources.

app/com/arpnetworking/database/h2/triggers/ExpressionsUpdateEtagTrigger.java
```diff
@@ -0,0 +1,68 @@
...
+                     final boolean before,
+                     final int type) throws SQLException {
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void fire(final Connection conn, final Object[] oldRow, final Object[] newRow) throws SQLException {
+        final PreparedStatement stmt = conn.prepareStatement("SELECT NEXTVAL('portal.expressions_etag_seq');");
```
BrandonArp: Try with resources.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,364 @@
...
+            LOGGER.info()
+                    .setMessage("Upserted alert")
+                    .addData("alert", alert)
+                    .addData("isCreated", created)
+                    .log();
+        } catch (final IOException e) {
+            LOGGER.error()
+                    .setMessage("Failed to close the transaction")
+                    .log();
+            //CHECKSTYLE.OFF: IllegalCatchCheck
```
BrandonArp: Shouldn't you also throw the PersistenceException here?

app/com/arpnetworking/metrics/portal/expressions/impl/DatabaseExpressionRepository.java
```diff
@@ -0,0 +1,302 @@
...
+            LOGGER.info()
+                    .setMessage("Upserted expression")
+                    .addData("expression", expression)
+                    .addData("isCreated", created)
+                    .log();
+        } catch (final IOException e) {
+            LOGGER.error()
+                    .setMessage("Failed to close the transaction")
+                    .log();
+            //CHECKSTYLE.OFF: IllegalCatchCheck
```
BrandonArp: Shouldn't you throw a PersistenceException here?

BrandonArp:
Make sure you test this with postgres too.

misradeepika:
Yeah, I tested the sqls for postgres

app/com/arpnetworking/database/h2/triggers/AlertsUpdateEtagTrigger.java
```diff
@@ -0,0 +1,67 @@
...
+            final String tableName,
+            final boolean before,
+            final int type) throws SQLException {}
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void fire(final Connection conn, final Object[] oldRow, final Object[] newRow) throws SQLException {
+        try (PreparedStatement stmt = conn.prepareStatement("SELECT NEXTVAL('portal.alerts_etag_seq');")) {
```
vjkoskela: Could this be generalized to any sequence by accepting the sequence name in the constructor?

app/com/arpnetworking/database/h2/triggers/AlertsUpdateEtagTrigger.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @author Deepika Misra (deepika at groupon dot com)
+ */
+public class AlertsUpdateEtagTrigger implements Trigger {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void init(final Connection conn,
```
vjkoskela: If you wrap one, then wrap all.

app/com/arpnetworking/database/h2/triggers/ExpressionsUpdateEtagTrigger.java
```diff
@@ -0,0 +1,68 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+/**
+ * Trigger to update Etag after every insert, delete or update statement.
+ *
+ * @author Deepika Misra (deepika at groupon dot com)
+ */
+public class ExpressionsUpdateEtagTrigger implements Trigger {
```
vjkoskela: Redundant class; see comments on AlertsUpdateEtagTrigger.java

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,365 @@
...
+        LOGGER.debug()
+                .setMessage("Querying")
+                .addData("query", query)
+                .log();
+
+        // Create the base query
+        final PagedList<models.ebean.Alert> pagedAlerts = _alertQueryGenerator.createAlertQuery(query);
+
+        // Compute the etag
+        final SqlQuery sqlQuery = Ebean.createSqlQuery("SELECT CURRVAL('portal.alerts_etag_seq') AS etag;");
```
vjkoskela: Isn't this the type of things that should be in the query generator?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,365 @@
...
+        final Long etag = sqlQuery.findUnique().getLong("etag");
+
+        final List<Alert> values = new ArrayList<>();
+        pagedAlerts.getList().forEach(ebeanAlert -> {
```
vjkoskela: This building code is duplicated from above. Perhaps using a private conversion method and function reference would be the way to go here.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,365 @@
...
+                            .setValue(ebeanAlert.getQuantityValue())
+                            .setUnit(ebeanAlert.getQuantityUnit())
+                            .build())
+                    .setExtensions(mergeExtensions(ebeanAlert.getNagiosExtension()))
+                    .build();
+            values.add(result);
+        });
+
+        // Transform the results
+        return new DefaultQueryResult<>(values, pagedAlerts.getTotalRowCount(), etag.toString());
```
vjkoskela: Can you file an issue against the project that we should obfuscate the etag value instead of returning the table version in plain? Please add a TODO here in the standard style and reference the issue (e.g. TODO(deepika): <desc> [ISSUE-?])

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,365 @@
...
+            ebeanAlert.setCluster(alert.getCluster());
+            ebeanAlert.setUuid(alert.getId());
+            ebeanAlert.setMetric(alert.getMetric());
+            ebeanAlert.setContext(alert.getContext());
+            ebeanAlert.setNagiosExtension(getNagiosExtensionFromMap(alert.getExtensions()));
+            ebeanAlert.setName(alert.getName());
+            ebeanAlert.setOperator(alert.getOperator());
+            ebeanAlert.setPeriod(alert.getPeriod().toStandardSeconds().getSeconds());
+            ebeanAlert.setQuantityValue(alert.getValue().getValue());
+            ebeanAlert.setQuantityUnit(alert.getValue().getUnit().isPresent() ? alert.getValue().getUnit().get() : null);
```
vjkoskela: Isn't there a getOrNull on the optional?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,365 @@
...
+
+            LOGGER.info()
+                    .setMessage("Upserted alert")
+                    .addData("alert", alert)
+                    .addData("isCreated", created)
+                    .log();
+        } catch (final IOException e) {
+            LOGGER.error()
+                    .setMessage("Failed to close the transaction")
+                    .log();
```
vjkoskela: Why is the exception not logged?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,365 @@
...
+            throw new PersistenceException(e);
+            //CHECKSTYLE.OFF: IllegalCatchCheck
+        } catch (final RuntimeException e) {
+            //CHECKSTYLE.ON
+            LOGGER.error()
+                    .setMessage("Failed to upsert alert")
+                    .addData("alert", alert)
+                    .log();
```
vjkoskela: Same here?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,365 @@
...
+            transaction.commit();
+
+            LOGGER.info()
+                    .setMessage("Upserted alert")
+                    .addData("alert", alert)
+                    .addData("isCreated", created)
+                    .log();
+        } catch (final IOException e) {
+            LOGGER.error()
+                    .setMessage("Failed to close the transaction")
```
vjkoskela: What does this mean? Is this the only possible cause?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,365 @@
...
+        }
+        final NagiosExtension nagiosExtension = new NagiosExtension();
+        nagiosExtension.setSeverity((String) extensionsMap.get(NAGIOS_EXTENSION_SEVERITY_KEY));
+        nagiosExtension.setNotify((String) extensionsMap.get(NAGIOS_EXTENSION_NOTIFY_KEY));
+        nagiosExtension.setMaxCheckAttempts((Integer) extensionsMap.get(NAGIOS_EXTENSION_MAX_CHECK_ATTEMPTS_KEY));
+        nagiosExtension.setFreshnessThreshold((Integer) extensionsMap.get(NAGIOS_EXTENSION_FRESHNESS_THRESHOLD_KEY));
+        return nagiosExtension;
+    }
+
+    private Map<String, Object> mergeExtensions(final NagiosExtension nagiosExtension) {
```
vjkoskela: So to be clear this obfuscation of the extensions into a map should happen at the view layer. If this is a temporary state to ensure backwards compatibility that's fine. The existing v1 endpoints will continue to return a map data structure for extensions while you should also write v2 endpoints which return structured extensions (e.g. NagiosExtensions with named fields). The latter will facilitate construction of the user interface while the former retains compatibility with TsdAggregator/ClusterAggregation and Nagios which query this endpoint.

app/com/arpnetworking/metrics/portal/alerts/impl/ReadOnlyAbstractAlertRepository.java
```diff
@@ -0,0 +1,42 @@
...
+ * @author Deepika Misra (deepika at groupon dot com)
+ */
+public abstract class ReadOnlyAbstractAlertRepository implements AlertRepository {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void addOrUpdateAlert(final Alert alert) {
+        LOGGER.error()
```
vjkoskela: You shouldn't log and throw.

app/com/arpnetworking/metrics/portal/expressions/impl/DatabaseExpressionRepository.java
```diff
@@ -0,0 +1,303 @@
...
+ *
+ * @author Deepika Misra (deepika at groupon dot com)
+ */
+public class DatabaseExpressionRepository implements ExpressionRepository {
+
+    /**
+     * Public constructor.
+     *
+     * @param environment Play's <code>Environment</code> instance.
+     * @param config      Play's <code>Configuration</code> instance.
```
vjkoskela: no aligned indent.

app/com/arpnetworking/metrics/portal/expressions/impl/DatabaseExpressionRepository.java
```diff
@@ -0,0 +1,303 @@
...
+        LOGGER.debug()
+                .setMessage("Querying")
+                .addData("query", query)
+                .log();
+
+        // Create the base query
+        final PagedList<models.ebean.Expression> pagedExpressions = _expressionQueryGenerator.createExpressionQuery(query);
+
+        // Compute the etag
+        final SqlQuery sqlQuery = Ebean.createSqlQuery("SELECT CURRVAL('portal.expressions_etag_seq') AS etag;");
```
vjkoskela: Similar comments here as alerts db repository:
 - etag fetch to query generator
 - ebean to model conversion function
 - etag obfuscation
 - exceptions in error messages
 - exception handling

app/com/arpnetworking/metrics/portal/expressions/impl/ReadOnlyAbstractExpressionRepository.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void addOrUpdateExpression(final Expression expression) {
+        LOGGER.error()
+                .setMessage("This is a read only repository.")
+                .log();
+        throw new UnsupportedOperationException("This is a read only repository.");
```
vjkoskela: Exception or log.

conf/db/migration/metrics_portal_ddl/h2/V4__create_etag_in_alerts_and_expressions.sql
```diff
@@ -0,0 +1,15 @@
+-- Alerts etag --
+CREATE SEQUENCE portal.alerts_etag_seq START WITH 1 INCREMENT BY 1 NO MINVALUE NO MAXVALUE;
+-- Trigger to be executed after each insert, update and delete statement --
+CREATE TRIGGER update_alerts_etag
+AFTER INSERT, DELETE, UPDATE ON portal.alerts
+FOR EACH ROW
+CALL "com.arpnetworking.database.h2.triggers.AlertsUpdateEtagTrigger";
```
vjkoskela: I see. I assume this is what prevents parameterization of the trigger with the sequence name? We could still refactor the code to have one logical instance and two stub subclasses.

test/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepositoryTest.java
```diff
@@ -0,0 +1,313 @@
...
+/**
+ * Tests class <code>DatabaseAlertRepository</code>.
+ *
+ * @author Deepika Misra (deepika at groupon dot com)
+ */
+public class DatabaseAlertRepositoryTest extends WithApplication {
+
+    @Override
+    protected FakeApplication provideFakeApplication() {
+        final int port = BASE_PORT + RANDOM.nextInt(1000);
```
vjkoskela: No. Random ports will be a disaster. When it does eventually collide you won't know it and people will just rerun. My recommendation, and we did it in PCE, was using global increasing port numbers. I would recommend a test helper class that maintained this state and generated the JDBC urls (for example).

test/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepositoryTest.java
```diff
@@ -0,0 +1,313 @@
...
+        Assert.assertEquals(TEST_CLUSTER, expected.get().getCluster());
+        Assert.assertEquals(TEST_METRIC, expected.get().getMetric());
+        Assert.assertEquals(TEST_EXTENSIONS, expected.get().getExtensions());
+        Assert.assertEquals(TEST_SERVICE, expected.get().getService());
+        Assert.assertEquals(TEST_NAME, expected.get().getName());
+        Assert.assertEquals(TEST_OPERATOR, expected.get().getOperator());
+        Assert.assertEquals(Period.seconds(TEST_PERIOD_IN_SECONDS), expected.get().getPeriod());
+        Assert.assertEquals(TEST_STATISTIC, expected.get().getStatistic());
+        Assert.assertEquals(TEST_QUANTITY, expected.get().getValue());
+        Assert.assertEquals(TEST_CONTEXT, expected.get().getContext());
```
vjkoskela: So you should probably use instances instead of per field constants and have an externalized equality method. External because if you write an equals on the bean then that will be used by the orm and I am not sure we want to go there (at least not yet).

test/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepositoryTest.java
```diff
@@ -0,0 +1,313 @@
...
+        Assert.assertEquals(0, alertRepo.getAlertCount());
+    }
+
+    @Test
+    public void testGetAlertCountWithMultipleAlert() throws JsonProcessingException {
+        Assert.assertEquals(0, alertRepo.getAlertCount());
+        final Transaction transaction = Ebean.beginTransaction();
+        try {
+            final models.ebean.Alert ebeanAlert1 = new models.ebean.Alert();
+            ebeanAlert1.setNagiosExtension(TEST_NAGIOS_EXTENSIONS);
```
vjkoskela: If you take a look at the some of the patterns we've used for creating instances for test we usually provide a TestBeanFactory which allows creation of instances with default random values in the fields. Tests should then set the fields they care about.

test/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepositoryTest.java
```diff
@@ -0,0 +1,313 @@
...
+        Assert.assertEquals(TEST_NAME, expected.get().getName());
+        Assert.assertEquals(TEST_OPERATOR, expected.get().getOperator());
+        Assert.assertEquals(Period.seconds(TEST_PERIOD_IN_SECONDS), expected.get().getPeriod());
+        Assert.assertEquals(TEST_STATISTIC, expected.get().getStatistic());
+        Assert.assertEquals(TEST_QUANTITY, expected.get().getValue());
+        Assert.assertEquals(TEST_CONTEXT, expected.get().getContext());
+    }
+
+    @Test
+    public void addAlertWithNoExtension() {
```
vjkoskela: Prefix method with test

test/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepositoryTest.java
```diff
@@ -0,0 +1,313 @@
...
+            .setValue(TEST_QUANTITY_VALUE)
+            .setUnit(TEST_QUANTITY_UNIT)
+            .build();
+    private static final NagiosExtension TEST_NAGIOS_EXTENSIONS = new NagiosExtension();
+    private static final int BASE_PORT = 50000;
+    private static final Random RANDOM = new Random();
+
+    static {
+        TEST_NAGIOS_EXTENSIONS.setSeverity("CRITICAL");
+        TEST_NAGIOS_EXTENSIONS.setNotify("abc@groupon.com");
```
vjkoskela: You shouldn't use a groupon address here. Please use example.com as the domain (for example).

test/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepositoryTest.java
```diff
@@ -0,0 +1,313 @@
...
+            .setStatistic(TEST_STATISTIC)
+            .setValue(TEST_QUANTITY);
+
+    private static final String TEST_CLUSTER = "test-cluster";
+    private static final String TEST_METRIC = "test-metric";
+    private static final String TEST_SERVICE = "test-service";
+    private static final Context TEST_CONTEXT = Context.CLUSTER;
+    private static final Map<String, Object> TEST_EXTENSIONS = ImmutableMap.of(
+            "severity", "CRITICAL",
+            "notify", "abc@groupon.com",
```
vjkoskela: You shouldn't use a groupon address here. Please use example.com as the domain (for example).

test/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepositoryTest.java
```diff
@@ -0,0 +1,313 @@
...
+            "max_check_attempts", 3,
+            "freshness_threshold", 5);
+    private static final String TEST_NAME = "test-name";
+    private static final Operator TEST_OPERATOR = Operator.EQUAL_TO;
+    private static final int TEST_PERIOD_IN_SECONDS = 600;
+    private static final String TEST_STATISTIC = "test-statistic";
```
vjkoskela: This is not a valid statistic name.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
misradeepika: The IOException here is a checked exception thrown by the autocloseable transaction object. So in this case, we know that the commit went through. We dont know if the commits were made in case of the other exception, like in case of an exception after the transaction.commit() statement. If the caller knows whether or not the queries went through, it will help show appropriate error messages to user. Right?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
misradeepika: I will create an issue for this for now and add a todo

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
vjkoskela: Are you sure that failing to close the transaction means that the commit was successful?

In general since the catch blocks are identical the only difference is in the stack traces which should reflect the difference you describe. Otherwise, it sounds like you are telling me that one exception should be propagated because its a failure and the other is not a failure - then why is it propagated and moreover why is it even an exception? -- my point is not to not propagate or not catch, but evidence that your assumption may not be correct.

misradeepika: Its late in the night, i will think about it tomorrow :)

app/com/arpnetworking/database/h2/triggers/AlertsUpdateEtagTrigger.java
```diff
@@ -0,0 +1,35 @@
...
+ * @author Deepika Misra (deepika at groupon dot com)
+ */
+public class AlertsUpdateEtagTrigger extends BaseUpdateEtagTrigger {
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void fire(final Connection conn, final Object[] oldRow, final Object[] newRow) throws SQLException {
+        baseFire(conn, oldRow, newRow, "portal.alerts_etag_seq");
```
vjkoskela: Why pass a constant to baseFire? Instead pass the constant from the AlertsUpdateEtagTrigger no-args constructor to super.

app/com/arpnetworking/database/h2/triggers/BaseUpdateEtagTrigger.java
```diff
@@ -0,0 +1,76 @@
...
+     * @param newRow the new row, or null if no new row is available (for DELETE)
+     * @param seqName The name of the sequence that is to be updated.
+     * @throws SQLException if the operation must be undone
+     */
+    public void baseFire(
+            final Connection conn,
+            final Object[] oldRow,
+            final Object[] newRow,
+            final String seqName) throws SQLException {
+        try (PreparedStatement stmt = conn.prepareStatement(String.format("SELECT NEXTVAL('%s');", seqName))) {
```
vjkoskela: By passing the argument you can format the statement on instantiation saving a string format call per invocation.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,356 @@
...
+        LOGGER.debug()
+                .setMessage("Querying")
+                .addData("query", query)
+                .log();
+
+        // Create the base query
+        final PagedList<models.ebean.Alert> pagedAlerts = _alertQueryGenerator.createAlertQuery(query);
+
+        // Compute the etag
+        // TODO(deepika)[ISSUE-7]: Obfuscate the etag
```
vjkoskela: Format is:

TODO(<USER>): Obfuscate the etag [ISSUE-<#>]

e.g.
TODO(deepika): Obfuscate the etag [ISSUE-7]

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -0,0 +1,356 @@
...
+                ebeanExpressionList = ebeanExpressionList.eq("cluster", query.getCluster().get());
+            }
+            if (query.getContext().isPresent()) {
+                ebeanExpressionList = ebeanExpressionList.eq("context", query.getContext().get().toString());
+            }
+            if (query.getService().isPresent()) {
+                ebeanExpressionList = ebeanExpressionList.eq("service", query.getService().get());
+            }
+
+            final Query<models.ebean.Alert> ebeanQuery = ebeanExpressionList.query();
```
vjkoskela: So the existing AlertQuery query supports a "contains" clause. I don't see support for that here. To prevent this from getting out of hand (you'll need to add a full text index and the corresponding query logic) please add an issue here and in Github and do the work as a fast-follow.

app/com/arpnetworking/metrics/portal/expressions/impl/DatabaseExpressionRepository.java
```diff
@@ -0,0 +1,312 @@
...
+        LOGGER.debug()
+                .setMessage("Querying")
+                .addData("query", query)
+                .log();
+
+        // Create the base query
+        final PagedList<models.ebean.Expression> pagedExpressions = _expressionQueryGenerator.createExpressionQuery(query);
+
+        // Compute the etag
+        // TODO(deepika)[ISSUE-7]: Obfuscate the etag
```
vjkoskela: As noted above.

app/com/arpnetworking/metrics/portal/expressions/impl/DatabaseExpressionRepository.java
```diff
@@ -0,0 +1,312 @@
...
+                ebeanExpressionList = ebeanExpressionList.eq("service", query.getService().get().toString());
+            }
+
+            final Query<models.ebean.Expression> ebeanQuery = ebeanExpressionList.query();
+
+            int pageOffset = 0;
+            if (query.getOffset().isPresent()) {
+                pageOffset = query.getOffset().get() / query.getLimit();
+            }
+            return ebeanQuery.findPagedList(pageOffset, query.getLimit());
```
vjkoskela: Same point about contains.

app/models/internal/NagiosExtension.java
```diff
@@ -0,0 +1,175 @@
...
+
+    public String getNotify() {
+        return _notify;
+    }
+
+    public int getMaxCheckAttempts() {
+        return _maxCheckAttempts;
+    }
+
+    public int getFreshnessThresholdInSeconds() {
```
vjkoskela: On the internal model this should be a Duration.

app/models/internal/NagiosExtension.java
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        @NotNull
+        @NotEmpty
+        private String _severity;
+        @NotNull
+        @NotEmpty
+        private String _notify;
+        @NotNull
+        private Integer _maxCheckAttempts;
```
vjkoskela: This should be at least one - right?

app/models/internal/NagiosExtension.java
```diff
@@ -0,0 +1,175 @@
...
+        @NotNull
+        private Integer _freshnessThresholdInSeconds;
```
vjkoskela: This should be at least zero - right?

test/com/arpnetworking/metrics/portal/TestBeanFactory.java
```diff
@@ -0,0 +1,134 @@
...
+                .setMetric(TEST_METRIC)
+                .setScript(TEST_SCRIPT)
+                .setService(TEST_SERVICE);
+    }
+
+    public static models.internal.Expression createExpression() {
+        return createExpressionBuilder().build();
+    }
+
+    private static final String TEST_CLUSTER = "test-cluster";
```
vjkoskela: All this data should be randomly generated. The point is the default values should _never_ be relied on by a test. That is, if a test wants to assert something about the service name for example it should set it to a specific value. The bean factory here is not intended to create well-known beans because: 1) having more tests eventually diverge the requirements for the bean and thus reduce its effectiveness; 2) it hides the test's preconditions/dependencies by hiding the specific data values in use. Happy to discuss this further.

test/com/arpnetworking/metrics/portal/TestUtil.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Utility class for tests.
+ *
+ * @author Deepika Misra (deepika at groupon dot com)
+ */
+public final class TestUtil {
```
vjkoskela: No. This is far to generic. Please split the functionality up into more appropriate classes.

test/com/arpnetworking/metrics/portal/TestUtil.java
```diff
@@ -0,0 +1,85 @@
...
+        final int nextCounter = UNIQUE_COUNTER.getAndIncrement();
+        final int port = BASE_PORT + nextCounter;
+        return "jdbc:h2:./target/h2/metrics" +
+                nextCounter +
+                ":portal;AUTO_SERVER=TRUE;AUTO_SERVER_PORT=" +
+                port +
+                ";MODE=PostgreSQL;INIT=create schema if not exists portal;DB_CLOSE_DELAY=-1";
+    }
+
+    public static boolean isAlertEbeanEquivalent(final Alert alert, final models.ebean.Alert ebeanAlert) {
```
vjkoskela: This is only used in one test file. Why does it need to be in a utility class? (remember: 1-2-refactor)

test/com/arpnetworking/metrics/portal/TestUtil.java
```diff
@@ -0,0 +1,85 @@
...
+        return Objects.equals(expression.getId(), ebeanExpression.getUuid())
+                && Objects.equals(expression.getCluster(), ebeanExpression.getCluster())
+                && Objects.equals(expression.getMetric(), ebeanExpression.getMetric())
+                && Objects.equals(expression.getScript(), ebeanExpression.getScript())
+                && Objects.equals(expression.getService(), ebeanExpression.getService());
+    }
+
+    private static final AtomicInteger UNIQUE_COUNTER = new AtomicInteger(1);
+    private static final int BASE_PORT = 50000;
+}
```
vjkoskela: So it looks like the only method re-used across tests in here is the jdbc connection. So you just need an H2ConnectionStringFactory (or some such).

test/com/arpnetworking/metrics/portal/expressions/impl/DatabaseExpressionRepositoryTest.java
```diff
@@ -0,0 +1,166 @@
...
+        try (Transaction transaction = Ebean.beginTransaction()) {
+            ebeanExpression1.setCluster("new-cluster1");
+            ebeanExpression2.setCluster("new-cluster2");
+            ebeanExpression2.save();
+            ebeanExpression1.save();
+            transaction.commit();
+        } catch (final IOException e) {
+            //Do Nothing
+        }
+    }
```
vjkoskela: I don't see any test cases for the actual query of alerts or expressions which is the most complex part of the code.

app/controllers/AlertController.java
```diff
@@ -176,6 +178,18 @@ public Result query(
         return viewAlert;
     }
 
+    private Map<String, Object> mergeExtensions(final NagiosExtension nagiosExtension) {
+        final Map<String, Object> extensionMap = Maps.newHashMap();
+        if (nagiosExtension == null) {
```
vjkoskela: You should do this check before creating the extensionMap

app/models/internal/NagiosExtension.java
```diff
@@ -0,0 +1,179 @@
...
+import net.sf.oval.constraint.NotNull;
+import org.joda.time.Duration;
+
+import java.util.Objects;
+
+/**
+ * Represents the nagios specific data for an alert.
+ *
+ * @author Deepika Misra (deepika at groupon dot com)
+ */
```
vjkoskela: Should be annotated @Loggable.

vjkoskela:
Just minor feedback. Otherwise looks good. :shipit: when done.

BrandonArp:
Is this good to merge?

PR #13, opened 2016-02-23
Quality of life improvements for users

PhoenixRion:
Move frame rate logic out of individual graph render flow.
Add render update while scrolling.
Allow scrolling of graph while paused.
Buffer data during pause.
Do not advance graph window during pause.


app/assets/javascripts/classes/GraphVM.ts
```diff
@@ -43,9 +43,11 @@ class GraphVM implements StatisticView {
     started: boolean = false;
     container: HTMLElement = null;
     data: Series[] = [];
+    buffer: number[][][] = [];
```
vjkoskela: Series may not be the right object, in fact there may not be one right now at all, but using raw number here may create problems in the future. Another option why not put the buffer into data and and make methods to modify the pause/unpaused state of the Series.

app/assets/javascripts/classes/GraphVM.ts
```diff
@@ -106,151 +108,140 @@ class GraphVM implements StatisticView {
...
+        var graphMax = -1000000000;
 
-            if (this.stop) {
-                return;
-            }
+        var now = new Date().getTime() - 1000;
+        if (this.paused) {
+            now = this.pauseTime
+        } else {
+            this.pauseTime = now;
```
vjkoskela: You should add a pause and unpause methods to this class and modify the pause boolean and pauseTime in that method.

vjkoskela:
Great work Ryan! It looks good to me, Brandon is still looking at it.

app/assets/javascripts/classes/GraphVM.ts
```diff
@@ -108,149 +109,129 @@ class GraphVM implements StatisticView {
             this.data.push(series);
         }
 
+        this.data[index].pushData(timestamp, dataValue, this.paused);
+    }
 
-        if (this.data[index].data.length == 0 || this.data[index].data[this.data[index].data.length - 1][0] < timestamp) {
-            this.data[index].data.push([timestamp, dataValue]);
-        }
+    togglePause() {
```
BrandonArp: What happens if you add some graphs, then pause them, then add a new graph?

BrandonArp:
Just the one comment from me.  If that works, then we're good and I'll merge it.

app/assets/javascripts/classes/GraphVM.ts
PhoenixRion: They get stuck out of sync with each other

public/stylesheets/graph-config.css
```diff
@@ -0,0 +1,14 @@
...
+    position: absolute;
+    top: 11%;
+    right: 9%;
+    width: 80%;
+    height: 70%;
+    background-color: rgba(255, 218, 185, 0.8);;
+    color: steelblue;
+}
+
+input[type="checkbox"] {
```
BrandonArp: I think this is going to kill all textboxes in the app.

vjkoskela:
Please add yourself as a contributor to DEPENDENCIES.md and to the generated pom in Build.scala.

PhoenixRion:
I don't see a section in the DEPENDENCIES.md file for contributors.

BrandonArp:
I think Ville meant "CONTRIBUTORS.md"

vjkoskela:
Sorry about that @PhoenixRion, @BrandonArp has it right.

PhoenixRion:
I probably could have looked harder

app/assets/javascripts/classes/GraphSpec.ts
```diff
@@ -18,11 +18,17 @@ class GraphSpec {
     service: string;
     metric: string;
     statistic: string;
+    points: any;
+    lines: any;
+    bars: any;
 
-    constructor(service: string, metric: string, statistic: string) {
+    constructor(service: string, metric: string, statistic: string, points: any, lines: any, bars: any) {
```
BrandonArp: Why are all of these "any" type?  That seems very bad in this case.  Also, there are all display parameters.  I'm not sure GraphSpec is the right place for them, but I'll think more on that.

app/assets/javascripts/classes/StatisticView.ts
```diff
@@ -22,14 +22,19 @@ interface StatisticView {
...
+    render(): void;
+    start(paused: boolean): void;
+    setPause(pause: boolean): void;
     postData(server: string, timestamp: number, dataValue: number, cvm: ConnectionVM): void;
     shutdown(): void;
     setViewDuration(duration: ViewDuration): void;
     updateColor(cvm: ConnectionVM): void;
     disconnectConnection(cvm: ConnectionVM): void;
+
+    renderDots: KnockoutObservable<boolean>;
```
BrandonArp: This certainly shouldn't be in StatisticView.  It doesn't generalize to other types of visualization.

BrandonArp:
The front-end works really nice.  I'm a huge fan!

app/assets/javascripts/classes/GraphSpec.ts
PhoenixRion: They are that type because that is what they are in Series.ts

app/assets/javascripts/classes/StatisticView.ts
PhoenixRion: The corresponding values in the Series object don't generalize either.  There is definitely looks like there is work to be done to properly support multiple types of visualization.  For the scope of this PR I think having these values here follows the existing patterns.

vjkoskela:
@BrandonArp Is this ready to be merged?

PR #14, opened 2016-02-24
Controller methods for adding/updating alerts and expressions

misradeepika:


app/controllers/AlertController.java
```diff
@@ -61,6 +70,53 @@ public AlertController(final Configuration configuration, final AlertRepository
     }
 
     /**
+     * Adds or updates an alert in the alert repository.
+     * If a valid UUID is passed, it will update the corresponding alert if it exists or create an alert with this id.
```
vjkoskela: Hrm. Should we not distinguish this based on PUT (update) versus POST (create)? There is likely common logic between these two controller methods but the entry points seem separate. @BrandonArp thoughts?

app/controllers/AlertController.java
```diff
@@ -61,6 +70,53 @@ public AlertController(final Configuration configuration, final AlertRepository
...
+        } catch (final IllegalArgumentException e) {
+            return badRequest("Invalid alert id");
+        }
+        final JsonNode jsonBody = request().body().asJson();
+        if (jsonBody == null) {
+            return badRequest("Request must have a non-empty body");
+        }
+        Alert alert;
+        try {
+            final DefaultAlert.Builder alertBuilder = OBJECT_MAPPER.readValue(jsonBody.toString(), DefaultAlert.Builder.class);
```
vjkoskela: You shouldn't be deserializing directly from the view model into the internal model.

app/controllers/AlertController.java
```diff
@@ -220,4 +276,16 @@ private AlertController(final int maxLimit, final AlertRepository alertRepositor
     private static final String NAGIOS_EXTENSION_NOTIFY_KEY = "notify";
     private static final String NAGIOS_EXTENSION_MAX_CHECK_ATTEMPTS_KEY = "max_check_attempts";
     private static final String NAGIOS_EXTENSION_FRESHNESS_THRESHOLD_KEY = "freshness_threshold";
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    static {
+        final SimpleModule module = new SimpleModule("AlertControlller");
```
vjkoskela: One too many "l"'s.

app/controllers/ExpressionController.java
```diff
@@ -58,6 +63,53 @@ public ExpressionController(final Configuration configuration, final ExpressionR
...
 
     /**
+     * Adds or updates an expression in the expression repository.
+     * If a valid UUID is passed, it will update the corresponding expression if it exists or create a new one with this id.
+     * If no id is passed, it will generate an id.
+     *
+     * @param expressionId The <code>String</code> representation of the id of the expression to be updated.
+     * @return Success if the expression was created or updated successfully, an error code otherwise.
+     */
+    public Result addOrUpdate(final String expressionId) {
```
vjkoskela: Same comments here as in the alert controller.

app/controllers/ExpressionController.java
```diff
@@ -177,4 +229,5 @@ private ExpressionController(final int maxLimit, final ExpressionRepository expr
 
     private static final int DEFAULT_MAX_LIMIT = 1000;
     private static final Logger LOGGER = LoggerFactory.getLogger(ExpressionController.class);
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
vjkoskela: You should probably add the build to the OM here too given if registers the referenced types recursively and so would be future-proofed against any changes to the view.

conf/portal.routes
```diff
@@ -35,10 +35,14 @@ GET        /v1/hosts/query              controllers.HostController.query(name: S
...
 GET        /v1/alerts/query             controllers.AlertController.query(contains: String ?= null, context: String ?= null, cluster: String ?= null, service: String ?= null, limit: java.lang.Integer ?= null, offset: java.lang.Integer ?= null)
 GET        /v1/alerts/:id               controllers.AlertController.get(id: String)
+POST       /v1/alerts                   controllers.AlertController.addOrUpdate(id: String ?= null)
+PUT        /v1/alerts/:id               controllers.AlertController.addOrUpdate(id: String)
 
 # Expressions
 GET        /v1/expressions/query        controllers.ExpressionController.query(contains: String ?= null, cluster: String ?= null, service: String ?= null, limit: java.lang.Integer ?= null, offset: java.lang.Integer ?= null)
 GET        /v1/expressions/:id          controllers.ExpressionController.get(id: String)
+POST       /v1/expressions              controllers.ExpressionController.addOrUpdate(id: String ?= null)
+PUT        /v1/expressions/:id          controllers.ExpressionController.addOrUpdate(id: String)
```
vjkoskela: Yeah I don't think this works.

test/controllers/AlertControllerTest.java
```diff
@@ -0,0 +1,396 @@
...
+    public static void shutdown() {
+        alertRepo.close();
+        if (app != null) {
+            Helpers.stop(app);
+        }
+    }
+
+    @Test
+    public void testCreateValidCase() {
+        final String body = "{\"context\":\"CLUSTER\", \"name\":\"test-name\", \"cluster\":\"test-cluster\", "
```
vjkoskela: Can we get rid of the inline Json please?

app/controllers/AlertController.java
```diff
@@ -61,6 +72,89 @@ public AlertController(final Configuration configuration, final AlertRepository
     }
 
     /**
+     * Adds an alert in the alert repository.
+     *
+     * @return Created if the alert was created successfully, an error code otherwise.
+     */
+    public Result add() {
+        Alert alert;
```
vjkoskela: I believe with java 8 you can declare this final since the try-catch either sets it or returns. You may have to include line 82 inline on line 83.

app/controllers/AlertController.java
```diff
@@ -61,6 +72,89 @@ public AlertController(final Configuration configuration, final AlertRepository
...
+            alert = convertToInternalAlert(viewAlert);
+        } catch (final IOException | ConstraintsViolatedException | IllegalArgumentException e) {
+            LOGGER.error()
+                    .setMessage("Failed to build an alert.")
+                    .setThrowable(e)
+                    .log();
+            return badRequest("Invalid request body.");
+        }
+
+        if (_alertRepository.get(alert.getId()).isPresent()) {
```
vjkoskela: This is unnecessary, the model being posted will always contain the id and the code should perform an upsert. The user's intention is either to create or update. According to @matthayter and @BrandonArp this was what was discussed as the standard for the CRUD REST controllers. So to recap:

 - PUT for add/update
 - Entity contains uuid
 - Optional uuid in the post uri -- @matthayter or @BrandonArp confirm

If we need stronger semantics on update (to prevent unintentional overwrites) we will need to version the widgets. In the case of @matthayter -'s evergreen project that is not necessary for the first version. For your alerts/expressions project that would probably be a good thing. One of the next features will be change history for alerts and expressions so that will require some versioning work anyway (we can leave the change conflict work until then).

app/controllers/AlertController.java
```diff
@@ -61,6 +72,89 @@ public AlertController(final Configuration configuration, final AlertRepository
...
+            LOGGER.error()
+                    .setMessage(String.format("An alert with id %s already exists", alert.getId().toString()))
+                    .log();
```
vjkoskela: Granted this statement will be removed with the above change, but please consider that this should not be an error. The client sent bad data for the operation intended. At most a warning, I could even believe it's an info. Second, you should never format messages into the structured log builder. The point of the builder is to accept supporting data and context in such a way that it is separate and parseable from the message.

app/controllers/AlertController.java
```diff
@@ -178,16 +272,54 @@ public Result query(
...
-        return extensionMap;
+        final Map<String, Object> nagiosMap = Maps.newHashMap();
+        nagiosMap.put(NAGIOS_EXTENSION_SEVERITY_KEY, nagiosExtension.getSeverity());
+        nagiosMap.put(NAGIOS_EXTENSION_NOTIFY_KEY, nagiosExtension.getNotify());
+        nagiosMap.put(NAGIOS_EXTENSION_MAX_CHECK_ATTEMPTS_KEY, nagiosExtension.getMaxCheckAttempts());
+        nagiosMap.put(NAGIOS_EXTENSION_FRESHNESS_THRESHOLD_KEY, nagiosExtension.getFreshnessThreshold().getStandardSeconds());
+        return ImmutableMap.of(NAGIOS_EXTENSION_KEY, nagiosMap);
     }
 
     /**
```
vjkoskela: public before private.

app/controllers/AlertController.java
```diff
@@ -61,6 +72,89 @@ public AlertController(final Configuration configuration, final AlertRepository
...
+            alert = convertToInternalAlert(viewAlert);
+        } catch (final IOException | ConstraintsViolatedException | IllegalArgumentException e) {
+            LOGGER.error()
+                    .setMessage("Failed to build an alert.")
+                    .setThrowable(e)
+                    .log();
+            return badRequest("Invalid request body.");
+        }
+
+        if (!_alertRepository.get(alert.getId()).isPresent()) {
```
vjkoskela: To resolve the deletion in addition to the version described above we will need to add a tombstone (aka soft delete) semantic to the data. I assume you would prefer to do these separately. Please file issues for each against the project.

app/controllers/AlertController.java
```diff
@@ -178,16 +272,54 @@ public Result query(
...
+        }
+        return new DefaultQuantity.Builder()
+                .setUnit(viewQuantity.getUnit())
+                .setValue(viewQuantity.getValue())
+                .build();
+    }
+
+    private NagiosExtension getNagiosExtensions(final Map<String, Object> extensionsMap) throws IOException {
+        if (extensionsMap == null || extensionsMap.isEmpty()) {
+            return null;
```
vjkoskela: rename: convertToInternalNagiosExtension

I am not a fan of returning null here. At the _very_ least you should mark the return result nullable. Alternatively, consider returning an optional.

test/resources/controllers/AlertControllerTest.testCreateInvalidPeriodCase.json
```diff
@@ -0,0 +1,21 @@
+{
+    "id" : "1a3d037c-ee11-11e1-a1e5-00259060b5c0",
```
vjkoskela: Looks like you have the same id in here for several creates. Is that intentional? Does it matter?

misradeepika: Nope doesnt matter

app/controllers/ExpressionController.java
BrandonArp: You should probably grab the Commons OM as a base instead of creating your own.

app/controllers/AlertController.java
```diff
@@ -61,6 +72,38 @@ public AlertController(final Configuration configuration, final AlertRepository
     }
 
     /**
+     * Adds an alert in the alert repository.
+     *
+     * @return Created if the alert was created successfully, an error code otherwise.
```
vjkoskela: This comment does not match behavior.

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

Please check all documentation given the changes these endpoints have gone through.

app/controllers/AlertController.java
```diff
@@ -220,4 +318,5 @@ private AlertController(final int maxLimit, final AlertRepository alertRepositor
     private static final String NAGIOS_EXTENSION_NOTIFY_KEY = "notify";
     private static final String NAGIOS_EXTENSION_MAX_CHECK_ATTEMPTS_KEY = "max_check_attempts";
     private static final String NAGIOS_EXTENSION_FRESHNESS_THRESHOLD_KEY = "freshness_threshold";
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
vjkoskela: You should use ObjectMapperFactory from commons here. Since you are not modifying the instance use .getInstance (as opposed to .createInstance()).

app/controllers/ExpressionController.java
```diff
@@ -58,6 +65,38 @@ public ExpressionController(final Configuration configuration, final ExpressionR
     }
 
     /**
+     * Adds an expression in the expression repository.
+     *
+     * @return Created if the expression was created successfully, an error code otherwise.
```
vjkoskela: Does not return Created.

app/controllers/ExpressionController.java
```diff
@@ -177,4 +235,5 @@ private ExpressionController(final int maxLimit, final ExpressionRepository expr
 
     private static final int DEFAULT_MAX_LIMIT = 1000;
     private static final Logger LOGGER = LoggerFactory.getLogger(ExpressionController.class);
+    private static final ObjectMapper OBJECT_MAPPER = ObjectMapperFactory.createInstance();
```
vjkoskela: Should use getInstance instead of createInstance.

app/models/view/Alert.java
```diff
@@ -113,7 +113,15 @@ public void setExtensions(final Map<String, Object> extensions) {
         _extensions = extensions;
     }
 
+    /**
+     * Returns the extensions.
+     *
+     * @return null if extensions are null, an unmodifiable map otherwise.
```
vjkoskela: Agreed. Further, I think you can leave this as an unprotected get (e.g. not wrapped in unmodifiable) as the setter above is not copy-write protected either. I'm okay with this since it's a view object and does not enter the codebase beyond the controller instance receiving/returning the view instance. @BrandonArp thoughts?

BrandonArp: I'd prefer to see immutable.  Can we have the setter take and ImmutableMap?  Then you wont have to protect it on the way out (the getter).

vjkoskela: Sold. Let's apply the same immutability principles to the view models.

vjkoskela:
@misradeepika Please fix the build and update the pr to the latest.

PR #18, opened 2016-03-09
Streaming Connection Proxying Fixes

vjkoskela:
Refuse to proxy connections to localhost and add a heartbeat check before considering the connection established.


BrandonArp:
LGTM.

PR #19, opened 2016-03-11
Add database evolutions and data models for evergreen.

matthayter:


vjkoskela:
Please fix the build:

https://travis-ci.org/ArpNetworking/metrics-portal/builds/115216672

app/models/ebean/VersionSet.java
```diff
@@ -0,0 +1,61 @@
...
+ * @author Matthew Hayter (mhayter at groupon dot com)
+ */
+@Entity
+@Table(name = "version_sets", schema = "portal")
+public class VersionSet {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @Column(name = "id")
+    private Long id;
```
vjkoskela: Take a look at the other ebean models, we've been able to stick to the member underscore prefix and we've also provided setters/getters.

app/models/ebean/VersionSpecification.java
```diff
@@ -0,0 +1,65 @@
...
+    @CreatedTimestamp
+    @Column(name = "created_at")
+    private Timestamp createdAt;
+
+    @UpdatedTimestamp
+    @Column(name = "updated_at")
+    private Timestamp updatedAt;
+
+    /*
+    Note that these precedence values are intended to be sequential; each new row inserted with some precedence will force the update
```
vjkoskela: Interesting. Can you think of other ways to maintain an ordering without requiring an O(N) update operation on insert to head?

app/models/ebean/VersionSet.java
matthayter: I should have commented already sorry - I was following the existing convention (naturally :P) but @BrandonArp specifically recommended I do it this way - I'll let you guys sort that out offline

conf/db/migration/metrics_portal_ddl/common/V6__add_evergreen_tables.sql
```diff
@@ -0,0 +1,47 @@
...
+    UNIQUE (version_set, package_version)
+);
+
+CREATE TABLE portal.version_specifications (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP NOT NULL DEFAULT now(),
+    updated_at TIMESTAMP NOT NULL DEFAULT now(),
+    precedence INTEGER NOT NULL UNIQUE,
+    version_set_id INTEGER REFERENCES portal.version_sets
+);
```
vjkoskela: Some of these entities will require external resource identifiers, for example, each version_specification likely needs a UUID to write the REST APIs. Are these modifications you plan on doing with that CR (as a separate data migration) or do you want to include these here?

app/models/ebean/VersionSpecification.java
matthayter: Definitely, but I feel this is worth the simplicity for some small number of rules; though there may be a nice simple way that I'm not thinking of :P

Other options are (with complexity that I don't think is worth it for now):
- Linked list style ordering
- 'Gapped' ordering - start with 100, 200, 300 etc
- ???

Now I'm rethinking it, gapped ordering may be not that much more complex... hm getting the constant down would be easy enough but to get less better than O(n) would be harder, and you may run into the limits of `long`... not sure

app/models/ebean/VersionSet.java
matthayter: Added issue #20 to track

conf/db/migration/metrics_portal_ddl/common/V6__add_evergreen_tables.sql
matthayter: Argh, added!

app/models/ebean/VersionSpecification.java
vjkoskela: Maybe I'm not seeing it, but this method seems to be more complex. When someone does a rule insertion you will need to change N - K entries in the database where K is the index to which the rule was added. The linked list solution seems the most straight forward where each row in the database has a link to the next rule and the last one has null. In either case the internal model should just be an ordered list, it's just a question of what the storage model will be. 

From a user perspective, will each edit be applied at once (e.g. add row here, change row content, delete row, etc.) or will it be change all the rows and save? Of course we can do either on the front-end (and should retain the possibility to change) and map it correctly to the back-end. Although, in the save everything case we need to overwrite the entire table anyway unless we're going to diff to figure out the increments.

So I don't think I have a problem with the write everything philosophy unless @BrandonArp has any objections to it.

Finally, I'm sure you will implement some sort of optimistic locking to make this nice and sane ;)

Ref:
http://stackoverflow.com/questions/21388/represent-ordering-in-a-relational-database

matthayter: Added linked-list style ordering!

conf/db/migration/metrics_portal_ddl/common/V6__add_evergreen_tables.sql
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+CREATE TABLE portal.package_versions (
+    id SERIAL PRIMARY KEY,
+    created_at TIMESTAMP NOT NULL DEFAULT now(),
+    updated_at TIMESTAMP NOT NULL DEFAULT now(),
+    name VARCHAR(255) NOT NULL,
+    version VARCHAR(255) NOT NULL,
+    uri VARCHAR(2047) NOT NULL,
+    UNIQUE (name, version, uri)
```
vjkoskela: Needs to be unique on name+version only.

app/models/ebean/VersionSpecification.java
```diff
@@ -0,0 +1,71 @@
...
+    @CreatedTimestamp
+    @Column(name = "created_at")
+    private Timestamp createdAt;
+
+    @UpdatedTimestamp
+    @Column(name = "updated_at")
+    private Timestamp updatedAt;
+
+    /*
+    Ordering is done in a 'linked-list' style.
```
vjkoskela: Let's trade this for more comprehensive class comments. In particular, a description of how a set of version specification instances should be interpreted AND that we currently only allow a single version specification (e.g. rule set). Any ideas on how to proceed with the extension to multiple rule sets would be a bonus.

conf/db/migration/metrics_portal_ddl/common/V6__add_evergreen_tables.sql
```diff
@@ -0,0 +1,50 @@
...
+);
+
+CREATE TABLE portal.version_sets (
+    id SERIAL PRIMARY KEY,
+    uuid UUID NOT NULL,
+    version SERIAL NOT NULL UNIQUE,
```
vjkoskela: I think this should probably be a string too.

matthayter: Doone

app/models/ebean/VersionSpecification.java
matthayter: Class comments done; multiple rule sets left out for now.

conf/db/migration/metrics_portal_ddl/common/V6__add_evergreen_tables.sql
matthayter: Done

app/models/ebean/PackageVersion.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public String getUri() {
+        return uri;
+    }
+
+    public void setUri(final String uri) {
+        this.uri = uri;
+    }
+}
+// CHECKSTYLE.ON
```
vjkoskela: You need to turn each on that you turned off. You should do so in inverse order.

app/models/ebean/VersionSpecification.java
```diff
@@ -0,0 +1,139 @@
...
+import javax.persistence.ManyToOne;
+import javax.persistence.OneToMany;
+import javax.persistence.OneToOne;
+import javax.persistence.Table;
+
+/**
+ * Data model for Version Specifications (for Evergreen).
+ *
+ * A VersionSpecification models a 'rule' by which a host making a query to Evergreen may be matched to a VersionSet
+ * for purposes of specifying the correct packages and versions that should be installed on the host. The VersionSpecifications
```
vjkoskela: "packages and versions" => "package-versions"

app/models/ebean/VersionSpecification.java
```diff
@@ -0,0 +1,139 @@
...
+ * form a total order via the `next` attribute (i.e. ordered via the linked-list method).
+ *
+ * A VersionSpecification is said to match a host iff each of the VersionSpecificationAttributes match the corresponding
```
vjkoskela: iff => "if and only if"

app/models/ebean/VersionSpecification.java
```diff
@@ -0,0 +1,139 @@
...
+ * propery provided by the host.
```
vjkoskela: propery => property

app/models/ebean/VersionSpecificationAttribute.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+/**
+ * Data model for Version Specification Attributes (for Evergreen). Holds a simple key-value String pair, intended to
```
vjkoskela: References to classes/code should be marked-up; e.g.:

```
<code>String</code>
```

You should also either use a class reference e.g. "VersionSpecificationAttribute" or some canonical non-proper name, but not a proper name. In this case I would prefer the former.

test/com/arpnetworking/metrics/portal/models/ebean/VersionSpecificationTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import play.test.WithApplication;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * @author Matthew Hayter (mhayter at groupon dot com)
```
vjkoskela: OCD: Please add standard test description to complete javadoc.

test/com/arpnetworking/metrics/portal/models/ebean/VersionSpecificationTest.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+public class VersionSpecificationTest extends WithApplication {
+
+    @SuppressWarnings("unchecked")
```
vjkoskela: Should be able to get rid of this.

matthayter:
@vjkoskela All addressed!

matthayter:
Please note Ville has merged #23 into this, thus addressing issue #20 here - @vjkoskela if this was not your intent, lmk

vjkoskela:
@matthayter I reviewed and merged a PR you sent me, if you did not intend for it be merged please in the future indicate this in the PR or preferably don't send the PR until is ready to be merged.

As for this change it looks good and I'll merge it. That's what you intend me to do, right? :smiley:

PR #24, opened 2016-03-24
modified: a boolean 'created' variable

damianball:
Since 'created' is past tense and technically the new object has yet to be created when 'created' is set to 'true', I think using "isNew[...]" as a variable name is easier to understand in this context.


BrandonArp:
Thanks much!  Will merge once build passes.

PR #43, opened 2016-04-20
Fix dragging graphs under config dialog

PhoenixRion:
Make the config modal a sibling of the sortable graph instead of a child.  This makes it so that the draggable handle is below the modal.


PhoenixRion:
change best viewed while ignoring whitespace

Issue #29, opened 2016-04-07
Graph Title Too Long


vjkoskela:
Add support to shorten the displayed graph title in a more reasonable way. Currently, it seems to center it which truncates both the beginning and the end (seems like the least useful option to me). More importantly, we should provide a hover showing the full metric name and service name.

PhoenixRion:
Hover for the full name should be possible.
While considering how to shorten the name, it seems that setting a static cap on the length might be the most practical.  Because Flotr doesn't allow the font to be set, we can't translate string length to pixels.

PhoenixRion:
https://github.com/ArpNetworking/metrics-portal/tree/actually-issue-29

vjkoskela:
I assume you mean static cap on the display length of the name? Is there an existing widget that cuts-off the name with trailing "..." or some such?

PS. Not what what the link is to.

PhoenixRion:
I linked to the branch, I guess I could have linked directly to the diff.  https://github.com/ArpNetworking/metrics-portal/compare/actually-issue-29

yes, I was suggesting a static cap on the display length (in characters)
The actual text is being rendered by Flotr.
Adding the "..." to the end as an indication that the name was truncated is a good idea.

vjkoskela:
Thanks! Sorry, I missed that. 

I understand the suggestion. My concern is that with a non-fixed length font the length difference between "www" and "iii" can get rather large The result may be an equally poor experience if the text is prematurely truncated. The counter argument is that on average the experience should be average ;)

Hence, my question is whether there are widgets that would truncate the text for us based on pixel length. This is outside my domain experience but a quick Google shows some promise:

http://stackoverflow.com/questions/15308061/how-to-avoid-text-overflow-in-twitter-bootstrap
https://css-tricks.com/line-clampin/
http://www.w3schools.com/cssref/css3_pr_text-overflow.asp

At first glance the hit on Stackoverflow looks like a good match. 

@PhoenixRion would you have some time to run these down to see if they are workable before we considering a blanket length truncation?

PhoenixRion:
So if I switch htmlText on for Flotr, I can get it to clip the text to the graph size but for some reason the ellipsis never shows up like some of the examples show.  I can, hopefully, take a look at adding the mouse over later this week but I am going on-call today.  :(

BrandonArp:
NOOOOOOOOO! htmlText is really bad for smooth scrolling mode.  It *murders*
smooth scrolling mode.

[...]

PR #49, opened 2016-06-17
GUI: Add a tooltip to the metric with the full metric name.

matthayter:
... so that when you hover over a metric in the tree, you see the full metric name.
screen shot 2016-06-16 at 6 16 34 pm

### DIFF
--- a/app/assets/javascripts/classes/MetricNodeVM.ts
+++ b/app/assets/javascripts/classes/MetricNodeVM.ts
@@ -22,6 +22,7 @@ import ko = require('knockout');
 import ns = require('naturalSort');
 
 class MetricNodeVM implements BrowseNode {
+    metricName: KnockoutObservable<string>;
     name: KnockoutObservable<string>;
     children: KnockoutObservableArray<BrowseNode>;
     subfolders: KnockoutObservableArray<BrowseNode>;
@@ -29,8 +30,9 @@ class MetricNodeVM implements BrowseNode {
     renderAs: KnockoutObservable<string>;
     visible: KnockoutObservable<boolean>;
 
-    constructor(name: string, id: string) {
-        this.name = ko.observable(name);
+    constructor(metricName: string, displayName: string, id: string) {
+        this.metricName = ko.observable(metricName);
+        this.name = ko.observable(displayName);
         this.children = ko.observableArray<BrowseNode>();
         this.subfolders = ko.observableArray<BrowseNode>();
         this.expanded = ko.observable(false);
--- a/app/assets/javascripts/classes/MetricsBrowseList.ts
+++ b/app/assets/javascripts/classes/MetricsBrowseList.ts
@@ -36,7 +36,7 @@ class MetricsBrowseList {
 
             service.children.forEach((metric) => {
                 var nameParts = this.getPathParts(metric.name);
-                var metricNode = this.addMetric(nameParts, serviceNodeVM);
+                var metricNode = this.addMetric(metric.name, nameParts, serviceNodeVM);
                 metric.children.forEach((statistic) => {
                     this.addStatistic(metricNode, new GraphSpec(service.name, metric.name, statistic.name,
                                                                 Series.defaultPoints(), Series.defaultLines(), Series.defaultBars()))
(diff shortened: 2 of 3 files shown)


vjkoskela:
Terrific! Should the text pop over be prefixed with the service name?

Since you are an expert on labels, how about the same change for he graph title? Ryan was going to work on that by extracting it into a Dom element first. Maybe check with him on the state of that.

app/assets/javascripts/classes/MetricNodeVM.ts
```diff
@@ -22,15 +22,17 @@ import ko = require('knockout');
...
     children: KnockoutObservableArray<BrowseNode>;
     subfolders: KnockoutObservableArray<BrowseNode>;
     expanded: KnockoutObservable<boolean>;
     renderAs: KnockoutObservable<string>;
     visible: KnockoutObservable<boolean>;
 
-    constructor(name: string, id: string) {
-        this.name = ko.observable(name);
+    constructor(metricName: string, displayName: string, id: string) {
+        this.metricName = ko.observable(metricName);
```
vjkoskela: Why do I feel that we had this duality before and explicitly removed it?

Is it possible to do this with the existing name attribute? Obviously it will need to be formatted for display.

matthayter: The `name` is the name to show on the part of the tree, i.e. just the final _segment_ of the metric name; so nah, I don't think it can be done in any reasonable way using the `name` attributes, but open to ideas

matthayter:
Ville, you make me scared to make small changes - it's like walking sideways along the side of a mountain, and you're dislodging rocks hoping for an avalanche 😄 

The graph title is a great idea, but please don't hold this change hostage : P @PhoenixRion Any word on the graph title thing from Ville's comment above?

app/assets/javascripts/classes/MetricNodeVM.ts
BrandonArp: There was this duality before, but it was much worse.  It used to be intertwined with another duality that existed between browse nodes and statistic nodes.  What Matt did seems alright to me.  The other option is to have references to the parent nodes.  That would be a significant refactoring and is roughly equivalent.

BrandonArp:
I'm good with merging this.  Thanks for the work.  I agree that it would be nice to get the title on the graphs, but this certainly shouldn't be held up for that.

matthayter:
I can defs add the service name - my non-designer brain doesn't how it should look though:
`Service: collectd | Metric: disk/root/free`
`(Service: collectd) disk/root/`
`disk/root/free (Service: collectd)`

`collectd disk/root/free`

`[collectd] disk/root/free`
`(collectd) disk/root/free`

app/assets/javascripts/classes/MetricNodeVM.ts
```diff
@@ -41,6 +45,10 @@ class MetricNodeVM implements BrowseNode {
     expandMe() {
         this.expanded(this.expanded() == false);
     }
+
+    tooltip = ko.pureComputed(function() {
```
matthayter: @BrandonArp can you verify this is a decent way to add this method? I can't wrap my brain around mixing the typescript syntax with a `ko.pureComputed` instance

matthayter:
I picked one I like, but can change:
screen shot 2016-06-17 at 1 31 41 pm

matthayter:
That failing build seems related to this: https://github.com/travis-ci/travis-ci/issues/3775

vjkoskela:
@matthayter The build is green at the moment. My intention was not to hold this change hostage, but rather to ensure consistency for pop-overs. Since you and Ryan work together I just want the popovers in both contexts to look and behave the same. 

However, if @PhoenixRion is no longer working on #29 then I would like to know if this something you can take on. It's a little more complex in that you need to disable the graph's internal title and add a title DOM element which can then be length restricted and have a tool-tip.

With that I will throw my hat in and say the format should be more condensed; how about:

```
{service}:{metric}
```

e.g.
```
collectd:cpu/idle
```

@BrandonArp So you are good with the second name attribute on the model? I know you fought this fight once so I want to make sure this is heading in a direction that is compatible with the long term evolution of the underlying code.

BrandonArp:
Yes, I'm good with this dual name. I was mostly fighting the dual tree thing, not the names.  Anyway, I like {service}:{metric} as well.  Change that and I'll get you merged up.

matthayter:
Oh nooo! I was making this PR so make it clear what the metric name is to someone who didn't write the code emitting the metrics, and so they didn't have to go to the `query.log`. Now it'll still cause confusion - what if someone has a `:` in the metric name?
`redis:status:400/count` ? booooo

matthayter:
We're back to the original appearance in the PR description.

PR #51, opened 2016-06-20
Assert typeahead's dependency on jquery in RequireJS config.

### LINKED ISSUE #50: Portal breaks on refresh
If you load up metrics portal for the first time, everything works great. If you hit refresh, you get this in the console:
`typeahead.bundle.js.src.js:26 Uncaught TypeError: Cannot read property 'isArray' of undefined `
with associated line of code:
screen shot 2016-06-20 at 10 31 56 am
After that, typing a host in and hitting connect results in:

screen shot 2016-06-20 at 10 33 36 am

Which never actually connects; pretty much totally broken.

I'm guessing its because of caching; typeahead loads first in the cache hit case and promptly breaks due to no jQuery.

matthayter:
Fixes #50

### DIFF
--- a/app/assets/javascripts/start_app.ts
+++ b/app/assets/javascripts/start_app.ts
@@ -63,6 +63,9 @@ requirejs.config({
         },
         'bootstrap' : {
             deps : [ 'jquery' ]
+        },
+        'typeahead' : {
+            deps : [ 'jquery']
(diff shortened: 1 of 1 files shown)


BrandonArp:
Good find!  Will merge once the build completes.

Issue #17, opened 2016-03-09
Websocket Proxy Feature Flag


vjkoskela:
The ability to proxy websocket connections for the client allows clients to be isolated from most of the network points generating telemetry. However, it also creates a potential back-door. We should feature flag the proxy behavior even if the default is to have it enabled.

BrandonArp:
This was closed with #54.

PR #54, opened 2016-06-27
Feature Flags

vjkoskela:
Added feature flags to control different aspects of the application: telemetry, hosts, expressions, alerts. Also added feature setting for telemetry to control proxy.

### DIFF
--- a/app/controllers/ApplicationController.java
+++ b/app/controllers/ApplicationController.java
@@ -1,0 +1,54 @@
+/**
+ * Copyright 2016 Inscope Metrics, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


app/controllers/ApplicationController.java
```diff
@@ -0,0 +1,67 @@
...
+        _alertsEnabled = configuration.getBoolean("portal.features.alerts.enabled");
+    }
+
+    /**
+     * Render the header view.
+     *
+     * @return Rendered header view.
+     */
+    public F.Promise<Result> getHeaderViewModel() {
+        return F.Promise.pure(ok(views.html.HeaderViewModel.render(
```
BrandonArp: Please make a features object that we can pass into the render function.  I don't like a function that takes half a dozen boolean parameters.  Any refactoring would be nearly impossible.

BrandonArp:
Just the one comment and then I'm good.  It would be nice to have a completed "Features" object that is injectable. ( :

app/controllers/ApplicationController.java
vjkoskela: Although I agree whole heartedly with the sentiment I don't see how this makes refactoring impossible the code path was private and extremely well constrained to a single interaction (between controller and view). Nevertheless, the feature configuration has been modeled and is now injected.

vjkoskela:
Updated review with injected Features instance.

BrandonArp:
Perfect!  But you're missing the Features classes in the commit.

vjkoskela:
Addresses #17.

PR #55, opened 2016-06-28
Latest Commons

vjkoskela:
Upgrade to latest commons and other packages. Stop catching ConstraintsViolatedException when deserializing JSON with Jackson.


vjkoskela:
Addresses #17.

PR #62, opened 2016-09-10
update play to 2.5.6

BrandonArp:
I will re-enable warnings when I port the websocket stuff to Akka Streaming.  The websocket deprecation warnings are the only ones right now.


vjkoskela:
Looks great to me. This is obviously at least a Y change. I wonder if we should treat it as an X change?

Otherwise, I trust you did reasonable manual testing since this project does not include many unit/integration tests.

BrandonArp:
Yeah, I did a fair amount of manual testing.  I was really pleased with how the WS proxying came out.  Much easier than before.

I'm thinking about making an X change with the addition of the organizations once that gets merged.  What do you think?

vjkoskela:
Agreed. Everything looks very nice. 

That works for me. Does the organizations actually cause an X change (e.g. break something else) or is it just because it's a large shift?

Merging.

PR #63, opened 2016-09-11
Organization support

BrandonArp:
Here's the big one.  

Organization support.  
Had to change the way etags were computed.
Postgres now works with UUIDs (I don't think it was working before the EBean update)

I tried to keep external APIs the same, so only the default organization can be used at this time.  Only very minor work will need to be done on the APIs to expose the organization id.  We should talk about authentication before we do that, though.

### DIFF
--- a/app/actors/ClusterShutdownActor.java
+++ b/app/actors/ClusterShutdownActor.java
@@ -1,0 +1,75 @@
+/**
+ * Copyright 2016 Smartsheet.com
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package actors;
+
+import akka.actor.Address;
+import akka.actor.Props;
+import akka.actor.UntypedActor;
+import akka.cluster.Cluster;
+import akka.cluster.ClusterEvent;
+import akka.cluster.Member;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Actor that subscribes to cluster events and triggers a CompletableFuture when it sees itself removed
+ * from the cluster.
+ *
+ * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
+ */
+public class ClusterShutdownActor extends UntypedActor {
+    /**
+     * Creates a {@link Props} for this actor.
+     *
+     * @param shutdownFuture A future to be completed at cluster shutdown.
+     * @return A new Props.
+     */
+    public static Props props(final CompletableFuture<Boolean> shutdownFuture) {
+        return Props.create(ClusterShutdownActor.class, shutdownFuture);
+    }
+
+    /**
+     * Public constructor.
+     *
+     * @param shutdownFuture A future to be completed at cluster shutdown.
+     */
+    public ClusterShutdownActor(final CompletableFuture<Boolean> shutdownFuture) {
+        _shutdownFuture = shutdownFuture;
+
+        final Cluster cluster = Cluster.get(context().system());
+        _selfAddress = cluster.selfAddress();
+        cluster.subscribe(self(), ClusterEvent.initialStateAsEvents(), ClusterEvent.MemberRemoved.class);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void onReceive(final Object message) throws Throwable {
+        if (message instanceof ClusterEvent.MemberRemoved) {
+            final ClusterEvent.MemberRemoved removed = (ClusterEvent.MemberRemoved) message;
+            final Member member = removed.member();
+            if (_selfAddress.equals(member.address())) {
+                _shutdownFuture.complete(Boolean.TRUE);
+            }
+        } else {
+            unhandled(message);
+        }
+    }
+
+    private final CompletableFuture<Boolean> _shutdownFuture;
+    private final Address _selfAddress;
+}
--- a/app/com/arpnetworking/database/h2/triggers/BaseUpdateEtagTrigger.java
+++ b/app/com/arpnetworking/database/h2/triggers/BaseUpdateEtagTrigger.java
@@ -16,11 +16,12 @@
 package com.arpnetworking.database.h2.triggers;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import models.ebean.Organization;
 import org.h2.api.Trigger;
 
 import java.sql.Connection;
-import java.sql.PreparedStatement;
 import java.sql.SQLException;
+import java.util.function.Consumer;
 
 /**
  * Base class for H2 triggers for updating etags.
@@ -32,10 +33,12 @@ public abstract class BaseUpdateEtagTrigger implements Trigger {
     /**
      * Public constructor.
      *
-     * @param sequenceName Name of the sequence that should be updated.
+     * @param incrementEtag Method to update the etag.
+     * @param orgIndex The index to find the organization for which to update the etag.
      */
-    public BaseUpdateEtagTrigger(final String sequenceName) {
-        _queryStatement = String.format("SELECT NEXTVAL('%s');", sequenceName);
+    public BaseUpdateEtagTrigger(final Consumer<Organization> incrementEtag, final int orgIndex) {
+        _incrementEtag = incrementEtag;
+        _orgIndex = orgIndex;
     }
 
     /**
@@ -56,9 +59,9 @@ public abstract class BaseUpdateEtagTrigger implements Trigger {
     @Override
     @SuppressFBWarnings("SQL_PREPARED_STATEMENT_GENERATED_FROM_NONCONSTANT_STRING")
     public void fire(final Connection conn, final Object[] oldRow, final Object[] newRow) throws SQLException {
-        try (PreparedStatement stmt = conn.prepareStatement(_queryStatement)) {
-            stmt.execute();
-        }
+        final long orgId = (long) (oldRow != null ? oldRow[_orgIndex] : newRow[_orgIndex]);
+        final Organization org = Organization.refById(orgId);
+        _incrementEtag.accept(org);
     }
 
     /**
@@ -77,5 +80,6 @@ public abstract class BaseUpdateEtagTrigger implements Trigger {
         // ignore
     }
 
-    protected String _queryStatement;
+    private final Consumer<Organization> _incrementEtag;
+    private final int _orgIndex;
 }
--- a/app/com/arpnetworking/database/h2/triggers/ExpressionsUpdateEtagTrigger.java
+++ b/app/com/arpnetworking/database/h2/triggers/ExpressionsUpdateEtagTrigger.java
@@ -15,6 +15,8 @@
  */
 package com.arpnetworking.database.h2.triggers;
 
+import models.ebean.ExpressionEtags;
+
 /**
  * Trigger to update Etag after every insert, delete or update statement.
  *
@@ -26,6 +28,6 @@ public class ExpressionsUpdateEtagTrigger extends BaseUpdateEtagTrigger {
      * Public no args constructor.
      */
     public ExpressionsUpdateEtagTrigger() {
-        super("portal.expressions_etag_seq");
+        super(ExpressionEtags::incrementEtag, 9);
     }
 }
--- a/app/com/arpnetworking/metrics/portal/alerts/AlertRepository.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/AlertRepository.java
@@ -17,6 +17,7 @@ package com.arpnetworking.metrics.portal.alerts;
 
 import models.internal.Alert;
 import models.internal.AlertQuery;
+import models.internal.Organization;
 import models.internal.QueryResult;
 
 import java.util.Optional;
@@ -43,16 +44,18 @@ public interface AlertRepository {
      * Get the <code>Alert</code> by identifier.
      *
      * @param identifier The <code>Alert</code> identifier.
+     * @param organization The organization owning the alert.
      * @return The matching <code>Alert</code> if found or <code>Optional.empty()</code>.
      */
-    Optional<Alert> get(final UUID identifier);
+    Optional<Alert> get(UUID identifier, Organization organization);
 
     /**
      * Create a query against the alerts repository.
      *
+     * @param organization Organization to search in.
      * @return Instance of <code>AlertQuery</code>.
      */
-    AlertQuery createQuery();
+    AlertQuery createQuery(Organization organization);
 
     /**
      * Query alerts.
@@ -65,14 +68,16 @@ public interface AlertRepository {
     /**
      * Retrieve the total number of alerts in the repository.
      *
+     * @param organization The organization owning the alerts.
      * @return The total number of alerts.
      */
-    long getAlertCount();
+    long getAlertCount(Organization organization);
 
     /**
      * Add a new alert or update an existing one in the repository.
      *
      * @param alert The alert to add to the repository.
+     * @param organization The organization owning the alert.
      */
-    void addOrUpdateAlert(Alert alert);
+    void addOrUpdateAlert(Alert alert, Organization organization);
 }
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
@@ -24,12 +24,13 @@ import com.avaje.ebean.ExpressionList;
 import com.avaje.ebean.Junction;
 import com.avaje.ebean.PagedList;
 import com.avaje.ebean.Query;
-import com.avaje.ebean.SqlQuery;
 import com.avaje.ebean.Transaction;
 import com.google.inject.Inject;
+import models.ebean.AlertEtags;
 import models.ebean.NagiosExtension;
 import models.internal.Alert;
 import models.internal.AlertQuery;
+import models.internal.Organization;
 import models.internal.QueryResult;
 import models.internal.impl.DefaultAlert;
 import models.internal.impl.DefaultAlertQuery;
@@ -104,16 +105,18 @@ public class DatabaseAlertRepository implements AlertRepository {
      * {@inheritDoc}
      */
     @Override
-    public Optional<Alert> get(final UUID identifier) {
+    public Optional<Alert> get(final UUID identifier, final Organization organization) {
         assertIsOpen();
         LOGGER.debug()
                 .setMessage("Getting alert")
                 .addData("alertId", identifier)
+                .addData("organization", organization)
                 .log();
 
         final models.ebean.Alert ebeanAlert = Ebean.find(models.ebean.Alert.class)
                 .where()
                 .eq("uuid", identifier)
+                .eq("organization.uuid", organization.getId())
                 .findUnique();
         if (ebeanAlert == null) {
             return Optional.empty();
@@ -125,10 +128,13 @@ public class DatabaseAlertRepository implements AlertRepository {
      * {@inheritDoc}
      */
     @Override
-    public AlertQuery createQuery() {
+    public AlertQuery createQuery(final Organization organization) {
         assertIsOpen();
-        LOGGER.debug().setMessage("Preparing query").log();
-        return new DefaultAlertQuery(this);
+        LOGGER.debug()
+                .setMessage("Preparing query")
+                .addData("organization", organization)
+                .log();
+        return new DefaultAlertQuery(this, organization);
     }
 
     /**
@@ -147,7 +153,7 @@ public class DatabaseAlertRepository implements AlertRepository {
 
         // Compute the etag
         // TODO(deepika): Obfuscate the etag [ISSUE-7]
-        final Long etag = _alertQueryGenerator.getEtag();
+        final Long etag = _alertQueryGenerator.getEtag(query.getOrganization());
 
         final List<Alert> values = new ArrayList<>();
         pagedAlerts.getList().forEach(ebeanAlert -> values.add(convertFromEbeanAlert(ebeanAlert)));
@@ -160,9 +166,11 @@ public class DatabaseAlertRepository implements AlertRepository {
      * {@inheritDoc}
      */
     @Override
-    public long getAlertCount() {
+    public long getAlertCount(final Organization organization) {
         assertIsOpen();
         return Ebean.find(models.ebean.Alert.class)
+                .where()
+                .eq("organization.uuid", organization.getId())
                 .findRowCount();
     }
 
@@ -170,11 +178,12 @@ public class DatabaseAlertRepository implements AlertRepository {
      * {@inheritDoc}
      */
     @Override
-    public void addOrUpdateAlert(final Alert alert) {
+    public void addOrUpdateAlert(final Alert alert, final Organization organization) {
         assertIsOpen();
         LOGGER.debug()
                 .setMessage("Upserting alert")
                 .addData("alert", alert)
+                .addData("organization", organization)
                 .log();
 
 
@@ -182,6 +191,7 @@ public class DatabaseAlertRepository implements AlertRepository {
             models.ebean.Alert ebeanAlert = Ebean.find(models.ebean.Alert.class)
                     .where()
                     .eq("uuid", alert.getId())
+                    .eq("organization.uuid", organization.getId())
                     .findUnique();
             boolean isNewAlert = false;
             if (ebeanAlert == null) {
@@ -189,6 +199,7 @@ public class DatabaseAlertRepository implements AlertRepository {
                 isNewAlert = true;
             }
 
+            ebeanAlert.setOrganization(models.ebean.Organization.findByOrganization(organization));
             ebeanAlert.setCluster(alert.getCluster());
             ebeanAlert.setUuid(alert.getId());
             ebeanAlert.setMetric(alert.getMetric());
@@ -207,6 +218,7 @@ public class DatabaseAlertRepository implements AlertRepository {
             LOGGER.info()
                     .setMessage("Upserted alert")
                     .addData("alert", alert)
+                    .addData("organization", organization)
                     .addData("isCreated", isNewAlert)
                     .log();
             // CHECKSTYLE.OFF: IllegalCatchCheck
@@ -215,6 +227,7 @@ public class DatabaseAlertRepository implements AlertRepository {
             LOGGER.error()
                     .setMessage("Failed to upsert alert")
                     .addData("alert", alert)
+                    .addData("organization", organization)
                     .setThrowable(e)
                     .log();
             throw new PersistenceException(e);
@@ -303,9 +316,10 @@ public class DatabaseAlertRepository implements AlertRepository {
         /**
          * Gets the etag for the alerts table.
          *
+         * @param organization The organization owning the alert.
          * @return The etag for the table.
          */
-        long getEtag();
+        long getEtag(Organization organization);
     }
 
     /**
@@ -319,6 +333,7 @@ public class DatabaseAlertRepository implements AlertRepository {
         @Override
         public PagedList<models.ebean.Alert> createAlertQuery(final AlertQuery query) {
             ExpressionList<models.ebean.Alert> ebeanExpressionList = Ebean.find(models.ebean.Alert.class).where();
+            ebeanExpressionList = ebeanExpressionList.eq("organization.uuid", query.getOrganization().getId());
             if (query.getCluster().isPresent()) {
                 ebeanExpressionList = ebeanExpressionList.eq("cluster", query.getCluster().get());
             }
@@ -364,9 +379,8 @@ public class DatabaseAlertRepository implements AlertRepository {
          * {@inheritDoc}
          */
         @Override
-        public long getEtag() {
-            final SqlQuery sqlQuery = Ebean.createSqlQuery("SELECT CURRVAL('portal.alerts_etag_seq') AS etag;");
-            return sqlQuery.findUnique().getLong("etag");
+        public long getEtag(final Organization organization) {
+            return AlertEtags.getEtagByOrganization(organization);
         }
     }
 }
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/NoAlertRepository.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/NoAlertRepository.java
@@ -22,6 +22,7 @@ import com.arpnetworking.steno.LoggerFactory;
 import com.google.inject.Inject;
 import models.internal.Alert;
 import models.internal.AlertQuery;
+import models.internal.Organization;
 import models.internal.QueryResult;
 import models.internal.impl.DefaultAlertQuery;
 import models.internal.impl.DefaultQueryResult;
@@ -68,11 +69,12 @@ public final class NoAlertRepository extends ReadOnlyAbstractAlertRepository {
      * {@inheritDoc}
      */
     @Override
-    public Optional<Alert> get(final UUID identifier) {
+    public Optional<Alert> get(final UUID identifier, final Organization organization) {
         assertIsOpen();
         LOGGER.debug()
                 .setMessage("Getting alert")
                 .addData("identifier", identifier)
+                .addData("organization", organization)
                 .log();
         return Optional.empty();
     }
@@ -81,9 +83,12 @@ public final class NoAlertRepository extends ReadOnlyAbstractAlertRepository {
      * {@inheritDoc}
      */
     @Override
-    public AlertQuery createQuery() {
-        LOGGER.debug().setMessage("Preparing query").log();
-        return new DefaultAlertQuery(this);
+    public AlertQuery createQuery(final Organization organization) {
+        LOGGER.debug()
+                .setMessage("Preparing query")
+                .addData("organization", organization)
+                .log();
+        return new DefaultAlertQuery(this, organization);
     }
 
     /**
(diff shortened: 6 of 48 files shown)


app/actors/ClusterShutdownActor.java
```diff
@@ -0,0 +1,72 @@
...
+import akka.actor.Props;
+import akka.actor.UntypedActor;
+import akka.cluster.Cluster;
+import akka.cluster.ClusterEvent;
+import akka.cluster.Member;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
```
vjkoskela: Incomplete javadoc.

app/com/arpnetworking/database/h2/triggers/AlertsUpdateEtagTrigger.java
```diff
@@ -26,6 +26,6 @@
      * Public no args constructor.
      */
     public AlertsUpdateEtagTrigger() {
-        super("portal.alerts_etag_seq");
+        super("portal.alerts_etags", 15);
```
vjkoskela: What is 15?

app/com/arpnetworking/database/h2/triggers/BaseUpdateEtagTrigger.java
```diff
@@ -56,9 +60,21 @@ public void init(
     @Override
     @SuppressFBWarnings("SQL_PREPARED_STATEMENT_GENERATED_FROM_NONCONSTANT_STRING")
     public void fire(final Connection conn, final Object[] oldRow, final Object[] newRow) throws SQLException {
-        try (PreparedStatement stmt = conn.prepareStatement(_queryStatement)) {
+        final int updateCount;
+        final int orgId = (int) (oldRow != null ? oldRow[_orgIndex] : newRow[_orgIndex]);
+        Logger.info("Updating etag for organization " + orgId);
+        try (final PreparedStatement stmt = conn.prepareStatement(_updateStatement)) {
+            stmt.setInt(1, orgId);
```
vjkoskela: We should use named not positional arguments. I believe both are supported by EBean.

app/com/arpnetworking/database/h2/triggers/BaseUpdateEtagTrigger.java
```diff
@@ -56,9 +60,21 @@ public void init(
...
             stmt.execute();
+            updateCount = stmt.getUpdateCount();
         }
+        if (updateCount == 0) {
+            try (final PreparedStatement insert = conn.prepareStatement(_updateStatement)) {
+                insert.setInt(1, orgId);
```
vjkoskela: Same here.

app/com/arpnetworking/database/h2/triggers/BaseUpdateEtagTrigger.java
```diff
@@ -32,10 +33,13 @@
...
      * Public constructor.
      *
-     * @param sequenceName Name of the sequence that should be updated.
+     * @param updateTableName Name of the table that should be updated.
+     * @param orgIndex The index to find the organization for which to update the etag.
      */
-    public BaseUpdateEtagTrigger(final String sequenceName) {
-        _queryStatement = String.format("SELECT NEXTVAL('%s');", sequenceName);
+    public BaseUpdateEtagTrigger(final String updateTableName, final int orgIndex) {
+        _updateStatement = String.format("UPDATE %s SET etag = etag + 1 where organization = ?", updateTableName);
```
vjkoskela: So is it not possible to model this with EBean? -- it wasn't before because we were using a counter.

app/com/arpnetworking/database/h2/triggers/ExpressionsUpdateEtagTrigger.java
```diff
@@ -26,6 +26,6 @@
      * Public no args constructor.
      */
     public ExpressionsUpdateEtagTrigger() {
-        super("portal.expressions_etag_seq");
+        super("portal.expressions_etags", 9);
```
vjkoskela: 9?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -318,7 +327,11 @@ private NagiosExtension convertToEbeanNagiosExtension(final models.internal.Nagi
          */
         @Override
         public PagedList<models.ebean.Alert> createAlertQuery(final AlertQuery query) {
+            if (query.getOrganization() == null) {
```
vjkoskela: Should it just be illegal to construct a query without an organization? I would prefer to push that invariant into the definition of a query unless there is a reason it is optional at that level.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -364,9 +377,17 @@ public void saveAlert(final models.ebean.Alert alert) {
          * {@inheritDoc}
          */
         @Override
-        public long getEtag() {
-            final SqlQuery sqlQuery = Ebean.createSqlQuery("SELECT CURRVAL('portal.alerts_etag_seq') AS etag;");
-            return sqlQuery.findUnique().getLong("etag");
+        public long getEtag(final Organization organization) {
+            final SqlQuery sqlQuery = Ebean.createSqlQuery("SELECT etag from portal.alerts_etags where organization = ?;");
```
vjkoskela: I'm not sure I love calling this etags in the database. We're tracking the version of the set of data (e.g. an organization's alerts). Would these be more appropriate in an organization table as a column?

app/com/arpnetworking/metrics/portal/expressions/impl/DatabaseExpressionRepository.java
```diff
@@ -276,7 +288,12 @@ private Expression convertFromEbeanExpression(final models.ebean.Expression ebea
          */
         @Override
         public PagedList<models.ebean.Expression> createExpressionQuery(final ExpressionQuery query) {
+            if (query.getOrganization() == null) {
```
vjkoskela: Same here.

app/com/arpnetworking/metrics/portal/hosts/impl/DatabaseHostRepository.java
```diff
@@ -290,6 +303,9 @@ private void assertIsOpen(final boolean expectedState) {
         @Override
         public PagedList<models.ebean.Host> createHostQuery(final HostQuery query) {
             ExpressionList<models.ebean.Host> ebeanExpressionList = Ebean.find(models.ebean.Host.class).where();
+            if (query.getOrganization() != null) {
```
vjkoskela: Isn't this required?

app/com/arpnetworking/metrics/portal/hosts/impl/ElasticSearchHostRepository.java
```diff
@@ -163,11 +176,13 @@ public void close() {
      * {@inheritDoc}
      */
     @Override
-    public void addOrUpdateHost(final Host host) {
+    public void addOrUpdateHost(final Host host, final Organization organization) {
+        //TODO(barp): Support organizational separation in ElasticSearch [?]
```
vjkoskela: This is fair. Just file an issue.

app/controllers/AlertController.java
```diff
@@ -89,7 +90,7 @@ public Result addOrUpdate() {
         }
 
         try {
-            _alertRepository.addOrUpdateAlert(alert);
+            _alertRepository.addOrUpdateAlert(alert, Organization.DEFAULT);
```
vjkoskela: How come this is default organization?

app/actors/ClusterShutdownActor.java
BrandonArp: Fixed.

app/com/arpnetworking/database/h2/triggers/BaseUpdateEtagTrigger.java
BrandonArp: Changed to using EBean for the accesses.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
BrandonArp: The thing we're representing is an etag.  Etags are not generic versions.  They don't need to increment to be valid (though they do in this case).  I'm not opposed to calling it something else in the database, but it's not just a version.  And as you pointed out, it's a "version" on a view of the alerts.  It's just the current implementation that we only version at the organization level.  For this reason, I'm strongly opposed to putting it in the organization table.  It's not an attribute of an organization.  It's an implementation detail about the alerts or expressions query.

app/controllers/AlertController.java
BrandonArp: Because I didn't plumb the organization through all the APIs, as mentioned in the original comment on the PR.

app/com/arpnetworking/database/h2/triggers/AlertsUpdateEtagTrigger.java
BrandonArp: The column (yes, positional) of the organization field in the table the trigger is on.  That's the best that H2 gives us: an array of new and old values.  I'm not pleased with the mechanism.

app/com/arpnetworking/database/h2/triggers/BaseUpdateEtagTrigger.java
BrandonArp: Fixed

app/com/arpnetworking/database/h2/triggers/BaseUpdateEtagTrigger.java
BrandonArp: Fixed.

app/com/arpnetworking/database/h2/triggers/ExpressionsUpdateEtagTrigger.java
BrandonArp: See above.

app/com/arpnetworking/metrics/portal/hosts/impl/DatabaseHostRepository.java
BrandonArp: Changed to be like the others

app/com/arpnetworking/metrics/portal/hosts/impl/ElasticSearchHostRepository.java
BrandonArp: Done.

app/actors/ClusterShutdownActor.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Actor that subscribes to cluster events and triggers a CompletableFuture when it sees itself removed
+ * from the cluster.
+ *
+ * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
+ */
+public class ClusterShutdownActor extends UntypedActor {
```
vjkoskela: So I see this being registered in Global. What are we still using clustering for? I thought we were trying to remove it from Metrics Portal?

app/com/arpnetworking/database/h2/triggers/AlertsUpdateEtagTrigger.java
vjkoskela: Sorry, I completely missed the classpath on this. That's fine. 

We really need to drop H2 and use Postgres via Docker. Unfortunately, as I recall you said SBT does not have a good Docker plugin ... and I doubt you like what I'm thinking of doing ;)

app/com/arpnetworking/database/h2/triggers/ExpressionsUpdateEtagTrigger.java
vjkoskela: Yup.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
vjkoskela: Sorry I didn't get a chance to call back. I sent you an email on this with some of my thoughts.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
vjkoskela: Hrm. I agree, it's not an attribute of the organization. Should we then have a single versions table? That contains organization, asset type, and version?

app/controllers/AlertController.java
vjkoskela: Doesn't need to be in this PR but I think we should do that before we release.

app/com/arpnetworking/metrics/portal/alerts/AlertRepository.java
```diff
@@ -43,9 +44,10 @@
      * Get the <code>Alert</code> by identifier.
      *
      * @param identifier The <code>Alert</code> identifier.
+     * @param organization The organization owning the alert.
      * @return The matching <code>Alert</code> if found or <code>Optional.empty()</code>.
      */
-    Optional<Alert> get(final UUID identifier);
+    Optional<Alert> get(UUID identifier, Organization organization);
```
vjkoskela: If we don't want to switch completely to a builder model for the query we could add organization to the createQuery APIs.

app/com/arpnetworking/metrics/portal/hosts/impl/ElasticSearchHostRepository.java
```diff
@@ -196,7 +211,8 @@ public void addOrUpdateHost(final Host host) {
      * {@inheritDoc}
      */
     @Override
-    public void deleteHost(final String hostname) {
+    public void deleteHost(final String hostname, final Organization organization) {
+        //TODO(barp): Support organizational separation il ElasticSearch [?]
```
vjkoskela: type-o: il -> in

app/com/arpnetworking/metrics/portal/hosts/impl/LocalHostRepository.java
```diff
@@ -226,8 +231,12 @@ private void assertIsOpen(final boolean expectedState) {
         }
     }
 
+    private Map<String, Host> orgMap(final Organization organization) {
+        return _temporaryStorage.computeIfAbsent(organization.getId().toString(), (v) -> Maps.newConcurrentMap());
+    }
+
     private final AtomicBoolean _isOpen = new AtomicBoolean(false);
-    private final Map<String, Host> _temporaryStorage = Maps.newConcurrentMap();
+    private final Map<String, Map<String, Host>> _temporaryStorage = Maps.newConcurrentMap();
```
vjkoskela: Couldn't this map be key'd by UUID?

app/com/arpnetworking/metrics/portal/hosts/impl/LocalHostRepository.java
```diff
@@ -226,8 +231,12 @@ private void assertIsOpen(final boolean expectedState) {
         }
     }
 
+    private Map<String, Host> orgMap(final Organization organization) {
```
vjkoskela: rename: orgMap -> getOrganizationMap (e.g.)

app/actors/ClusterShutdownActor.java
```diff
@@ -0,0 +1,75 @@
+/**
+ * Copyright 2016 Inscope Metrics
```
vjkoskela: Please fix.

app/models/ebean/AlertEtags.java
```diff
@@ -0,0 +1,112 @@
...
+    @Column(nullable = false)
+    private int etag;
+
+    /**
+     * Increments an etag record or creates one if it does not exist.
+     *
+     * @param organization the organization
+     */
+    public static void incrementEtag(final Organization organization) {
+        AlertEtags etag = FINDER.setForUpdate(true)
```
vjkoskela: final

app/models/ebean/AlertEtags.java
```diff
@@ -0,0 +1,112 @@
...
+                .where()
+                .eq("organization", organization)
+                .findUnique();
+        if (etag == null) {
+            etag = new AlertEtags();
+            etag.setOrganization(organization);
+            etag.setEtag(1);
```
vjkoskela: Should it be zero? (match default value of get)

app/models/ebean/AlertEtags.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public void setOrganization(final Organization value) {
+        organization = value;
+    }
+
+    public int getEtag() {
+        return etag;
+    }
+
+    public void setEtag(final int value) {
```
vjkoskela: Long?

app/models/ebean/ExpressionEtags.java
```diff
@@ -0,0 +1,112 @@
...
+    @Column(nullable = false)
+    private int etag;
+
+    /**
+     * Increments an etag record or creates one if it does not exist.
+     *
+     * @param organization the organization
+     */
+    public static void incrementEtag(final Organization organization) {
+        ExpressionEtags etag = FINDER.setForUpdate(true)
```
vjkoskela: final

app/models/ebean/ExpressionEtags.java
```diff
@@ -0,0 +1,112 @@
...
+                .where()
+                .eq("organization", organization)
+                .findUnique();
+        if (etag == null) {
+            etag = new ExpressionEtags();
+            etag.setOrganization(organization);
+            etag.setEtag(1);
```
vjkoskela: zero?

app/models/ebean/ExpressionEtags.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public Organization getOrganization() {
+        return organization;
+    }
+
+    public void setOrganization(final Organization value) {
+        organization = value;
+    }
+
+    public int getEtag() {
```
vjkoskela: long?

app/models/ebean/ExpressionEtags.java
```diff
@@ -0,0 +1,112 @@
...
+        return etag;
+    }
+
+    public void setEtag(final int value) {
+        etag = value;
+    }
+
+    private static final Finder<Long, ExpressionEtags> FINDER = new Finder<>(ExpressionEtags.class);
+}
+// CHECKSTYLE.ON: MemberNameCheck
```
vjkoskela: I'm starting to see a pattern here. Would it make more sense to have a single table with a resource type column? This would make it easier to add new resources without changing the schema in the future.

app/models/ebean/Organization.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @CreatedTimestamp
+    @Column(name = "created_at")
+    private Timestamp createdAt;
+
+    @UpdatedTimestamp
+    @Column(name = "updated_at")
+    private Timestamp updatedAt;
+
+    @Column(name = "name")
```
vjkoskela: So this is what concerns me. How many data are we mastering in MP about the organization. I don't think we should master any and it should just be a foreign key definition where the actual data can live somewhere else. Thoughts?

app/models/ebean/Organization.java
```diff
@@ -0,0 +1,143 @@
...
+
+    public void setId(final Long value) {
+        id = value;
+    }
+
+    public Long getVersion() {
+        return version;
+    }
+
+    public void setVersion(final Long value) {
```
vjkoskela: If we're using long version on the entities we should definitely use it on the etag table too.

app/models/internal/HostQuery.java
```diff
@@ -34,6 +34,14 @@
     HostQuery partialHostname(final Optional<String> partialHostname);
 
     /**
+     * Set the organization to query for. Required. Cannot be null.
+     *
+     * @param organization The partial or complete hostname to match.
```
vjkoskela: copy-paste comment error

app/models/internal/AlertQuery.java
```diff
@@ -25,6 +25,14 @@
 public interface AlertQuery {
 
     /**
+     * Set the organization. Required. Cannot be null.
+     *
+     * @param organization The organization.
+     * @return This instance of <code>ExpressionQuery</code>.
```
vjkoskela: copy-paste comment error

app/models/internal/Organization.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+package models.internal;
+
+import models.internal.impl.DefaultOrganization;
+
+import java.util.UUID;
+
+/**
+ * @author Brandon Arp (brandon dot arp at smartsheet dot com)
```
vjkoskela: Missing class javadoc.

app/models/internal/impl/DefaultExpressionQuery.java
```diff
@@ -152,6 +170,7 @@ public String toString() {
     }
 
     private final ExpressionRepository _repository;
+    private Organization _organization;
```
vjkoskela: So we can make this final if it's accepted via the constructor and via the repository createQuery API. That's probably the simplest way forward. Refactoring to a builder model would work too. If there will be other required fields sticking with this pattern will eventually mean undoing it. However, I'm not sure I can see a lot (any?) other required fields on the query.

app/models/internal/impl/DefaultOrganization.java
```diff
@@ -0,0 +1,142 @@
...
+        if (other == this) {
+            return true;
+        }
+
+        if (!(other instanceof DefaultOrganization)) {
+            return false;
+        }
+
+        final DefaultOrganization otherOrg = (DefaultOrganization) other;
+        return Objects.equal(_name, otherOrg._name);
```
vjkoskela: Shouldn't this on id?

app/models/internal/impl/DefaultOrganization.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int hashCode() {
+        return Objects.hashCode(_name);
```
vjkoskela: Same.

app/models/internal/impl/DefaultOrganization.java
```diff
@@ -0,0 +1,142 @@
...
+    private DefaultOrganization(final Builder builder) {
+        _name = builder._name;
+        _id = builder._id;
+    }
+
+    private final String _name;
+    private final UUID _id;
+
+    /**
+     * Implementation of builder pattern for <code>DefaultHost</code>.
```
vjkoskela: Copy-paste comment error.

app/models/internal/impl/DefaultOrganization.java
```diff
@@ -0,0 +1,142 @@
...
+     *
+     * @author Ville Koskela (ville dot koskela at inscopemetrics dot com)
```
vjkoskela: I don't recall writing this ;)

conf/db/migration/metrics_portal_ddl/common/V7__add_organization_part_1.sql
```diff
@@ -0,0 +1,71 @@
...
+  updated_at TIMESTAMP NOT NULL DEFAULT now(),
+  name VARCHAR(255) NOT NULL
+);
+
+CREATE UNIQUE INDEX organizations_uuid_idx ON portal.organizations (uuid);
+
+INSERT INTO portal.organizations (uuid, name) VALUES ('0eb03110-2a36-4cb1-861f-7375afc98b9b', 'Default Org');
+
+ALTER TABLE portal.hosts ADD COLUMN organization INTEGER DEFAULT NULL;
+/*ALTER TABLE portal.hosts ADD COLUMN organization INTEGER REFERENCES portal.organizations(id) DEFAULT NULL;*/
```
vjkoskela: ?

conf/portal.application.conf
```diff
@@ -187,14 +187,14 @@ hostProvider {
 # Alerts
 # ~~~~~
 alertRepository {
-  type = com.arpnetworking.metrics.portal.alerts.impl.NoAlertRepository
+  type = com.arpnetworking.metrics.portal.alerts.impl.DatabaseAlertRepository
```
vjkoskela: These features are off by default I believe. Hence no db.

vjkoskela:
Ok. Completed the feedback.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
BrandonArp: For those not on the email, the options are refactoring to a builder pattern or putting the organization in the constructor.  Given that both of these are X changes, we're free to do either one.  Our general preference is to use builders because we can easily add validation rules with the OVal framework.  However, in this case, a completely "empty" query was acceptable (unlike with the builders).  This is the first and the only foreseeable required member.  So, I think I'm going to add it as a constructor arg and hope that we don't see any more required args.

app/controllers/AlertController.java
BrandonArp: Agreed.

app/actors/ClusterShutdownActor.java
BrandonArp: The active host updaters are cluster singletons.  Also might use cluster sharding (maybe?) for preliminary alert evaluation.  We had cluster shutdown before, this is just nicer.

app/com/arpnetworking/metrics/portal/hosts/impl/ElasticSearchHostRepository.java
BrandonArp: Fixed.

app/com/arpnetworking/metrics/portal/alerts/AlertRepository.java
BrandonArp: I took a shot at this.  Ie, keeping organization out of the query args and keeping it as arguments on the methods.  Let me know what you think.

app/com/arpnetworking/metrics/portal/hosts/impl/LocalHostRepository.java
BrandonArp: Yep, much better, too.

app/com/arpnetworking/metrics/portal/hosts/impl/LocalHostRepository.java
BrandonArp: Sure...

app/actors/ClusterShutdownActor.java
BrandonArp: Fixed

app/models/ebean/AlertEtags.java
BrandonArp: Can't be final, we re-assign.

app/models/ebean/AlertEtags.java
BrandonArp: Nope.  When you're updating th etag, it means that something has changed.  We really really want this to be different.

app/models/ebean/AlertEtags.java
BrandonArp: Sure

app/models/ebean/ExpressionEtags.java
BrandonArp: Can't, gets reassigned.

app/models/ebean/ExpressionEtags.java
BrandonArp: See above

app/models/ebean/ExpressionEtags.java
BrandonArp: Done

app/models/ebean/ExpressionEtags.java
BrandonArp: Possibly.  That'd be a great refactoring for the 3rd time we do it ( ;  Seriously, though, I'd be OK doing it here, but I don't know if we want to keep this pattern going forward.  I'd like to talk to you more about etags in general first.

app/models/ebean/Organization.java
BrandonArp: I wanted the name for minimal branding.  IE, something to show what org you were in.  I don't want to master information here, but I think we need some way to tell the user which org they're seeing.

app/models/ebean/Organization.java
BrandonArp: Yeah yeah yeah, you've mentioned.

app/models/internal/HostQuery.java
BrandonArp: Removed

app/models/internal/AlertQuery.java
BrandonArp: Removed

app/models/internal/Organization.java
BrandonArp: Fixed.

app/models/internal/impl/DefaultExpressionQuery.java
BrandonArp: Added to methods instead.

app/models/internal/impl/DefaultOrganization.java
BrandonArp: Yup

app/models/internal/impl/DefaultOrganization.java
BrandonArp: Yup

app/models/internal/impl/DefaultOrganization.java
BrandonArp: Fixed

app/models/internal/impl/DefaultOrganization.java
BrandonArp: No? Huh....

conf/db/migration/metrics_portal_ddl/common/V7__add_organization_part_1.sql
BrandonArp: Removed.

conf/portal.application.conf
BrandonArp: Should we not encourage people to use them now?  Should I also turn on those feature flags?

BrandonArp:
I've updated the PR, but there are still a couple things to do:

- [ ] Review moving organization to the host query methods.  If you like it, I'll do the same on Alerts and Expressions.  Just didn't want to waste time if you weren't a fan.
- [ ] Change the version fields of pre-existing tables to BIGINT"?  They are INTEGER right now with a 'long' type on the bean.

conf/portal.application.conf
vjkoskela: My preference at the moment on both counts is no. Alerts one could argue, but they tend to be very organization specific. Expressions just plain don't work (yet).

app/models/ebean/Organization.java
vjkoskela: That's a very good point. However, I am not convinced it should come from this table. In fact, I think that relates more to how we pass session, authN and authZ information to the MP from whatever system is in front of it. 

Then I can see there be several ways to obtain this information. One could be from another table (e.g. organizationdetails; if the data _is_ mastered here). Another could be from the session and a third from a remote service (or other database). 

So, in my opinion the retrieval mechanism for organization details should be separate from the foreign key support we're building. Second, we should plan how to consume that information on MP to show things like what organization you are in.

However, I think adding name in this PR is putting the cart before the horse.

app/models/ebean/ExpressionEtags.java
vjkoskela: I would actually use the exact same argument to argue for a single table so it's less to rip out when we realize how wrong it is ;)

Sure, let's chat about this.

app/models/ebean/Organization.java
BrandonArp: One important point: MP should stand on its own.  We need to consider the OSS deployment scenarios.

vjkoskela:
@BrandonArp let me know when this is ready for final review.

BrandonArp:
Back over to you.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -160,17 +163,19 @@ public AlertQuery createQuery() {
...
+                .eq("organization.uuid", organization.getId())
                 .findRowCount();
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
-    public void addOrUpdateAlert(final Alert alert) {
+    public void addOrUpdateAlert(final Alert alert, final Organization organization) {
```
vjkoskela: You should add the organization to the debug statement below.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -104,16 +105,18 @@ public void close() {
      * {@inheritDoc}
      */
     @Override
-    public Optional<Alert> get(final UUID identifier) {
+    public Optional<Alert> get(final UUID identifier, final Organization organization) {
         assertIsOpen();
         LOGGER.debug()
                 .setMessage("Getting alert")
                 .addData("alertId", identifier)
+                .addData("organization.uuid", organization.getId())
```
vjkoskela: Should we just add the entire organization to the debug statement?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
```diff
@@ -125,10 +128,10 @@ public void close() {
      * {@inheritDoc}
      */
     @Override
-    public AlertQuery createQuery() {
+    public AlertQuery createQuery(final Organization organization) {
         assertIsOpen();
         LOGGER.debug().setMessage("Preparing query").log();
```
vjkoskela: Should we include organization here too?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertRepository.java
vjkoskela: There are two more log statements farther down.

app/com/arpnetworking/metrics/portal/alerts/impl/NoAlertRepository.java
```diff
@@ -68,7 +69,7 @@ public void close() {
      * {@inheritDoc}
      */
     @Override
-    public Optional<Alert> get(final UUID identifier) {
+    public Optional<Alert> get(final UUID identifier, final Organization organization) {
         assertIsOpen();
         LOGGER.debug()
                 .setMessage("Getting alert")
```
vjkoskela: I think this is probably the last item. Add organization to all the relevant log statements.

BrandonArp:
Back to you one last time.

PR #71, opened 2016-10-02
Improve foreman host provider logging.

vjkoskela:


app/com/arpnetworking/metrics/portal/hosts/impl/ForemanHostProvider.java
```diff
@@ -83,12 +84,22 @@ public void onReceive(final Object message) throws Exception {
             }
 
             if (response.getTotal() > response.getPage() * response.getPerPage()) {
-                PatternsCS.pipe(_client.getHostPage(response.getPage() + 1, response.getPerPage()), context().dispatcher())
+                PatternsCS.pipe(
+                        _client.getHostPage(response.getPage() + 1, response.getPerPage()),
+                        context().dispatcher())
```
BrandonArp: This formatting is wrong.

vjkoskela: So it's due to chaining after the first call. There are few ways to approach this.

1) Indent first call. This would keep the style consistent with the rest.

```
PatternsCS
    .pipe(
        _client ...)
    .to(self(), self());
```

1) Drop parenthesis to the next line. This forces the next call to the correct level. We used a derivative of this style at Amazon where both parenthesis were on separate lines which allowed to line them up visually.

```
PatternsCS.pipe(
    _client ...
)
.to(self(), self());
```

Or as describe with both parenthesis on new lines:

```
PatternsCS.pipe
(
    _client ...
)
.to(self(), self());
```

This last format scales really nicely but is not compatible with our existing style.

PR #73, opened 2016-10-13
Added client for Consul and matching HostProvider implementation.

vjkoskela:


app/com/arpnetworking/metrics/portal/hosts/impl/ConsulClient.java
```diff
@@ -0,0 +1,189 @@
...
+        return _client
+                .url(_baseUrl + "/v1/catalog/nodes")
+                .get()
+                .thenApply(this::parseWSResponse);
+    }
+
+    private ImmutableList<Host> parseWSResponse(final WSResponse response) {
+        try {
+            if (response.getStatus() / 100 != 2) {
+                throw new IOException("Non-200 response from Consul");
```
BrandonArp: Would be nice to know what the response code is.

BrandonArp: Also, you throw IOException just to catch it and throw it again as a runtime exception. I do not like that at all.

vjkoskela: Well, I'll give you one guess as to who's code I copied. Fixing...

Issue #76, opened 2016-11-20
Injectable HealthProvider


vjkoskela:
Refactor the static health check in MetaController into an injected HealthProvider. This will enable wrapper projects to expose the health check by whatever organization specific endpoints are required.

Changes required include:
 - HealthProvider interface
 - Default healthy implementation
 - Configuration defined health provider class
 - Module changes to set configured health provider in Guice
 - Changes to MetaController to accept and use HealthProvider

vjkoskela:
This was completed with #77

PR #77, opened 2016-11-21
Injectable Health Provider

vjkoskela:
Refactor health check into configurable and injectable health provider interface and default implementation.

This addresses #76

### DIFF
--- a/app/global/MainModule.java
+++ b/app/global/MainModule.java
@@ -29,6 +29,7 @@ import com.arpnetworking.metrics.impl.TsdLogSink;
 import com.arpnetworking.metrics.impl.TsdMetricsFactory;
 import com.arpnetworking.metrics.portal.alerts.AlertRepository;
 import com.arpnetworking.metrics.portal.expressions.ExpressionRepository;
+import com.arpnetworking.metrics.portal.health.HealthProvider;
 import com.arpnetworking.metrics.portal.hosts.HostRepository;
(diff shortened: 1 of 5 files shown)


app/global/MainModule.java
```diff
@@ -84,6 +86,21 @@ protected void configure() {
...
     @Singleton
+    @Provides
+    @SuppressFBWarnings("UPM_UNCALLED_PRIVATE_METHOD") // Invoked reflectively by Guice
+    private HealthProvider getHealthProvider(final Environment environment, final Configuration config) {
+        try {
+            final Class<? extends HealthProvider> clazz = ConfigurationHelper.getType(
+                    environment,
+                    config,
+                    "http.healthProvider");
+            return clazz.newInstance();
```
BrandonArp: You're going to need to support parameters on the object here.  You'll probably also want it dependency injected.  I recommend you take a look at what I did in Artemis: https://github.com/groupon/artemis/blob/master/app/global/ProdModule.java#L100-L120

vjkoskela: Actually considered this and figured it could be added later. I'm actually not looking to replace the health check logic, but rather the endpoint from which it is served and what format it takes.

Anyway, your example doesn't seem to fit this case. I've put something together along the same lines as other providers using an injected injector. This delegates back to Guice the construction of the configured instance allowing its constructor to be marked as @Inject.

I'll post it shortly.

vjkoskela:
@BrandonArp updated.

PR #83, opened 2017-04-10
Switch from file based metrics publication to http publication.

vjkoskela:


conf/logback.xml
```diff
@@ -21,7 +21,7 @@
   <!-- IMPORTANT: This logger should only used when debugging and testing -->
   <appender name="MetricsPortalLogger" class="ch.qos.logback.core.rolling.RollingFileAppender">
     <file>logs/metrics-portal.log</file>
-    <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
```
BrandonArp: This doesn't seem right.  Won't the %i in the fileNamePattern mess this up?

vjkoskela: Unfortunately, this is where there is some poor decoupling in the logback model in my opinion. I think it should be `SizeAndTimeBasedRollingPolicy` but if you do this it doesn't work right. This is what we use in MAD:

https://github.com/ArpNetworking/metrics-aggregator-daemon/blob/master/config/logback.xml

I think we need to address the implementation of the trigger policy to make it work correctly with logback -- or better yet if they ever fix their trigger policy just get rid of ours.

PR #84, opened 2017-05-09
use relative paths to allow for a path prefix

BrandonArp:
I looked for other paths, but didn't find any.  Hopefully this is all of them.


vjkoskela:
Relative paths are good. Can you elaborate on what this is for?

BrandonArp:
This will allow proxying with a prefix.

PR #90, opened 2017-08-21
Add basic units support.

vjkoskela:
This is really basic plumbing of single unit metrics to the user interface.


conf/portal.application.conf
```diff
@@ -33,6 +33,10 @@ play.modules.enabled += "global.MainModule"
 play.http.actionCreator = "global.ActionCreator"
 play.http.errorHandler = "global.ErrorHandler"
 
+# Http Server
+# ~~~~~
+play.server.http.port = 8080
```
BrandonArp: Uh... this is a breaking change. Please don't do this here.

vjkoskela: Sure I can pull this out. But we need to do this. Port 9000 is a terrible choice ;)

PR #97, opened 2017-11-15
rework typeahead

BrandonArp:
fix the typeahead knockout binding
fix header use of typeahead


vjkoskela:
I trust you've tested it; no problems here.

PR #101, opened 2018-01-17
make proxying more play 2.6 friendly

BrandonArp:


vjkoskela:
Looks reasonable to me.

PR #102, opened 2018-01-17
Build fixes

BrandonArp:
disable rjs due to js version and cdn references
fix pid file
add universal plugin


vjkoskela:
I'm going to trust you that this works.

PR #106, opened 2018-02-06
Add feedback and questions

scottw-smartsheet:
Just some questions.


BrandonArp:
Addressed the concerns in the original commit.

PR #120, opened 2018-08-14
fix test failure in akka networking config

BrandonArp:


BrandonArp:
@vjkoskela Please review.

PR #121, opened 2018-08-15
filter the transfer encoding since we are rewriting it

BrandonArp:


BrandonArp:
@vjkoskela Please review

PR #126, opened 2018-11-10
add docker login and publish

BrandonArp:


vjkoskela:
This needs to be reworked now that the project is on Maven. Do you have a few minutes to push the secret references into `settings.xml`?

PR #129, opened 2018-11-29
Reports

speezepearson:


app/com/arpnetworking/metrics/portal/reports/EmailBuilder.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.mail.BodyPart;
+import javax.mail.Message;
+import javax.mail.MessagingException;
+import javax.mail.Session;
+import javax.mail.internet.MimeBodyPart;
+import javax.mail.internet.MimeMessage;
+import javax.mail.internet.MimeMultipart;
+
+
+public class EmailBuilder {
```
vjkoskela: A builder typically refers to a type of class that constructs instances of another class. While it obviously can have other meanings, I would prefer to avoid that word here, especially given how heavily this codebase uses the Builder design pattern.

app/com/arpnetworking/metrics/portal/reports/Scraper.java
```diff
@@ -0,0 +1,184 @@
...
+import com.github.kklisura.cdt.launch.ChromeLauncher;
+import com.github.kklisura.cdt.protocol.commands.Page;
+import com.github.kklisura.cdt.services.ChromeDevToolsService;
+import com.github.kklisura.cdt.services.ChromeService;
+import com.github.kklisura.cdt.services.types.ChromeTab;
+import java.io.PrintWriter;
+
+
+public class Scraper {
+    public static void main(String[] args) throws Exception {
```
vjkoskela: This type of testing code should be removed from the PR.

vjkoskela:
General notes:

* Please add unit tests
* Design the interfaces 

= Tests =

Most of the tests should be possible with appropriate interfaces (see below) and mocking. However, for some of the more integration type tests you should probably use WireMock for testing obtaining HTML from a web server. Not sure what the best way is to test the email send.

= Interfaces =

1) ReportSink : This should be an interface for sending a rendered report to a destination. For example, implementations can include Email, Slack, S3, Dropbox. The actual implementation (e.g. `EmailReportSink`) should probably be constructed using a builder.

2) ReportGenerator : This should be an interface for generating reports. There will need to be a `Report` data class that encapsulates the generated report. We should probably have different implementations of this (possibly with a shared base or helper class) for generating HTML and PDF reports.

The idea is that we can schedule a report, we can generate a report in one or more types (e.g. HTML, PDF) and we can send a report on more or more mediums (e.g. Email, Slack, etc.).

PR #132, opened 2018-12-08
make sure to shutdown cassandra sessions

BrandonArp:
Clean up the Cassandra shutdown as well as let Play shutdown the ActorSystem properly.  This addresses a race condition in shutdown and notification of cluster membership leaving that causes hangs in Travis.

### DIFF
--- a/app/global/Global.java
+++ b/app/global/Global.java
@@ -16,18 +16,12 @@
 
 package global;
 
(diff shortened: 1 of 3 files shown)


BrandonArp:
Merging with approval from @vjkoskela.

PR #134, opened 2018-12-15
Implement a SQL-backed storage layer for Reporting

cwbriones:
This currently does not map the data model to the internal model, so the interfaces expose the data models right now (but nothing uses them).

Code is mostly boilerplate and supports CRUD operations on a `ReportingJob` instance.

There is currently no notion of an execution history as we discussed, but I figured I can implement that some time after this to keep down the PR size.


app/com/arpnetworking/metrics/portal/reports/ReportingJobRepository.java
```diff
@@ -0,0 +1,63 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.arpnetworking.metrics.portal.reports;
+
+// TODO(cwbriones): Decouple from ebean model and expose via the internal model.
```
vjkoskela: Would it be possible to include the interfaces for the internal model in this PR and adjust the code to use those?

app/com/arpnetworking/metrics/portal/reports/ReportingJobRepository.java
```diff
@@ -0,0 +1,63 @@
...
+import models.ebean.ReportingJob;
+
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * A repository for storing and retrieving instances of <code>ReportingJob</code>.
```
vjkoskela: Ideally, we should use `{@link ...}` or `{@code ...}`. The use of `<code>...</code>` is generally frowned upon. Unfortunately, this project has not been updated.

app/models/ebean/ChromeScreenshotReportSource.java
```diff
@@ -0,0 +1,102 @@
...
+import javax.persistence.Entity;
+
+/**
+ * Data Model for SQL storage of a chrome screenshot based report generation scheme.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+// CHECKSTYLE.OFF: MemberNameCheck
+@Entity
+@DiscriminatorValue("chromeScreenshot")
```
vjkoskela: This is the "enumerated" type descriptor? Should we use capital snake case to link it to such?

app/models/ebean/ChromeScreenshotReportSource.java
```diff
@@ -0,0 +1,102 @@
...
+    @Column(name = "title")
+    private String title;
+
+    @Column(name = "ignore_certificate_errors")
+    private boolean ignoreCertificateErrors;
+
+    @Column(name = "triggering_event_name")
+    private String triggeringEventName;
+
+    @Column(name = "pdf_width_inches")
```
vjkoskela: Why does the report source contain format specific flags?

app/models/ebean/ReportRecipient.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private ReportRecipient(final RecipientType typeValue, final String recipientValue) {
+        type = typeValue;
+        recipient = recipientValue;
+    }
+
+    /**
+     * Create a new ReportRecipient with the given emailAddress.
+     * @param emailAddress The address of the recipient
```
vjkoskela: nit: blank line between description and tags.

app/models/ebean/ReportRecipient.java
```diff
@@ -0,0 +1,128 @@
...
+     * @return A new email recipient.
+     */
+    public static ReportRecipient newEmailRecipient(final String emailAddress) {
+        return new ReportRecipient(RecipientType.EMAIL, emailAddress);
+    }
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @Column(name = "id")
+    private Long id;
```
vjkoskela: Existing conventions would have the fields at the top, followed by methods, followed by subtypes.

app/models/ebean/ReportRecipient.java
```diff
@@ -0,0 +1,128 @@
...
+        return recipient;
+    }
+
+    public RecipientType getType() {
+        return type;
+    }
+
+    @Override
+    public String toString() {
+        return "ReportRecipient{"
```
vjkoskela: Use `MoreObjects` to string helper.

app/models/ebean/ReportRecipient.java
```diff
@@ -0,0 +1,128 @@
...
+                + ", type=" + type
+                + '}';
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
```
vjkoskela: IMO you should use `.equals` for the class equality.

vjkoskela: Alternatively, marking this class `final` and using `instanceof` works nicely here since you don't need a null check.

app/models/ebean/ReportRecipientGroup.java
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Data Model for SQL storage of groups of report recipients.
+ * <p>
+ * A recipient group can contain multiple {@link models.ebean.ReportRecipient} instances.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+@Entity
+@Table(name = "report_recipient_groups", schema = "portal")
+// CHECKSTYLE.OFF: MemberNameCheck
```
vjkoskela: The class above has `// CHECKSTYLE.ON: FinalClassCheck`,  but this does not.

app/models/ebean/ReportRecipient.java
```diff
@@ -0,0 +1,128 @@
...
+    public Long getId() {
+        return id;
+    }
+
+    /**
+     * Get the address of this recipient.
+     * @return The address of the recipient.
+     */
+    public String get() {
+        return recipient;
```
vjkoskela: Are there any cases where this is insufficient? Consider common targets such as Slack, S3, Dropbox, etc.

vjkoskela: Should this be discriminated and "polymorphic" as well?

app/models/ebean/ReportSource.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public void setUuid(final UUID value) {
+        uuid = value;
+    }
+
+    public long getTimeoutInSeconds() {
+        return timeoutInSeconds;
+    }
+
+    public void setTimeoutInSeconds(final long value) {
```
vjkoskela: So the assertion is that all report types have a timeout?

app/models/ebean/ReportingSchedule.java
```diff
@@ -0,0 +1,96 @@
...
+        WEEKLY
+    }
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @Column(name = "id")
+    private Integer id;
+
+    @Column(name = "send_at")
+    private Timestamp sendAt;
```
vjkoskela: Do we want to give the user this much control over scheduling?

app/models/ebean/ReportingSchedule.java
```diff
@@ -0,0 +1,96 @@
...
+         */
+        NONE,
+        /**
+         * A schedule which recurs daily.
+         */
+        DAILY,
+        /**
+         * A schedule which recurs weekly.
+         */
+        WEEKLY
```
vjkoskela: So this union structure will capture all data for all types of schedules? e.g. third Monday of month, weekly from date x until date y, daily from date x for y occurrences, etc.

vjkoskela: Will this table also be updated with data necessary to evaluate the schedule? Or are we going to obtain that elsewhere? (e.g. most recently sent date)

conf/db/migration/metrics_portal_ddl/common/V11__create_reporting_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+CREATE TABLE portal.report_sources (
+    id BIGSERIAL PRIMARY KEY,
+    uuid UUID NOT NULL,
+    version INTEGER NOT NULL DEFAULT 1,
+    created_at TIMESTAMP NOT NULL DEFAULT now(),
+    updated_at TIMESTAMP NOT NULL DEFAULT now(),
+    timeout_seconds INTEGER NOT NULL,
+    type VARCHAR(255) NOT NULL,
+
+    -- Chrome
```
vjkoskela: Is this the actual value of `type` for when these fields apply?

conf/db/migration/metrics_portal_ddl/common/V11__create_reporting_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+    url VARCHAR(255),
+    title VARCHAR(255) DEFAULT '',
+    ignore_certificate_errors BIT DEFAULT 0,
+    triggering_event_name VARCHAR(255),
+    pdf_width_inches FLOAT,
+    pdf_height_inches FLOAT,
```
vjkoskela: Thoughts or patterns for avoiding column collisions between types?

conf/db/migration/metrics_portal_ddl/common/V11__create_reporting_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+);
+
+CREATE UNIQUE INDEX report_sources_uuid_idx ON portal.report_sources (uuid);
+
+CREATE TABLE portal.report_recipient_groups (
+    id BIGSERIAL PRIMARY KEY,
+    uuid UUID NOT NULL,
+    version INTEGER NOT NULL DEFAULT 1,
+    name VARCHAR(255) DEFAULT '',
```
vjkoskela: So is the idea to have full on group management? Or does each report create its own "group" (just wondering what the role of name is)

conf/db/migration/metrics_portal_ddl/common/V11__create_reporting_tables.sql
```diff
@@ -0,0 +1,73 @@
...
+    created_at TIMESTAMP NOT NULL DEFAULT now(),
+    updated_at TIMESTAMP NOT NULL DEFAULT now(),
+);
+
+CREATE UNIQUE INDEX report_recipient_groups_uuid_idx ON portal.report_recipient_groups (uuid);
+CREATE UNIQUE INDEX report_recipient_groups_name_idx ON portal.report_recipient_groups (name);
+
+CREATE TABLE portal.report_recipients (
```
vjkoskela: should this be `report_group_members`?

test/com/arpnetworking/metrics/portal/TestBeanFactory.java
```diff
@@ -44,6 +51,49 @@
  */
 public final class TestBeanFactory {
 
+    private static final String TEST_EMAIL = "noreply+email-recipient@test.com";
```
vjkoskela: All these should be randomly generated or at least randomly selected from sets where generation is not possible.

app/models/ebean/ChromeScreenshotReportSource.java
cwbriones: I'll pull this out into formats for the report jobs as per discussion.

conf/db/migration/metrics_portal_ddl/common/V11__create_reporting_tables.sql
cwbriones: No, that was more for my convenience of tracking which field belonged to which instance. I can rename it though to line up with the discriminator value.

conf/db/migration/metrics_portal_ddl/common/V11__create_reporting_tables.sql
cwbriones: The name might not be precise enough, I wanted to encapsulate who the report is sent to, rather than storing it as a "Sink" since that concept only makes sense to me at runtime. 

Really all this is trying to be is a set of recipients for a single report, with all of the individual addresses in an umbrella underneath it.

app/models/ebean/ReportRecipient.java
cwbriones: This was autogenerated by IntelliJ, maybe I can configure what it actually spits out? I'll still update this though.

app/models/ebean/ReportRecipientGroup.java
cwbriones: Noted, although I'm not sure why Checkstyle isn't complaining about this instance.

app/models/ebean/ReportSource.java
cwbriones: Yes. This can arguably moved to the actual job itself, but my thought was you should be able to specify a limit on how long the execution can run. Now that I think about it though, this is another hook in which external users could then influence the scheduler so maybe this is better left off as an internal detail.

conf/db/migration/metrics_portal_ddl/common/V11__create_reporting_tables.sql
cwbriones: I don't like this either; I wanted to store the properties of the subclasses in their own tables to avoid this problem but Ebean only allows `InheritanceType.SINGLE_TABLE`. I'll give this some more thought, I also don't like that I can't enforce NONNULL with this strategy.

app/models/ebean/ReportingSchedule.java
cwbriones: > So this union structure will capture all data for all types of schedules? e.g. third Monday of month, weekly from date x until date y, daily from date x for y occurrences, etc.

I updated this to be polymorphic in the type, from a quick run-through I think that should be sufficient to express any of the complex logic that we might want. (e.g. 3rd monday of the month, weekly on M W F)

> Will this table also be updated with data necessary to evaluate the schedule?

Job executions will be in a separate table, that data could be pulled out from there.

app/models/ebean/ReportRecipient.java
cwbriones: I think this should be sufficient, each of those is described by a URI and customization beyond that wouldn't be a requirement of the sink. (Slack - channel to post to, S3 - bucket, Dropbox - filepath)

I suppose there might be other parameters we might want to associate with it (e.g. dropbox uploads with tags, slack posts with a custom message along with the content itself)

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportRepository.java
```diff
@@ -0,0 +1,423 @@
...
+
+    @Override
+    public Optional<Instant> getLastRun(final UUID reportId, final Organization organization) throws NoSuchElementException {
+        return Ebean.find(ReportExecution.class)
+                .orderBy()
+                .desc("completed_at")
+                .where()
+                .eq("report.uuid", reportId)
+                .eq("report.organization.uuid", organization.getId())
+                .eq("state", ReportExecution.State.SUCCESS)
```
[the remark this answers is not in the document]
cwbriones: Ebean uses the `@SoftDelete` annotation to mark a field as soft-deleted and automatically filters those out by default unless you include `includeSoftDeleted()` in your query.

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportRepository.java
```diff
@@ -0,0 +1,423 @@
...
+                scheduled,
+                ReportExecution.State.STARTED,
+                execution -> execution.setStartedAt(Instant.now())
+        );
+    }
+
+    @Override
+    public void jobSucceeded(final UUID reportId, final Organization organization, final Instant scheduled, final Report.Result result) {
+        assertIsOpen();
+        if (!(result instanceof DefaultReportResult)) {
```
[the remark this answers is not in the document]
cwbriones: You're right, I've addressed this by configuring the Result with a `JsonTypeInfo` annotation.

cwbriones:
I've gone ahead and addressed most of your feedback, one question about groups:

We've discussed and settled upon the idea that a group can have multiple types of recipients, that is, the internal model should be pretty much the same as the data model here. But the internal model as it stands separates groups into homogenous collections, (e.g. `EmailRecipientGroup`) Should we just leave this as-is for now and update later on? It would be less than ideal but there's quite a bit to untangle there if we do want to change that.

As far as recipient encoding, I still think URIs should suffice. Slack is a channel id + a workspace, and this is already encoded in their URLs. Even if not, I don't think it would be much effort to move this to a JSON encoding later on.

cwbriones:
@speezepearson - I think we discussed having `NeverSchedule` be a testing-only type? I don't think we'd want users creating actual production reports that never run.

app/models/ebean/Recipient.java
```diff
@@ -0,0 +1,192 @@
...
+
+    @CreatedTimestamp
+    @Column(name = "created_at")
+    private Timestamp createdAt;
+
+    @UpdatedTimestamp
+    @Column(name = "updated_at")
+    private Timestamp updatedAt;
+
+    @OneToMany(
```
vjkoskela: This should not be necessary and couples generic recipients to all consumptions of them (e.g. reports, alerts, etc.)

cwbriones:
rebased with master.

PR #141, opened 2019-01-08
add Job and JobRepository interfaces

speezepearson:


app/com/arpnetworking/metrics/portal/scheduling/JobRepository.java
```diff
@@ -0,0 +1,65 @@
...
+    Optional<Job> getJob(UUID id);
+
+    /**
+     * Get the last time that a job with a given UUID was run.
+     *
+     * @param id The id assigned to the Job by a previous call to {@code add}.
+     * @return The last time that that job was executed.
+     * @throws NoSuchElementException if no job has the given UUID.
+     */
+    Optional<Instant> getLastRun(UUID id) throws NoSuchElementException;
```
speezepearson: Using an `Instant` here instead of a `ZonedDateTime` because it's nonsensical to talk about "the time zone in which the job was executed." This will result in needing some extra glue-code when we're computing the next time a Job should be executed: I feel strongly that it's worth it to avoid a proliferation of time zones that might get out of sync with each other and result in subtle bugs that are easy to forget to test.

vjkoskela: That's reasonable. Although since the schedule produces times with locale, and the job executes on that, the last time it executed could be easily expressed in the same way. However, your interpretation is strictly more correct. This may impact Christian's storage to ensure that the last run storage round-trip is not lossy.

app/com/arpnetworking/metrics/portal/scheduling/JobRepository.java
```diff
@@ -0,0 +1,67 @@
...
+
+import models.internal.scheduling.Job;
+
+import java.time.Instant;
+import java.util.NoSuchElementException;
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * A storage medium for {@link Job}s. Essentially a Map that mints unique keys for new Job values.
```
vjkoskela: This `Essentially a Map that mints unique keys for new Job values.` seems more implementation than interface.

app/models/internal/scheduling/Job.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    UUID getId();
+
+    /**
+     * @return The schedule on which the Job should be repeated.
+     */
+    Schedule getSchedule();
+
+    /**
+     * @return A {@link CompletionStage} that completes exceptionally iff the job throws an exception.
```
vjkoskela: In complete Javadoc.

app/models/internal/scheduling/Job.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * The unique identifier of the job.
+     *
+     * @return The unique identifier of the job.
+     */
+    UUID getId();
+
+    /**
+     * @return The schedule on which the Job should be repeated.
```
vjkoskela: In complete Javadoc.

app/models/internal/scheduling/Job.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.concurrent.CompletionStage;
+
+/**
+ * A (possibly recurring) job describing a task to perform and how often to repeat.
+ *
+ * @param <T> The type of result the Job computes.
+ *
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public interface Job<T> {
```
cwbriones: Can you add the `State` inner enum to this? Representing whether the Job ended in `SUCCESS` or `FAILURE`. We'll include the running case in later. This is referred to as "result" in the spec but state captures the fact that it could still be running.

app/com/arpnetworking/metrics/portal/scheduling/JobRepository.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @param <T> The type of result produced by the {@link Job}s.
+ *
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public interface JobRepository<T> {
```
cwbriones: This needs a `jobCompleted(Job<T> job, Job.State state, T result)`. I had considered renaming this since completed doesn't really make sense when the status update is `RUNNING` but we can cross that bridge when we get there with a rename. 

I would also consider if `result` should be `@Nullable` since it seems nonsensical to expect a result when the Job failed.

app/com/arpnetworking/metrics/portal/scheduling/JobRepository.java
cwbriones: I think this is the right way to go, I came to a similar conclusion yesterday when implementing this functionality in the data model. I think the time zone should be confined to the view and that layer should be where we deal with `ZonedDateTime`.

PR #142, opened 2019-01-08
Added KairosDBClient methods for querying and modifying rollups. 

ddimensia:
This includes the methods needed to list, create, update and delete rollup tasks.

I also added tests for existing methods that had none.


app/com/arpnetworking/kairos/client/KairosDBRequestException.java
```diff
@@ -0,0 +1,91 @@
...
+package com.arpnetworking.kairos.client;
+
+import akka.http.javadsl.model.Uri;
+
+/**
+ * An exception that represents a non-2xx KairosDB request.
+ *
+ * @author Gilligan Markham (gmarkham at dropbox dot com)
+ */
+public class KairosDBRequestException extends RuntimeException {
```
vjkoskela: nit: should be `Db` not `DB`; I believe the existing test class (at least) is also wrong.

app/com/arpnetworking/kairos/client/models/RollupTask.java
```diff
@@ -0,0 +1,135 @@
...
+            super(RollupTask::new);
+        }
+
+        /**
+         * Sets the id. Optional. Cannot be empty.
+         *
+         * @param value the id
+         * @return this {@link Builder}
+         */
+        public Builder setId(final String value) {
```
vjkoskela: The argument should be annotated as `@Nullable`.

app/com/arpnetworking/kairos/client/models/Sampling.java
```diff
@@ -69,6 +70,28 @@ public Builder setPeriod(final Period value) {
             return this;
         }
 
+        /**
+         * Sets the count of units that the Sampling covers.
+         *
+         * @param value the count of units
+         * @return this {@link Builder}
+         */
+        public Builder setValue(final int value) {
```
vjkoskela: This should be an `Integer` and marked as `@Nullable` based on the constraints below. The conversion from `Integer` to `int` should be in the constructor of the class being built.

PR #143, opened 2019-01-09
Schedule fixups

speezepearson:
- bugfix in how PeriodicSchedule interacts with DST switchovers (and add tests for such cases)
- prefer `Instant` to `ZonedDateTime` ~everywhere; PeriodicSchedule still uses zones, but the `Schedule` interface now uses plain `Instant`s, because Schedules should have the option to not care about time zones.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicSchedule.java
+++ b/app/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicSchedule.java
@@ -88,7 +88,7 @@ public final class PeriodicSchedule extends BaseSchedule {
         @NotNull
         private ChronoUnit _period;
         @NotNull
-        protected ZoneId _zone;
+        private ZoneId _zone;
         @NotNull
         @ValidateWithMethod(methodName = "validateOffset", parameterType = Duration.class)
         private Duration _offset = Duration.ZERO;
--- a/app/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicSchedule.java
+++ b/app/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicSchedule.java
@@ -15,6 +15,8 @@
  */
 package com.arpnetworking.metrics.portal.scheduling.impl;
 
+import com.arpnetworking.steno.Logger;
+import com.arpnetworking.steno.LoggerFactory;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import net.sf.oval.constraint.NotNull;
 import net.sf.oval.constraint.ValidateWithMethod;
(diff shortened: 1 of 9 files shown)


app/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicSchedule.java
```diff
@@ -32,26 +36,68 @@
...
+
+        // I am not intimately familiar with timekeeping standards, but:
+        //   it's conceivable that, for time-wonkiness reasons, the truncation might have moves us backwards
+        //   farther than adding `_period` moves us forward, meaning that `result` might still be too early.
+        // Just in case, and not very principled-ly, let's try larger steps.
+        if (!result.toInstant().isAfter(boundary)) {
+            result = truncated.plus(_period.getDuration().multipliedBy(2)).truncatedTo(_period);
+        }
+        if (!result.toInstant().isAfter(boundary)) {
+            result = truncated.plus(_period.getDuration().multipliedBy(3)).truncatedTo(_period);
```
vjkoskela: So it looks like we add 1 period above, then may add 2 more, and then 3 more for a total of 6 periods. In what cases might these be necessary? Why is it adding 1, 3 or 6 periods (e.g. as opposed to 1, 2, or 3)?

The truncate zeros all fields less than _period in magnitude. Adding durations of period should not impact truncation at all.

Failing to understand what we're concerned about. Is this for avoiding the DST change?

app/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicSchedule.java
```diff
@@ -32,26 +36,68 @@
...
+    /**
+     * Returns the first instant after the given boundary that is [period]-aligned in our time zone.
+     *
+     * @param boundary The exclusive lower bound.
+     * @return A boundary that is guaranteed to be [period]-aligned and after the given boundary.
+     */
+    private Instant nextAlignedBoundaryAfter(final Instant boundary) {
+        final ZonedDateTime zonedLastRun = ZonedDateTime.ofInstant(boundary, _zone);
+        final ZonedDateTime truncated = zonedLastRun.truncatedTo(_period);
+        ZonedDateTime result = truncated.plus(_period.getDuration()).truncatedTo(_period);
```
vjkoskela: This implementation seems rather complicated, and reading the comments I am unclear as to the cause. 

First, the caller, `unboundedNextRun`, takes the `runAtAndAfter` and substracts one nanosecond. Why is this done?

= First Run =
* Take runAtAndAfter and truncate to period
* Add period until the date is equal to or after runAtAndAfter (e.g. not before); this will be zero or one addition

= Next Run =
* Truncate the last run date to the period to remove the offset which must be non-negative and less than the period. Also accounts for any period changes since the last run.
* Take the truncated last run date and add the interval. 


After, to either of these results add the offset.

app/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicSchedule.java
speezepearson: > (I can't actually produce that failure mode, but my knowledge of TZDB is far from precise enough for me to state with confidence that nothing like this could possibly happen.)

Aha! I don't know why I couldn't produce this before, but I just wrote some code that tries 3 different implementations of "add one day and truncate to days", and they all fail catastrophically on a 25-hour day:
```
>>> import java.time.*
>>> import java.time.temporal.ChronoUnit
>>> 
>>> fun <T> printEquality(x: T, y: T) {
...     println("$x ${if (x==y) "==" else "!="} $y")
... }
>>> 
>>> val add24hAndTruncate = {zt: ZonedDateTime -> (zt + Duration.ofHours(24)).truncatedTo(ChronoUnit.DAYS)}
>>> val add1dAndTruncate = {zt: ZonedDateTime -> (zt + Duration.ofDays(1)).truncatedTo(ChronoUnit.DAYS)}
>>> val addPAndTruncate = {zt: ZonedDateTime, p: ChronoUnit -> (zt + p.duration).truncatedTo(p)}
>>> 
>>> val zt = ZonedDateTime.of(LocalDateTime.of(2018, 11, 4, 0, 0, 0), ZoneId.of("America/Los_Angeles"))
>>> 
>>> printEquality(zt, add24hAndTruncate(zt))
2018-11-04T00:00-07:00[America/Los_Angeles] == 2018-11-04T00:00-07:00[America/Los_Angeles]
>>> 
>>> printEquality(zt, add1dAndTruncate(zt))
2018-11-04T00:00-07:00[America/Los_Angeles] == 2018-11-04T00:00-07:00[America/Los_Angeles]
>>> 
>>> printEquality(zt, addPAndTruncate(zt, ChronoUnit.DAYS))
2018-11-04T00:00-07:00[America/Los_Angeles] == 2018-11-04T00:00-07:00[America/Los_Angeles]
```

test/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicScheduleTest.java
```diff
@@ -55,8 +55,11 @@ public void testNextRunWithAlignedBounds() {
                 Optional.of(Instant.parse("2019-01-02T00:00:00Z")),
                 schedule.nextRun(Optional.of(Instant.parse("2019-01-01T00:00:00Z"))));
         assertEquals(
-                Optional.of(Instant.parse("2019-01-01T00:00:00Z")),
-                schedule.nextRun(Optional.of(Instant.parse("2018-12-20T12:34:56Z"))));
```
speezepearson: This was just bad copy-paste.

PR #146, opened 2019-01-14
add barebones internal models for reports

cwbriones:
Defines the minimal amount necessary to build a report repository. These models are mostly POJOs but additional methods defining behavior are expected (particularly in the case of `ReportSource` and a `Sink` wrapper around `RecipientGroup`).


app/models/internal/impl/ChromeScreenshotReportSource.java
```diff
@@ -0,0 +1,203 @@
...
+         * @param ignoreCertificateErrors Whether this source should ignore certificate errors.
+         * @return This instance of {@code Builder}
+         */
+        public Builder setIgnoreCertificateErrors(final boolean ignoreCertificateErrors) {
+            _ignoreCertificateErrors = ignoreCertificateErrors;
+            return this;
+        }
+
+        @NotNull
+        @NotEmpty
```
vjkoskela: I don't think empty can be applied to a `UUID`.

app/models/internal/impl/ChromeScreenshotReportSource.java
```diff
@@ -0,0 +1,203 @@
...
+        private UUID _id;
+        @NotNull
+        @NotEmpty
+        @AssertURL
```
vjkoskela: Probably better to type this as a `URI`. We may need to assert some properties of the URI; e.g. that it's protocol is either `http` or `https`.

app/models/internal/impl/ChromeScreenshotReportSource.java
```diff
@@ -0,0 +1,203 @@
...
+        private String _url;
+        @NotNull
+        @NotEmpty
+        private String _title;
+        @NotNull
+        @NotEmpty
+        private String _triggeringEventName;
+        private boolean _ignoreCertificateErrors;
```
vjkoskela: Should be `Boolean` as above; probably also `@NotNull`

app/models/internal/impl/DefaultReport.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    public ReportSource getSource() {
+        return _source;
+    }
+
+    @Override
+    public Collection<RecipientGroup> getRecipientGroups() {
+        return Collections.unmodifiableCollection(_groups);
```
vjkoskela: The type should likely be one of Guava's `Immutable*` types. Then you won't need to wrap here.

app/models/internal/impl/DefaultReport.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    public Optional<Instant> getLastRun() {
+        return Optional.empty();
```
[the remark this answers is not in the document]
vjkoskela: Not sure I agree. The two choices are:

1) The internal model object is created with the current last run from the database when it is fetch from the repository using `getJobI(UUID)`.

2) The internal model object does not have this field and it must be fetched separately from the repository. 

There are pros/cons to each approach. Regardless, the caller needs to understand the staleness/freshness of the internal model data and act accordingly. @speezepearson is probably in the best position to say which works better since he is writing the primary consumer code for this data.

@BrandonArp thoughts?

app/models/internal/impl/DefaultReport.java
```diff
@@ -0,0 +1,167 @@
...
+    @Override
+    public CompletionStage<Report.Result> execute(final ActorRef scheduler, final Instant scheduled) {
+        return CompletableFuture.completedFuture(null);
+    }
+
+    private final UUID _id;
+    private final String _name;
+    private final Schedule _schedule;
+    private final ReportSource _source;
+    private final Collection<RecipientGroup> _groups;
```
vjkoskela: As mentioned above this should be `Immutable*` type from Guava (e.g. set or list).

app/models/internal/impl/DefaultReport.java
```diff
@@ -0,0 +1,167 @@
...
+    /**
+     * Builder implementation that constructs {@code DefaultReport}.
+     */
+    public static final class Builder extends OvalBuilder<DefaultReport> {
+        /**
+         * Public Constructor.
+         */
+        public Builder() {
+            super(DefaultReport::new);
+            _schedule = NeverSchedule.getInstance();
```
vjkoskela: nit: move defaults to member declarations.

app/models/internal/impl/DefaultReport.java
```diff
@@ -0,0 +1,167 @@
...
+            _groups = Collections.emptySet();
+        }
+
+        /**
+         * Set the report id. Required. Cannot be null.
+         * @param id The report id.
```
vjkoskela: nit: add blank between description and params.

app/models/internal/impl/DefaultReport.java
vjkoskela: ^ in fact your comments state that these are required, so they should not even have defaults.

app/models/internal/impl/DefaultReport.java
```diff
@@ -0,0 +1,167 @@
...
+            _source = source;
+            return this;
+        }
+
+        /**
+         * Set the report recipients. Required. Cannot be null.
+         * @param groups The report recipient groups.
+         * @return This instance of {@code Builder}.
+         */
+        public Builder setRecipientGroups(final Collection<RecipientGroup> groups) {
```
vjkoskela: The problem with accepting Collection like this is you don't know whether you need to create a copy; otherwise your caller can mutate your internal model object thus breaking your immutability.

Instead, we generally prefer to accept Guava's `Immutable*` to avoid the copy. If the caller doesn't have one then they are forced to copy, so the cost is never any worse, but the contract is clearer, immutability is preserved and when the caller does have an `Immutable*` it's zero-copy.

app/models/internal/impl/EmailRecipientGroup.java
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * A {@code RecipientGroup} that consists of a set of email addresses.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class EmailRecipientGroup implements RecipientGroup {
+    private final UUID _id;
+    private final String _name;
+    private final Set<String> _emails;
```
vjkoskela: Guava `ImmutableSet`

app/models/internal/impl/EmailRecipientGroup.java
```diff
@@ -0,0 +1,133 @@
...
+    private final List<ReportFormat> _formats;
```
vjkoskela: Guava `ImmutableList`

app/models/internal/impl/EmailRecipientGroup.java
```diff
@@ -0,0 +1,133 @@
...
+        /**
+         * Public constructor.
+         */
+        public Builder() {
+            super(EmailRecipientGroup::new);
+        }
+
+        /**
+         * The group id. Required. Cannot be null or empty.
+         * @param id The id of the group.
```
vjkoskela: nit: blank line between.

app/models/internal/impl/EmailRecipientGroup.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        @NotNull
+        private UUID _id;
+        @NotNull
+        @NotBlank
+        private String _name;
+        @NotNull
+        @MinSize(value = 1)
+        private Set<String> _emails;
```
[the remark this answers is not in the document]
vjkoskela: Here's the implementation:

https://github.com/sebthom/oval/blob/master/src/main/java/net/sf/oval/constraint/EmailCheck.java

Seem to work:

https://www.regexpal.com/?fam=107035

Of course there may be a bug in the checker; can you provide a test case?

Granted, it does not appear to be RFC compliant:

https://gist.github.com/gregseth/5582254

app/models/internal/impl/HTMLReportFormat.java
```diff
@@ -0,0 +1,56 @@
...
+
+import com.google.common.base.MoreObjects;
+import models.internal.reports.ReportFormat;
+
+/**
+ * An HTML report format.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class HTMLReportFormat implements ReportFormat {
```
vjkoskela: nit: `HTML` to `Html`; if it's common enough to use an abbreviation treat the abbreviation as a word and camel case it.

app/models/internal/impl/HTMLReportFormat.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final HTMLReportFormat INSTANCE = new HTMLReportFormat();
```
vjkoskela: So this is being done because the HTML report format has no parameters; what happens if we add one? I would recommend that we setup the correct skeleton but have the `Builder` return a static instance.

app/models/internal/impl/PDFReportFormat.java
```diff
@@ -0,0 +1,129 @@
...
+
+import java.util.Objects;
+import javax.annotation.Nonnegative;
+
+/**
+ * A PDF report format.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class PDFReportFormat implements ReportFormat {
```
vjkoskela: nit: `PDF` to `Pdf`

app/models/internal/impl/PDFReportFormat.java
```diff
@@ -0,0 +1,129 @@
...
+        /**
+         * Public constructor.
+         */
+        public Builder() {
+            super(PDFReportFormat::new);
+        }
+
+        /**
+         * Set the width. Required. Must be a positive value.
+         * @param widthInches The width, in inches.
```
vjkoskela: nit: add blank line

app/models/internal/reports/RecipientGroup.java
```diff
@@ -0,0 +1,56 @@
...
+     * @return the id for this group.
+     */
+    UUID getId();
+
+    /**
+     * Get the members of this group.
+     *
+     * @return A collection containing the group members.
+     */
+    Collection<String> getMembers();
```
vjkoskela: `Immutable*`

app/models/internal/reports/RecipientGroup.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Get the formats requested by this group.
+     *
+     * @return A collection of formats requested by this group.
+     */
+    Collection<ReportFormat> getFormats();
```
vjkoskela: `Immutable*`

app/models/internal/reports/Report.java
```diff
@@ -0,0 +1,54 @@
...
+     * @return The the source for this report.
+     */
+    ReportSource getSource();
+
+    /**
+     * Get the recipient groups that will receive this report.
+     *
+     * @return The recipient groups that will receive this report.
+     */
+    Collection<RecipientGroup> getRecipientGroups();
```
vjkoskela: `Immutable*`

app/models/internal/reports/Report.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Marker interface for a result created from a report.
+     */
+    interface Result {}
```
vjkoskela: I don't think this should be a marker, it should be an internal model interface for an internal model implementation. Both of these can be empty _for now_; however, it's not intended to be a marker in the medium-term.

app/models/internal/reports/ReportFormat.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * {@code Visitor} abstracts over operations which could potentially handle various
+     * implementations of ReportFormat.
+     *
+     * @param <T> the return type of the visitor.
+     */
+    interface Visitor<T> {
+        /**
+         * Visit an {@code PDFReportFormat}.
+         * @param pdfReportFormat The format to visit.
```
vjkoskela: nit: blank line missing

app/models/internal/impl/ChromeScreenshotReportSource.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.UUID;
+
+/**
+ * Internal model for a report source that pulls screenshots from a Chrome web browser.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class ChromeScreenshotReportSource implements ReportSource {
+    private final UUID _id;
+    private final String _url;
```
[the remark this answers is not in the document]
cwbriones: 👍 I was concerned that `AssertURL` would only work on String but it looks like it converts it and parses it back out again for the validation: https://github.com/sebthom/oval/blob/master/src/main/java/net/sf/oval/constraint/AssertURLCheck.java#L134

app/models/internal/impl/DefaultReport.java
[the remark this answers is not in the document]
cwbriones: After talking to @speezepearson I realized this functionality is [already on JobRepository](https://github.com/ArpNetworking/metrics-portal/blob/cd8be32ec3be1dede49047bca43fbe1f5144ad2c/app/com/arpnetworking/metrics/portal/scheduling/JobRepository.java#L70) which coincides with option (2). Unless you have any objections I'll remove this and we can commit to the latter.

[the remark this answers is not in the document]
vjkoskela: Yes, that's fine. I spoke with Spencer about this yesterday afternoon; just make sure you're on the same page.

app/models/internal/impl/EmailRecipientGroup.java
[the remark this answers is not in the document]
vjkoskela: Is this the relevant issue on commons?

https://github.com/ArpNetworking/commons/issues/23

I agree, this is not the time to fork more work. I assume you can work around it for now? 

Please add a TODO to to the Metrics Portal code here with a link to issue 23 above.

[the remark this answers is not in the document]
cwbriones: I created a new issue here: https://github.com/ArpNetworking/commons/issues/79, I'll link to it in the TODO. This is fine, I can just use a method validator for the time being.

app/models/internal/impl/ChromeScreenshotReportSource.java
```diff
@@ -0,0 +1,208 @@
...
+import java.net.URI;
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * Internal model for a report source that pulls screenshots from a Chrome web browser.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class ChromeScreenshotReportSource implements ReportSource {
```
vjkoskela: Probably should mark this `@Loggable`

app/models/internal/impl/DefaultReport.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+/**
+ * Default implementation of {@link Report}.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class DefaultReport implements Report {
```
vjkoskela: Probably should mark this `@Loggable`

app/models/internal/impl/EmailRecipientGroup.java
```diff
@@ -0,0 +1,186 @@
...
+
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * A {@code RecipientGroup} that consists of a set of email addresses.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class EmailRecipientGroup implements RecipientGroup {
```
vjkoskela: Probably should mark this `@Loggable`

app/models/internal/impl/HtmlReportFormat.java
```diff
@@ -0,0 +1,72 @@
...
+import com.arpnetworking.commons.builder.OvalBuilder;
+import com.google.common.base.MoreObjects;
+import models.internal.reports.ReportFormat;
+
+/**
+ * An HTML report format.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class HtmlReportFormat implements ReportFormat {
```
vjkoskela: Probably should mark this `@Loggable`

app/models/internal/impl/PdfReportFormat.java
```diff
@@ -0,0 +1,131 @@
...
+
+import java.util.Objects;
+import javax.annotation.Nonnegative;
+
+/**
+ * A PDF report format.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class PdfReportFormat implements ReportFormat {
```
vjkoskela: Probably should mark this `@Loggable`

PR #150, opened 2019-01-23
Job executor overhaul

speezepearson:
**Sorry this is a Big Thing.** Unfortunately, this is _after_ slicing it thinner. The remaining changes are sufficiently tightly coupled that I'm pretty sure that the most efficient thing to do is to make you review this big thing.

New functionality:
- Make executor hit the repository less (encapsulating that logic in the new `CachedJob` class)
- Add `Reload` message to executor, for anti-entropy
  - (related)Add `getETag()` method to `Job` interface
- [testing] Add `DummyJob` class for tests

Refactors:
- Undo a previous change: executor now has periodic ticks, rather than scheduling its next tick after completing each execution. This introduces some threading concerns, which is sad; but it's much simpler than ensuring that only one series of ticks is running at a time, while accounting for restarts.
- Pass `JobRef` into executor via messages rather than via props -- this will be necessary because Akka's ClusterSharding requires all its entities to have the same props.

### DIFF
--- a/test/com/arpnetworking/metrics/portal/scheduling/JobExecutorActorTest.java
+++ b/test/com/arpnetworking/metrics/portal/scheduling/JobExecutorActorTest.java
@@ -174,13 +174,14 @@ public final class JobExecutorActorTest {
 
     @Test
     public void testOnlyExecutesOneAtATime() {
-        final Instant startAt = t0.minus(ChronoUnit.HOURS.getDuration());
+        final ChronoUnit period = ChronoUnit.MINUTES;
+        final Instant startAt = t0.minus(period.getDuration());
         final CompletableFuture<Void> blocker = new CompletableFuture<>();
         final DummyJob<Integer> job = addJobToRepo(new DummyJob.Builder<Integer>()
                 .setSchedule(new PeriodicSchedule.Builder()
                         .setRunAtAndAfter(startAt)
                         .setZone(ZoneId.of("UTC"))
-                        .setPeriod(ChronoUnit.MINUTES)
+                        .setPeriod(period)
                         .build())
                 .setResult(123)
                 .setBlocker(blocker)
@@ -202,10 +203,9 @@ public final class JobExecutorActorTest {
 
         blocker.complete(null);
 
-        // NOW we should be able to run again
-        executor.tell(JobExecutorActor.Tick.INSTANCE, null);
+        // NOW we should be able to run again; the necessary tick should have been triggered by job completion
         Mockito.verify(repo, Mockito.timeout(1000)).jobStarted(job.getId(), organization, job.getSchedule().nextRun(Optional.of(startAt)).get());
-        // ...but still, only two executions should ever have started (one for t0, one for the moment after t0)
+        // ...but still, only two executions should ever have started (one for t0, one for t0+period i.e. now)
         Mockito.verify(repo, Mockito.timeout(1000).times(2)).jobStarted(Mockito.eq(job.getId()), Mockito.eq(organization), Mockito.any());
     }
 


app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -45,39 +51,28 @@
...
      */
-    protected static <T> Props props(final Injector injector, final JobRef<T> jobRef, final Clock clock) {
-        return Props.create(JobExecutorActor.class, () -> new JobExecutorActor<>(injector, jobRef, clock));
+    public static Props props(final Injector injector,
+                              final Clock clock) {
+        return Props.create(JobExecutorActor.class, () -> new JobExecutorActor<>(injector, clock));
     }
 
-    private JobExecutorActor(final Injector injector, final JobRef<T> jobRef, final Clock clock) {
+    private JobExecutorActor(final Injector injector,
```
vjkoskela: Same.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -87,7 +82,13 @@ private JobExecutorActor(final Injector injector, final JobRef<T> jobRef, final
     @Override
     public void preStart() throws Exception {
         super.preStart();
-        timers().startSingleTimer("TICK", Tick.INSTANCE, TICK_INTERVAL);
+        timers().startPeriodicTimer("TICK", Tick.INSTANCE, TICK_INTERVAL);
```
vjkoskela: nit: pull "TICK" into private static final?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +97,65 @@ public void preStart() throws Exception {
...
-        } else {
-            return TICK_INTERVAL;
+            return Optional.of(delta);
         }
+        return Optional.empty();
     }
 
-    private void scheduleNextTick(final Instant wakeUpBy) {
-        timers().startSingleTimer("TICK", Tick.INSTANCE, timeUntilNextTick(wakeUpBy));
+    private void scheduleExtraTickIfNecessary(final Instant wakeUpBy) {
```
vjkoskela: Is it an "extra" tick or an irregular/shorter?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +97,65 @@ public void preStart() throws Exception {
...
+        timeUntilExtraTick(wakeUpBy).ifPresent(timeRemaining -> {
+            timers().startSingleTimer("TICK", Tick.INSTANCE, timeRemaining);
```
vjkoskela: Does this replace the `"TICK"` timer?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +97,65 @@ public void preStart() throws Exception {
...
-                    } else {
-                        repo.jobFailed(job.getId(), org, scheduled, error);
-                    }
-                    return null;
-                })
-                .thenApply(whatever -> {
-                    nextScheduled.ifPresent(this::scheduleNextTick);
-                    return null;
-                });
+    private void stopForever() {
```
vjkoskela: This isn't entirely true right? I mean it's just until the coordinator resurrects it?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +97,65 @@ public void preStart() throws Exception {
...
+
+        final JobRef<T> oldRef = _cachedJob.get().getRef();
+        if (!oldRef.equals(ref)) {
+            LOGGER.error()
+                    .setMessage("JobRef in received message does not match cached JobRef")
+                    .addData("cached", oldRef)
+                    .addData("new", ref)
+                    .log();
+            stopForever();
+            throw new IllegalStateException("got JobRef id=" + ref.getJobId() + " but already initialized with id=" + oldRef.getJobId());
```
vjkoskela: nit: Prefer `String.format("...")` for readability.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -140,49 +164,159 @@ public Receive createReceive() {
...
                     _periodicMetrics.recordCounter("job-executor-actor-ticks", 1);
 
-                    final JobRepository<T> repo = _jobRef.getRepository(_injector);
-                    final Optional<Job<T>> job = _jobRef.get(_injector);
-                    if (!job.isPresent()) {
-                        LOGGER.warn()
-                                .setMessage("no such job")
-                                .addData("jobRef", _jobRef)
+                    if (!_cachedJob.isPresent()) {
+                        LOGGER.info()
```
vjkoskela: This should never happen right? Or conversely, should we wait to schedule the timer until after initialization? (the isPresent check is probably still required but at that point becomes and error)

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -140,49 +164,159 @@ public Receive createReceive() {
...
+        private Reload(final Builder<T> builder) {
+            _jobRef = builder._jobRef;
+            _eTag = builder._eTag;
+        }
+
+        public JobRef<T> getJobRef() {
+            return _jobRef;
+        }
+
+        @Nullable
```
vjkoskela: Should be represented as `Optional` on the message itself.

app/models/internal/impl/DefaultReport.java
```diff
@@ -53,6 +55,11 @@ public UUID getId() {
         return _id;
     }
 
+    @Override
+    public String getETag() {
+        return DigestUtils.sha1Hex(toString());
```
vjkoskela: Really shouldn't use `toString()` to compute etag; this should probably come from the repository/data model. Feel free to leave it for now with a `TODO`. @cwbriones let me know if you think otherwise.

test/com/arpnetworking/metrics/portal/scheduling/impl/MapJobRepository.java
```diff
@@ -29,6 +29,7 @@
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Stream;
```
vjkoskela: Unused?

test/com/arpnetworking/metrics/portal/scheduling/mocks/DummyJob.java
```diff
@@ -0,0 +1,131 @@
+package com.arpnetworking.metrics.portal.scheduling.mocks;
```
vjkoskela: Missing copyright header.

test/com/arpnetworking/metrics/portal/scheduling/mocks/DummyJob.java
```diff
@@ -0,0 +1,131 @@
...
+import net.sf.oval.constraint.NotNull;
+import net.sf.oval.constraint.ValidateWithMethod;
+
+import java.time.Instant;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+public class DummyJob<T> implements Job<T> {
```
vjkoskela: Missing class javadoc

app/com/arpnetworking/metrics/portal/scheduling/CachedJob.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Override
+    public Schedule getSchedule() {
+        return _cached.getSchedule();
+    }
+
+    @Override
+    public CompletionStage<T> execute(final ActorRef scheduler, final Instant scheduled) {
+        final JobRepository<T> repo = _ref.getRepository(_injector);
+        repo.jobStarted(_cached.getId(), _ref.getOrganization(), scheduled);
```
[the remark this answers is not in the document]
speezepearson: I opted to refactor ~all the execution logic into a method of `JobExecutorActor`. `CachedJob.execute` is now a pure delegation to the underlying job, and the reloading / repo-notification is [done by the actor](https://github.com/ArpNetworking/metrics-portal/pull/150/files#diff-0cd7c6e23360f8f50d83def6cc256213R163). Acceptable?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +100,108 @@ public void preStart() throws Exception {
...
+
+    private void scheduleExtraTickIfNecessary(final Instant wakeUpBy) {
+        timeUntilExtraTick(wakeUpBy).ifPresent(timeRemaining -> {
+            timers().startSingleTimer(EXTRA_TICK_TIMER_NAME, Tick.INSTANCE, timeRemaining);
+        });
+    }
+
+    private void killSelf() {
+        // TODO(spencerpearson): Per https://doc.akka.io/docs/akka/2.5.4/java/cluster-sharding.html#remembering-entities ,
+        //   might we need to send a Passivate message to our parent?
```
vjkoskela: The simple policy is to always send the passivate and be removed from the cluster (once you add that functionality). 

However, this policy can be optimized (for background) based on the case you that you are shutting down. In the case the job does not exist you want to send the passivate and not be resurrected. In the case that there is a transient error you probably don't mind getting resurrected elsewhere. Although we don't have retries the job will continue to execute a scheduled run until it is able to record failure for that run (so it the face of meta-failures we try forever). Removing from the cluster in this case (and any other where the job still exists) doesn't matter because the JobCoordinator will resurrect the job anyway.

Anyway, doesn't apply until the next (?) review when you add cluster registration of these actors.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +100,108 @@ public void preStart() throws Exception {
...
+        }
+
+        final JobRef<T> oldRef = _cachedJob.get().getRef();
+        if (!oldRef.equals(ref)) {
+            LOGGER.error()
+                    .setMessage("JobRef in received message does not match cached JobRef")
+                    .addData("cached", oldRef)
+                    .addData("new", ref)
+                    .log();
+            killSelf();
```
vjkoskela: Instead of logging here and killing can we throw an `IllegalStateException`? We can then log the exception from the `preRestart`. This way I feel the control flow is simpler and less explicit. We also have the opportunity to log an exception instead of just a message which should have other integration paths in the future. The only downside I see is losing some the structured nature in the logs; however, since these should _never_ happen, I think it's a reasonable trade-off.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +100,108 @@ public void preStart() throws Exception {
...
-            final JobRepository<T> repo,
-            final Organization org,
-            final Job<T> job,
-            final Instant scheduled
-    ) {
-        final Optional<Instant> nextScheduled = job.getSchedule().nextRun(Optional.of(scheduled));
-        repo.jobStarted(job.getId(), org, scheduled);
+    private void attemptExecuteAndUpdateRepository(final Instant scheduled) {
+        if (!_cachedJob.isPresent()) {
+            LOGGER.error()
```
vjkoskela: See note above about `IllegalStateException`; this feels like a similar situation.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +100,108 @@ public void preStart() throws Exception {
...
+
+        if (_currentlyExecuting.getAndSet(true)) {
+            return;
+        }
+
+        try {
+            repo.jobStarted(ref.getJobId(), ref.getOrganization(), scheduled);
+        } catch (final NoSuchElementException error) {
+            _currentlyExecuting.set(false);
+            killSelf();
```
vjkoskela: Should we log something here?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +100,108 @@ public void preStart() throws Exception {
...
-                        repo.jobFailed(job.getId(), org, scheduled, error);
+                    try {
+                        if (error == null) {
+                            repo.jobSucceeded(ref.getJobId(), ref.getOrganization(), scheduled, result);
+                        } else {
+                            repo.jobFailed(ref.getJobId(), ref.getOrganization(), scheduled, error);
+                        }
+                        cachedJob.reload(_injector);
+                    } catch (final NoSuchElementException error2) {
+                        killSelf();
```
vjkoskela: Should we log something here?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +100,108 @@ public void preStart() throws Exception {
...
                     }
-                    return null;
+                    return result;
                 })
-                .thenApply(whatever -> {
-                    nextScheduled.ifPresent(this::scheduleNextTick);
-                    return null;
-                });
+                .whenComplete((result, error) -> _currentlyExecuting.set(false));
```
vjkoskela: Probably more akka-esq to send a `JobCompletion` message (or success/failure separate messages) as discussed.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -96,42 +100,108 @@ public void preStart() throws Exception {
...
+
         job.execute(getSelf(), scheduled)
                 .handle((result, error) -> {
-                    if (error == null) {
-                        repo.jobSucceeded(job.getId(), org, scheduled, result);
-                    } else {
-                        repo.jobFailed(job.getId(), org, scheduled, error);
+                    try {
+                        if (error == null) {
+                            repo.jobSucceeded(ref.getJobId(), ref.getOrganization(), scheduled, result);
```
vjkoskela: IMO the succeeded/failed branches should log some debug/info (or at least trace) level data about the change in state

vjkoskela: What about having metrics here?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -140,49 +210,159 @@ public Receive createReceive() {
...
                     _periodicMetrics.recordCounter("job-executor-actor-ticks", 1);
 
-                    final JobRepository<T> repo = _jobRef.getRepository(_injector);
-                    final Optional<Job<T>> job = _jobRef.get(_injector);
-                    if (!job.isPresent()) {
-                        LOGGER.warn()
-                                .setMessage("no such job")
-                                .addData("jobRef", _jobRef)
+                    if (!_cachedJob.isPresent()) {
+                        LOGGER.error()
```
vjkoskela: See note above about `IllegalStateException`

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -87,7 +83,15 @@ private JobExecutorActor(final Injector injector, final JobRef<T> jobRef, final
     @Override
     public void preStart() throws Exception {
         super.preStart();
-        timers().startSingleTimer("TICK", Tick.INSTANCE, TICK_INTERVAL);
+        timers().cancel(REGULAR_TICK_TIMER_NAME);
```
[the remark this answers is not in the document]
BrandonArp: IIRC, AbstractActorWithTimers takes care of canceling all of the timers for you.  https://doc.akka.io/japi/akka/current/akka/actor/AbstractActorWithTimers.html

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -18,171 +18,486 @@
...
+            attemptExecuteAndUpdateRepository(nextRun.get());
         }
     }
 
-    private void scheduleNextTick(final Instant wakeUpBy) {
-        timers().startSingleTimer("TICK", Tick.INSTANCE, timeUntilNextTick(wakeUpBy));
+    private void reload(final Reload<T> message) {
+        final Optional<String> eTag = message.getETag();
+        final CachedJob<T> cachedJob;
+        _periodicMetrics.recordCounter("job_executor_actor_reloads", 1);
```
vjkoskela: This seems redundant because the cached actor is already instrumented for the same. See comment above on it. One avenue could be to put the instrumentation only here. Not sure if that helps.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -18,171 +18,486 @@
...
+                    .log();
+            return;
+        }
+        final CachedJob<T> cachedJob = _cachedJob.get();
+        final JobRef<T> ref = cachedJob.getRef();
+
+        @SuppressWarnings("unchecked")
+        final JobCompleted<T> typedMessage = (JobCompleted<T>) message;
+        final JobRepository<T> repo = ref.getRepository(_injector);
+        try {
```
vjkoskela: This structure and the comments above almost seem to argue for separate message types.

app/com/arpnetworking/metrics/portal/scheduling/CachedJob.java
```diff
@@ -0,0 +1,130 @@
...
+        if (upToDate) {
+            return;
+        }
+        LOGGER.debug()
+                .setMessage("job is stale; reloading")
+                .addData("jobRef", _ref)
+                .addData("oldETag", _cached.getETag())
+                .addData("newETag", upToDateETag)
+                .log();
+        reload(injector);
```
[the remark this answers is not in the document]
speezepearson: Ah, I was intending the metric in `reloadIfOutdated` to be a measure of how often a conditional reload results in needing to reload at all. I just retitled this metric `cached_job_conditional_reload_necessary`, which is... hopefully clearer? Now we'll have two metrics (`cached_job_conditional_reload_necessary` and `cached_job_reload_success`) which, together, will tell us:
- the [number / success rate] of reloads; and
- the [number / out-of-dateness rate] of conditional reloads.

Are there other metrics worth adding?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -18,171 +18,486 @@
...
 import com.arpnetworking.steno.Logger;
 import com.arpnetworking.steno.LoggerFactory;
+import com.google.common.base.MoreObjects;
 import com.google.inject.Injector;
-import models.internal.Organization;
 import models.internal.scheduling.Job;
+import net.sf.oval.constraint.NotNull;
 import scala.concurrent.duration.Duration;
 import scala.concurrent.duration.FiniteDuration;
+import scala.util.Either;
```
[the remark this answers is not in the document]
speezepearson: This _does_ feel gross. But I love compile-time errors infinitely more than writing `if ((foo==null) == (bar==null)) throw new IllegalArgumentException(...)`. The type system is a _great_ tool to enforce constraints like this, and if there's a way to get that without importing `scala.util` and without typing the words `public final class Either<A, B> {...`, I'm 100% in favor.

The Java community seems less trigger-happy on dependencies than the JavaScript community, so I suppose you'd also feel kinda >:[ about [a dependency](https://mvnrepository.com/artifact/io.github.franzgranlund/either/1.3) (or [a larger one](https://mvnrepository.com/artifact/org.functionaljava/functionaljava))?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -18,171 +18,486 @@
...
-     * @param <T> The type of result produced by the {@link JobRef}'s job.
-     * @param injector The Guice injector to use to load the {@link JobRepository} referenced by the {@link JobRef}.
-     * @param jobRef The job to intermittently execute.
-     * @return A new props to create this actor.
-     */
-    public static <T> Props props(final Injector injector, final JobRef<T> jobRef) {
-        return props(injector, jobRef, Clock.systemUTC());
-    }
+    private final PeriodicMetrics _periodicMetrics;
+    private final AtomicBoolean _currentlyExecuting = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
speezepearson: It doesn't! It is now only accessed by the actor (i.e. in a message-handling call stack). I left it Atomic out of... out of an abundance of caution, I guess. Could unwrap if you prefer.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
speezepearson: Maybe I went a bit metrics-happy. This certainly is, information-theoretically, redundant, because any aggregate of this metric can be computed from aggregates of the reload / conditional-reload metrics. But I don't like writing MetaQueries in Grafana. Could still remove this if you feel strongly.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -18,171 +18,486 @@
...
+                    .setMessage("initializing")
+                    .addData("ref", ref)
+                    .log();
+            _cachedJob = Optional.of(new CachedJob<>(_injector, ref, _periodicMetrics));
+        }
+
+        final JobRef<T> oldRef = _cachedJob.get().getRef();
+        if (!oldRef.equals(ref)) {
+            killSelf();
+            throw new IllegalStateException(String.format("got JobRef %s, but already initialized with %s", ref, oldRef));
```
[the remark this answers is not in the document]
BrandonArp: This one is actually pretty easy to answer.  killSelf() sends a poison pill message to itself.  This means that the message will be enqueued to the mailbox (not in the actor), then the actor will throw an exception, be restarted, and the newly spawned replacement actor will process the poison pill message.

app/com/arpnetworking/metrics/portal/scheduling/CachedJob.java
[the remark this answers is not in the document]
vjkoskela: That's fine.

Separating the case of condition reload success from unconditional reload success can be added if ever needed. 

The new names are much better!

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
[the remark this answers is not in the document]
vjkoskela: You are not wrong. 

I couldn't find anything in a brief search that would be better. My main objection is not necessarily the dependency, although you are correct on both counts that the Java community and I are more cautious when it comes to adding dependencies; but rather that the dependency in this case is from Scala! (something my experience tells me rarely ends well)

In this particular case, given how you handle the success and failure entirely independently, it seems that my suggestion below of using separate messages solves the problem quite nicely. No? (at the cost of yet another message class of course)

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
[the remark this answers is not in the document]
vjkoskela: My concern with leaving it is that it typically sends a warning to the reader that there is concurrency going on when this is read or updated when in fact that is not the case. My preference would be to unwrap even if that means making it non-final.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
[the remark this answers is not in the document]
vjkoskela: Ouch! That's not good either ;)

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -18,171 +18,486 @@
...
+                    .addData("scheduled", scheduled)
+                    .log();
+            return;
+        }
+
+        PatternsCS.pipe(
+                job.execute(getSelf(), scheduled)
+                        .handle((result, error) -> new JobCompleted.Builder<T>()
+                                    .setScheduled(scheduled)
+                                    .setOutcome(error, result)
```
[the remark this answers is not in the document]
vjkoskela: ^ based on feedback I recommend separate messages

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
vjkoskela: No, that's fine. Just remember changing metric names, definitions or removing them is effectively changing the "operational interface" of the service. Graphs, alerts and runbooks can depend on these. Metrics should have clear intention, naming and therefore obvious meaning to operators (the hallmarks of a good "interface"). Not suggesting this doesn't meet that bar, just figured I'd point and ask. Leaving it is fine.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -260,33 +258,36 @@ private void jobCompleted(final JobCompleted<?> message) {
         try {
             _periodicMetrics.recordCounter(
                     "job_executor_actor_execution_successes",
-                    message.getOutcome().isRight() ? 1 : 0);
-            if (message.getOutcome().isLeft()) {
+                    (message.getError() == null) ? 1 : 0);
+            if (message.getError() == null) {
+                if (typedMessage.getResult() == null) {
+                    throw new IllegalArgumentException(String.format("JobCompleted message for %s has null error *and* result", ref));
```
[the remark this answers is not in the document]
speezepearson: That'd be acceptable. IntelliJ is sad if I rely on such implicit guarantees about non-nullness, though -- do you have recommendations about how to prove to the typechecker that `result` is non-null here?

PR #152, opened 2019-01-25
Add periodic MetricsDiscovery cluster singleton actor for rollups.

ddimensia:


app/com/arpnetworking/rollups/MetricsDiscovery.java
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Actor for discovering the list of metrics available to be rolled up on a periodic basis.
+ *
+ * This actor maintains an internal set of metric names and acts as a source for downstream
+ * actors that perform the actual rollups.  This is intended to be used as a singleton in the
+ * cluster.
+ *
+ * @author Gilligan Markham (gmarkham at dropbox dot com)
+ */
+public class MetricsDiscovery extends AbstractActorWithTimers {
```
vjkoskela: nit: `final` class?

app/com/arpnetworking/rollups/MetricsDiscovery.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Override
+    public Receive createReceive() {
+        return receiveBuilder()
+                .matchEquals(FETCH_MSG,
```
vjkoskela: Convention is wrap-all; e.g.

not:
```
foo(argA,
        argB,
        argC)
```

instead:
```
foo(
        argA,
        argB,
        argC)
```

app/com/arpnetworking/rollups/MetricsDiscovery.java
```diff
@@ -0,0 +1,117 @@
...
+                        work -> fetchMetricsForRollup())
+                .match(KairosMetricNamesQueryResponse.class,
+                        this::updateMetricsSet)
+                .match(Status.Failure.class,
+                        failure -> LOGGER.warn("Failed to get metrics from Kairo", failure.cause()))
```
vjkoskela: nit: `Kairo` => `Kairos`

app/com/arpnetworking/rollups/MetricsDiscovery.java
```diff
@@ -0,0 +1,117 @@
...
+    private void updateMetricsSet(final KairosMetricNamesQueryResponse response) {
+        response.getResults().stream()
+                .filter(ROLLUP_METRIC_RE.asPredicate().negate())
+                .forEach(_metricsSet::add);
+        _setIterator = null;
+    }
+
+    private Optional<String> getNextMetric() {
+        final String next;
+        if (_setIterator == null) {
```
vjkoskela: Should we protect calls to this if `_metricsSet` is still `null` that they return absent?

vjkoskela: Alternatively, if you set `_setIterator` in `updateMetricsSet` and clear it when there are none left then a `null` value for `_setIterator` means there is no data.

app/com/arpnetworking/rollups/MetricsDiscovery.java
```diff
@@ -0,0 +1,117 @@
...
+            next = null;
+        }
+
+        return Optional.ofNullable(next);
+    }
+
+    private final KairosDbClient _kairosDbClient;
+    private final Set<String> _metricsSet;
+    private Iterator<String> _setIterator = null;
+    private static final String HOURLY_TIMER = "hourly_timer";
```
vjkoskela: You should probably rename this since the interval is config driven and not necessarily hourly.

app/com/arpnetworking/rollups/MetricsDiscovery.java
```diff
@@ -0,0 +1,117 @@
...
+                .match(KairosMetricNamesQueryResponse.class,
+                        this::updateMetricsSet)
+                .match(Status.Failure.class,
+                        failure -> LOGGER.warn("Failed to get metrics from Kairo", failure.cause()))
+                .match(MetricFetch.class, work -> {
+                    final Optional<String> metricName = getNextMetric();
+                    if (metricName.isPresent()) {
+                        getSender().tell(metricName.get(), getSelf());
+                    } else {
+                        getSender().tell(NoMoreMetrics.getInstance(), getSelf());
```
vjkoskela: Should the reply of `NoMoreMetrics` carry the time until the next refresh?

app/com/arpnetworking/rollups/MetricsDiscovery.java
```diff
@@ -0,0 +1,117 @@
...
+        return Optional.ofNullable(next);
+    }
+
+    private final KairosDbClient _kairosDbClient;
+    private final Set<String> _metricsSet;
+    private Iterator<String> _setIterator = null;
+    private static final String HOURLY_TIMER = "hourly_timer";
+    private static final Object FETCH_MSG = new Object();
+    private static final Logger LOGGER = LoggerFactory.getLogger(MetricsDiscovery.class);
+    private static final Pattern ROLLUP_METRIC_RE = Pattern.compile(".*_1[hd]");
```
vjkoskela: Probably clear as a I read on, but is the naming imposed by KairosDb, the "save-as" or something else? What I'm getting at is whether this regex should be configurable.

vjkoskela: Alternatively, should it be `.*_1[hd]$`?

app/global/MainModule.java
```diff
@@ -303,6 +308,33 @@ public ActorRef get() {
...
+                        "rollup-metrics-discovery"
+                );
+            }
+            return null;
+        }
+
+        private final Injector _injector;
+        private final ActorSystem _system;
+
+        private static final String METRICS_DISCOVERY_ROLE = "rollup_metrics_discovery";
```
vjkoskela: Rename `METRICS_DISCOVERY_ROLE` to `ROLLUP_METRICS_DISCOVERY_ROLE`; I can see this becoming a pattern.

app/com/arpnetworking/rollups/MetricsDiscovery.java
ddimensia: done.

app/com/arpnetworking/rollups/MetricsDiscovery.java
ddimensia: done

app/com/arpnetworking/rollups/MetricsDiscovery.java
ddimensia: done

app/com/arpnetworking/rollups/MetricsDiscovery.java
ddimensia: Switched to always have _setIterator defined so that it should never be null.

app/com/arpnetworking/rollups/MetricsDiscovery.java
ddimensia: done

app/global/MainModule.java
ddimensia: done

app/com/arpnetworking/rollups/MetricsDiscovery.java
ddimensia: This isn't enforced in any way by Kairos, this is convention on our end.  We could make this configurable but I'd rather it just be opinionated and not incur the complexity of the extra flexibility.

app/com/arpnetworking/rollups/MetricsDiscovery.java
ddimensia: I'm not sure if I can really calculate that.  I guess I could stamp the next fetch time as the last_start + interval but since it's using a Akka periodic timer I feel that's just guessing.  I would assume the cost of sending a periodic message to see if anything is available is relatively low and therefore not worth the risk of having some time skew problem causing metrics to be missed for a period.

vjkoskela: You could always add a configurable jitter to the advertised target. I think any inadvertent polling would be acceptable. This is similar to the deadline walk that Spencer is dealing with.

The one difference is that your actor is remote to the timestamp. So perhaps sending an interval instead of a target date/time would avoid any host-to-host clock skews. I agree this does nothing to account for Akka timer drift, but the jitter should make-up for most/all of that.

app/com/arpnetworking/rollups/MetricsDiscovery.java
vjkoskela: That's fair. I think it should still be `.*_1[hd]$` - no?

ddimensia: done and expanded tests to verify

app/com/arpnetworking/rollups/MetricsDiscovery.java
ddimensia: I don't fully agree that the complexity vs periodic polling is worth it but I added the number of milliseconds to the next refresh in the message.

PR #153, opened 2019-01-26
Query support within JobRepository

cwbriones:
This PR implements pagination for `DatabaseReportRepository` by adding additional `createQuery` and `query` methods to `JobRepository`.

In addition to this, there's a new null implementation of `ReportRepository` to match the existing ones for other repos, as well as Guice providers for wiring all of this up.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/scheduling/JobQuery.java
+++ b/app/com/arpnetworking/metrics/portal/scheduling/JobQuery.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2019 Dropbox, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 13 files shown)


app/com/arpnetworking/metrics/portal/scheduling/JobQuery.java
```diff
@@ -0,0 +1,74 @@
...
+
+import models.internal.Organization;
+import models.internal.QueryResult;
+import models.internal.scheduling.Job;
+
+import java.util.Optional;
+
+/**
+ * A query against a {@link JobRepository}.
+ * @param <T> The result type of the Jobs returned by this query.
```
vjkoskela: nit: blank line between

app/models/internal/impl/DefaultReport.java
```diff
@@ -85,13 +89,50 @@ public ReportSource getSource() {
         return CompletableFuture.completedFuture(null);
     }
 
+    @Override
+    public String toString() {
+        final List<UUID> recipientIds =
```
vjkoskela: Is there a reason to custom serialize this versus relying on `List`'s `toString` and to do it?

cwbriones: ah. I had this in while running the tests, it was a little bit easier to spot when using the hamcrest matchers but I did intend to revert that.

PR #154, opened 2019-01-29
Added rollup metric name filtering, add default configuration and made rollups a feature.

ddimensia:
I added regex based whitelist/blacklist support.  The behavior is that if a metric isn't in either list then it is accepted, if it's in the blacklist then it is filtered out unless it is also found in the whitelist.

I also add 'rollups' as a feature in the configuration and provided a default configuration.

### DIFF
--- a/app/com/arpnetworking/rollups/MetricsDiscovery.java
+++ b/app/com/arpnetworking/rollups/MetricsDiscovery.java
@@ -30,8 +30,10 @@ import scala.concurrent.duration.FiniteDuration;
 
 import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import javax.inject.Inject;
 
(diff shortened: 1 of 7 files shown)


conf/portal.application.conf
```diff
@@ -231,6 +234,14 @@ expressionRepository {
   expressionQueryGenerator.type = "com.arpnetworking.metrics.portal.expressions.impl.DatabaseExpressionRepository$GenericQueryGenerator"
 }
 
+# Rollups
+# ~~~~~
+rollups {
+  fetch.interval = "1h"
+  metric.whitelist = [""]
```
BrandonArp: Can you make these empty lists instead of a single element list with an empty string?

ddimensia: Oops, that's what I meant, just spaced out a little.

app/com/arpnetworking/rollups/MetricsDiscovery.java
```diff
@@ -113,13 +118,24 @@ private void updateMetricsSet(final KairosMetricNamesQueryResponse response) {
         return Optional.ofNullable(next);
     }
 
+    private Predicate<String> toPredicate(final List<String> regexList, final boolean defaultResult) {
+        return regexList
+                .stream()
+                .map(r -> "^" + r + "$")
```
BrandonArp: Ville brought up a good point here.  Lets not mess with the regex in a weird way.  Simplicity is probably better here.  Lets just pass the string through unchanged and not create a weird regex dialect by mucking with it.  If the user wants to anchor, lets let them anchor.

PR #156, opened 2019-01-31
Initial UI for creating and editing a report

cwbriones:
@BrandonArp @vjkoskela 

This adds a UI for report creation and modification, without any associated controller behavior. 

screenshot 2019-01-31 13 57 27

I've also included a `ReportController` with some stub implementations (i.e. returning not found or 200 regardless of the request). The reports tab is feature gated in the same manner as hosts and alerts.

It's also an open question of whether or not I should hide some of the fields from the UI while keeping them present in the request - for example, I highly doubt that anyone will want to specify the browser event themselves for every report. The customization could be helpful for us on the implementation but once it's working I'd expect it to be locked down. Similar reasoning applies to the PDF sizing, at least for MVP.


cwbriones:
There were some other changes I made (some actual code in the ReportController, API changes to the JobQuery necessary for that), but I left them out to keep this PR slim.

cwbriones:
My typescript-fu is pretty unpracticed, is there a way to group behavior with enum data? The enums I used all have a display representation as well as a request representation and as it stands that logic is a bit scattered. I wanted to group that together but it doesn't look like typescript allows objects within enums or methods on them.

vjkoskela:
Ideally, you could group the unwanted settings behind a collapse or alternatively just marked them as invisible so we can always show them via dev tools if we need to much with something. Either these would be fine by me.

public/html/reports/EditReportViewModel.html
```diff
@@ -0,0 +1,110 @@
...
+                                    </div>
+                                </div>
+                                <hr>
+                            </div>
+                            <div class="dropdown">
+                                <button class="btn btn-default dropdown-toggle"
+                                        type="button"
+                                        id="addRecipientDropdown"
+                                        data-toggle="dropdown"
+                                        aria-haspopup="true" aria-expanded="true">
```
BrandonArp: What is handling the toggle here? Does it make sure to re-set aria-expanded to false when it isn't shown?

app/assets/javascripts/classes/KnockoutBindings.ts
```diff
@@ -47,6 +50,53 @@ module kobindings {
         }
     };
 
+    ko.bindingHandlers['popover'] = {
```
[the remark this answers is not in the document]
cwbriones: So bootstrap is available here for whatever reason but I suspect you're right in that it's probably not being pulled in as a direct dependency here.

I've added an import and I now see it in the generated javascript but I think I'll make an issue to look over its usage since it's possibly being loaded in everywhere.

See https://stackoverflow.com/questions/22762890/typescript-9-5-import-bootstrap-into-an-amd-module-to-use-jquery-extensions

public/html/reports/EditReportViewModel.html
cwbriones: Bootstrap handles this when the dropdown is initialized and adds the appropriate event handlers. It immediately gets set to false on page load but I'll set it to that.

Although the fact that this works also indicates that bootstrap is being loaded into the global scope somehow.

cwbriones: Nevermind, it's not being loaded globally here - it's loaded through the same module as above in`KnockoutBindings.ts`

PR #157, opened 2019-01-31
Refine the ReportRepository API to allow for querying Reports

cwbriones:
The current API only allows for querying for the Reports as Jobs.

`ReportRepository#queryReports` can be used to retrieve the Report instances directly (e.g. for controller pagination). This doesn't require much of an implementation change since we can add a default for `ReportRepository#query` that just wraps `queryReport`.

Alternatively, I could have used wildcard types in JobRepository, but this ends up polluting client-side code and the query result types with wildcards themselves.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/reports/ReportRepository.java
+++ b/app/com/arpnetworking/metrics/portal/reports/ReportRepository.java
@@ -16,9 +16,13 @@
 
 package com.arpnetworking.metrics.portal.reports;
 
+import com.arpnetworking.metrics.portal.scheduling.JobQuery;
 import com.arpnetworking.metrics.portal.scheduling.JobRepository;
 import models.internal.Organization;
+import models.internal.QueryResult;
+import models.internal.impl.DefaultQueryResult;
 import models.internal.reports.Report;
+import models.internal.scheduling.Job;
 
 import java.util.Optional;
 import java.util.UUID;
@@ -45,5 +49,23 @@ public interface ReportRepository extends JobRepository<Report.Result> {
      * @param organization The {@code Organization} which owns the report.
      */
     void addOrUpdateReport(Report report, Organization organization);
+
+    /**
+     * Query reports.
+     *
+     * This is a specialization of {@link JobRepository#query} for cases where you want to
+     * handle the reports themselves, outside of a Job context.
+     *
+     * @param query The {@code JobQuery} instance to execute.
+     * @return The reports resulting from executing the query.
+     */
+    QueryResult<Report> queryReports(JobQuery<Report.Result> query);
+
+    @Override
+    default QueryResult<Job<Report.Result>> query(JobQuery<Report.Result> query) {
+        final QueryResult<Report> reports = queryReports(query);
+        // This re-wrap step is necessary to satisfy the compiler typecheck.
+        return new DefaultQueryResult<>(reports.values(), reports.values().size());
+    }
 }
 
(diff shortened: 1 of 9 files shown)


app/com/arpnetworking/metrics/portal/reports/ReportRepository.java
```diff
@@ -45,5 +49,23 @@
...
+    /**
+     * Query reports.
+     *
+     * This is a specialization of {@link JobRepository#query} for cases where you want to
+     * handle the reports themselves, outside of a Job context.
+     *
+     * @param query The {@code JobQuery} instance to execute.
+     * @return The reports resulting from executing the query.
+     */
+    QueryResult<Report> queryReports(JobQuery<Report.Result> query);
```
vjkoskela: Should this accept a `ReportQuery`? Initially, this could just be an empty subclass of `JobQuery`, but it would allow for report specific query fields to be specified when dealing with reports queries. For example, you could query by title of report (which does not apply to job). Of course this does mean that you would have to be able to create a `ReportQuery` from a `JobQuery` in `query` below. I don't think it's necessary for Spencer's usage, but it depends on what you had in mind for the list user interface.

app/com/arpnetworking/metrics/portal/reports/ReportRepository.java
```diff
@@ -45,5 +49,23 @@
...
+
+    @Override
+    default QueryResult<Job<Report.Result>> query(JobQuery<Report.Result> query) {
+        final QueryResult<Report> reports = queryReports(query);
+        // This re-wrap step is necessary to satisfy the compiler typecheck.
+        return new DefaultQueryResult<>(reports.values(), reports.values().size());
+    }
```
vjkoskela: Should there also be a `createQuery` method?

cwbriones: Yes, you're right I missed that.

app/com/arpnetworking/metrics/portal/reports/ReportRepository.java
cwbriones: I thought this would work but now I'm thinking it shouldn't be a subclass. I agree that we should have a separate type to allow for filtering by Report-specific fields, but as a subclass the `JobQuery#execute` method becomes less useful (if I used a `ReportQuery#execute` it would still return a Job which isn't what I want)

app/models/internal/impl/DefaultJobQuery.java
```diff
@@ -35,7 +35,7 @@
  */
 @Loggable
 public final class DefaultJobQuery<T> implements JobQuery<T> {
-    private static final int DEFAULT_LIMIT = 100;
+    private static final int DEFAULT_LIMIT = 1000;
```
vjkoskela: nit: blank line between static and non-static member declaration.

app/models/internal/impl/DefaultReportQuery.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Public constructor.
+     *
+     * @param repository The {@code ReportRepository} to query against.
+     * @param jobQuery   The corresponding {@code JobQuery}.
+     */
+    public DefaultReportQuery(final ReportRepository repository, final JobQuery<Report.Result> jobQuery) {
+        _repository = repository;
+        _jobQuery = jobQuery;
```
vjkoskela: The job query is mutable. Should we really store its reference instead of copying its fields? Essentially what we've done is created two linked representations of the data.

PR #162, opened 2019-02-04
Job coordinator

speezepearson:
Add a `JobCoordinator` actor, which performs anti-entropy on a `JobRepository`'s executor-actors.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
+++ b/app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
@@ -26,6 +26,7 @@ import com.arpnetworking.metrics.incubator.PeriodicMetrics;
 import com.arpnetworking.metrics.util.PagingIterator;
 import com.arpnetworking.steno.Logger;
 import com.arpnetworking.steno.LoggerFactory;
+import com.google.common.collect.ImmutableSet;
 import com.google.inject.Injector;
 import models.internal.Organization;
 import models.internal.scheduling.Job;
@@ -36,6 +37,7 @@ import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Iterator;
 import java.util.Optional;
+import java.util.function.Supplier;
 
 /**
  * Coordinates a {@link JobRepository}'s {@link JobExecutorActor}s to ensure that exactly one actor exists for each job.
@@ -48,7 +50,7 @@ public final class JobCoordinator<T> extends AbstractPersistentActorWithTimers {
     private final Injector _injector;
     private final Clock _clock;
     private final Class<? extends JobRepository<T>> _repositoryType;
-    private final Organization _organization;
+    private final Supplier<ImmutableSet<Organization>> _organizationsProvider;
     private final ActorRef _jobExecutorRegion;
     private final PeriodicMetrics _periodicMetrics;
 
@@ -60,7 +62,7 @@ public final class JobCoordinator<T> extends AbstractPersistentActorWithTimers {
      * @param <T> The type of result produced by the {@link JobRepository}'s jobs.
      * @param injector The Guice injector to load the {@link JobRepository} from.
      * @param repositoryType The type of the repository to load.
-     * @param organization The {@link Organization} whose jobs to coordinate.
+     * @param organizationsProvider Provides the set of all {@link Organization}s to monitor in the repository.
      * @param jobExecutorRegion The ref to the Akka cluster-sharding region that dispatches to {@link JobExecutorActor}s.
      * @param periodicMetrics The {@link PeriodicMetrics} that this actor will use to log its metrics.
      * @return A new props to create this actor.
@@ -68,10 +70,10 @@ public final class JobCoordinator<T> extends AbstractPersistentActorWithTimers {
     public static <T> Props props(
             final Injector injector,
             final Class<? extends JobRepository<T>> repositoryType,
-            final Organization organization,
+            final Supplier<ImmutableSet<Organization>> organizationsProvider,
             final ActorRef jobExecutorRegion,
             final PeriodicMetrics periodicMetrics) {
-        return props(injector, Clock.systemUTC(), repositoryType, organization, jobExecutorRegion, periodicMetrics);
+        return props(injector, Clock.systemUTC(), repositoryType, organizationsProvider, jobExecutorRegion, periodicMetrics);
     }
 
     /**
@@ -81,7 +83,8 @@ public final class JobCoordinator<T> extends AbstractPersistentActorWithTimers {
      * @param injector The Guice injector to load the {@link JobRepository} from.
      * @param clock The clock the actor will use to determine when the anti-entropy process should run.
      * @param repositoryType The type of the repository to load.
-     * @param organization The {@link Organization} whose jobs to coordinate.
+     * @param organizationsProvider Provides the set of all {@link Organization}s to monitor in the repository.
+     * @param jobExecutorRegion The ref to the Akka cluster-sharding region that dispatches to {@link JobExecutorActor}s.
      * @param periodicMetrics The {@link PeriodicMetrics} that this actor will use to log its metrics.
      * @return A new props to create this actor.
      */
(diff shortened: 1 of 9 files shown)


app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -0,0 +1,237 @@
...
+            final Class<? extends JobRepository<T>> repositoryType,
+            final Organization organization,
+            final ActorRef jobExecutorRegion,
+            final PeriodicMetrics periodicMetrics) {
+        return Props.create(
+                JobCoordinator.class,
+                () -> new JobCoordinator<>(injector, clock, repositoryType, organization, jobExecutorRegion, periodicMetrics));
+    }
+
+    private JobCoordinator(final Injector injector,
```
vjkoskela: nit: wrap to the next line

app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -0,0 +1,237 @@
...
+                        .offset(offset)
+                        .limit(JOB_QUERY_PAGE_SIZE)
+                        .execute()
+                        .values())
+                .build();
+    }
+
+    private void runAntiEntropy() {
+        final Instant startTime = _clock.instant();
+        System.out.println("in runAntiEntropy");
```
vjkoskela: Either use the logger or remove.

app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -0,0 +1,237 @@
...
+        final Iterator<? extends Job<T>> js = getAllJobs();
```
vjkoskela: nit: rename `js`

app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -0,0 +1,237 @@
...
+        js.forEachRemaining(job -> {
+            System.out.println("considering " + job);
```
vjkoskela: Same.

app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -0,0 +1,237 @@
...
+
+        // We now know that all jobs in the repo have current actors.
+        // There might still be actors which don't correspond to jobs, but that's fine:
+        //   they should self-terminate next time they execute.
+
+        _periodicMetrics.recordTimer(
+                "job_coordinator_tick_time",
+                ChronoUnit.NANOS.between(startTime, _clock.instant()),
+                Optional.of(NANOS));
+        System.out.println("done with runAntiEntropy");
```
vjkoskela: Same.

app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    @Override
+    public Receive createReceive() {
+        return receiveBuilder()
+                .match(AntiEntropyTick.class, message -> {
+
+                    LOGGER.info()
```
vjkoskela: `debug` level?

app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -0,0 +1,237 @@
...
+                    getContext().getSystem().scheduler().scheduleOnce(
+                            scala.concurrent.duration.Duration.Zero(),
+                            () -> {
+                                try {
+                                    runAntiEntropy();
+                                } finally {
+                                    getSelf().tell(AntiEntropyFinished.INSTANCE, getSelf());
+                                }
+                            },
+                            getContext().getSystem().dispatcher()); // TODO(spencerpearson): probably not the right dispatcher
```
vjkoskela: @BrandonArp confirm?

app/com/arpnetworking/metrics/util/PagingIterator.java
```diff
@@ -0,0 +1,113 @@
...
+ * </pre>
+ *
+ * @param <E> The type of element yielded by the paging function.
+ *
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public final class PagingIterator<E> implements Iterator<E> {
+
+    private final Function<Integer, List<? extends E>> _getPage;
+    private Iterator<? extends E> _buffer = new ArrayList<E>().iterator();
```
vjkoskela: While this should work, it's a little out of the norm to declare a list and only reference it indirectly via the iterator.

app/com/arpnetworking/metrics/util/PagingIterator.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private void repopulateBufferIfNeeded() {
+        if (_exhaustedSource) {
+            return;
+        }
+        if (_buffer.hasNext()) {
+            return;
+        }
+        synchronized (this) {
```
vjkoskela: If this is expected to be thread safe it's not, as it can return for multiple threads even though there is just one element in the underlying store (not both threads can claim that instance).

vjkoskela: Otherwise, if it's not supposed to be thread safe then there should be no need for synchronized.

app/com/arpnetworking/metrics/util/PagingIterator.java
```diff
@@ -0,0 +1,113 @@
...
+import net.sf.oval.constraint.NotNull;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.function.Function;
+
+/**
+ * Iterator that wraps a paging function, turning a "yield batches of elements at once" interface into a plain {@link Iterator}.
```
vjkoskela: This works for back-end where you can store the iterator instance. The other change we talked about which link the Query and QueryResult to allow for one to continue from the other would be helpful to API/frontend interactions. I wonder if that would simplify the iterator as well?

speezepearson: That would simplify this! I was imagining refactoring a common `Query` interface out of `HostQuery`, `AlertQuery`, and `ReportQuery`, and then writing a method somewhere of a type like `Query<E> -> Iterator<QueryResult<E>>`. But this ends up requiring a nontrivial refactor to deal with slight differences between the `...Query` classes, and... I didn't think it was worth it at this time, given that (in my opinion?) this `PagingIterator` class is conceptually simple and is only slightly less elegant to use.

I could definitely go down the other path, though. I didn't see anything that looked like a dealbreaker, it just seemed harder.

app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -0,0 +1,237 @@
...
+    @Override
+    public Receive createReceiveRecover() {
+        return receiveBuilder()
+                .build();
+    }
+
+    @Override
+    public String persistenceId() {
+        return "com.arpnetworking.metrics.portal.scheduling.JobCoordinator:"
+                + "repositoryType=" + _repositoryType + "; "
```
BrandonArp: Are you using the class name here? If so, please call the appropriate getter for the string you want.

app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
BrandonArp: What is the reason you're kicking this off to a background executor?  Running it on the system dispatcher is likely a bad thing to do.  At least using the context().dispatcher() would let us configure a specific thread pool if we chose to.  Overall, though, I don't like you running this separate from the message processing loop.  For instance, using the _clock variable in the runAntiEntropy() method is already a race condition, and very easy to do by accident in an actor (because you assume that you're already synchronized on the message handling).

speezepearson: > What is the reason you're kicking this off to a background executor?
Pretty much just following the "don't block the actor" rule: iterating over all the jobs might take a long time, so I backgrounded it in order to free up the JobCoordinator to handle other messages. (There aren't _currently_ any other messages it might receive, but that could change.)

I hadn't realized that `_clock` and several other instance variables were being referenced asynchronously, but... it seems fine to me? All the instance variables (except `_currentlyExecuting`) are final and (conceptually) immutable.

I guess it does seem a potentially slippery slope. The best alternative I can think of is to have `runAntiEntropy(...)` not reference any instance variables, instead having the necessary objects passed in, with the javadoc explicitly stating that those objects must be (conceptually) immutable. Acceptable?

speezepearson: Made `runAntiEntropy` static: https://github.com/ArpNetworking/metrics-portal/pull/162/commits/41c9e4e68bd94649479c6d2e77358e6a89144812
(Also wrote a wrapper method that ensures all instance variables are captured at Runnable-_creation_ time rather than Runnable-_execution_ time.)

Also changed to use `context().dispatcher()`. But I left the thread limit the same, because this actor should only ever have one background task running. The dispatcher that all the _executors_ use, though, might want a higher thread count. Intend to sync with you about that after standup.

PR #163, opened 2019-02-05
Refactor organization provider into organization repository with pagination

vjkoskela:


app/com/arpnetworking/metrics/portal/organizations/OrganizationRepository.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    void close();
+
+    /**
+     * Get the organization in the current request.
+     *
+     * @param request request that serves as the context
+     * @return the organization to use for the request
+     */
+    Organization get(Http.Request request);
```
BrandonArp: This doesn't really make sense to me to go in the Repository.  It's a different concept than the storage of the organization.

vjkoskela: I completely agree. However, what's missing is that middle layer. It has to go somewhere for now, and that somewhere must join a repository with the http request.

vjkoskela: ^ I don't think this is any worse than before.

PR #168, opened 2019-03-29
JDBC clustering

vjkoskela:
JDBC Clustering Extension:
https://github.com/dnvriend/akka-persistence-jdbc


conf/portal.application.conf
```diff
@@ -272,14 +331,46 @@ akka {
...
+      handoff-timeout="60 s"
+      rebalance-interval="10 s"
+      snapshot-interval="720 s"
+      state-store-mode="persistence"
+      least-shard-allocation-strategy {
+        rebalance-threshold=10
+        max-simultaneous-rebalance=3
+      }
     }
-    serialization-bindings {
```
BrandonArp: Any particular reason you dropped this?

vjkoskela: I don't think it's necessary, is it?

BrandonArp: In that case, the serializers config shouldn't be required as well.  Please remove it

PR #178, opened 2019-04-24
fix rollup singleton proxy

BrandonArp:
also fix test issues and a @Nullable error


app/global/MainModule.java
```diff
@@ -460,7 +462,7 @@ public ActorRef get() {
         @Override
         public ActorRef get() {
             final Cluster cluster = Cluster.get(_system);
-            final int actorCount = _configuration.getInt("rollups.worker.count");
+            final int actorCount = _configuration.getInt("rollup.worker.count");
```
ddimensia: Were all the other config references to 'rollup.' and not 'rollups.'?

BrandonArp: Correct.  They were all referencing "rollup"

PR #190, opened 2019-05-02
Report list view frontend

cwbriones:
Adds a paginated list view for the reports page, and factors out shared `Base*ViewModel` types from both the `EditReportViewModel` and `ReportViewModel`.

~An open problem is how to unify the html template with the pager from `AlertsViewModel`. The durandal components were incompatible because of how the template is structured; this has several templatized portions while `AlertsViewModel` has only one. I had attempted to extract out the entity-agnostic functionality but due to the way durandal binds components after the DOM is created, `<tr>` elements seem to disappear or get hoisted upwards since they are not yet enclosed in a `<table>`.~

UPDATE: In somewhat of a hack, I was able to include the new paging template after all by embedding the row elements into a temporary table that is only present at component binding time (and removed after by knockout bindings). See `ReportsViewModel.html` and `components/Pager.html` for details.

See #191 for the associated backend changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,7 @@ eclipse-bin/
 
 # Akka test persistence data
 test-snapshots/
+
+# VSCode
+.vscode/
+.project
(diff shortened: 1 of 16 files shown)


cwbriones:
Screenshots:
Screenshot 2019-05-01 19 56 30

Screenshot 2019-05-01 19 56 37

PR #191, opened 2019-05-02
Report Controller CRUD and View Models

cwbriones:
This PR adds view models and associated internal model mappings to the reports feature. It also adds controller endpoints for creating, updating, and querying reports.

As a consequence of this mapping, ETag fields on the internal `Report` (and thus `Job`) were made optional since they are unavailable when created from a view model.

See #190 for the associated client code.


app/models/internal/scheduling/Job.java
```diff
@@ -41,9 +42,9 @@
     /**
      * Gets an <a href="https://en.wikipedia.org/wiki/HTTP_ETag">ETag</a> that changes each time the job changes.
      *
-     * @return The ETag.
+     * @return The ETag, if any.
      */
-    String getETag();
+    Optional<String> getETag();
```
cwbriones: Right now ETags are computed in [models.ebean.Report::toInternal](https://github.com/ArpNetworking/metrics-portal/blob/master/app/models/ebean/Report.java#L203) since they use data model specific fields such as  `lastModified`. This creates a problem where a view model cannot be mapped into an internal model and persisted without an etag - but an etag cannot be computed without already having persistence metadata.

To resolve this dependency I just went ahead and codified the behavior - we don't have an etag in the view model case so the model shouldn't mark it as required. This means cached jobs will always reload if there is no etag to compare. Another possibility was bundling the etag with the Repository query methods - since those are really the only locations where we should have a tag - but those changes would have been much more widespread.

app/com/arpnetworking/metrics/portal/scheduling/impl/NeverSchedule.java
```diff
@@ -19,7 +19,7 @@
 import java.util.Optional;
 
 /**
- * Schedule that should never fire.
+ * BaseScheduleViewModel that should never fire.
```
vjkoskela: Refactoring error?

app/com/arpnetworking/metrics/portal/scheduling/impl/OneOffSchedule.java
```diff
@@ -21,7 +21,7 @@
 import java.util.Optional;
 
 /**
- * Schedule that should be executed exactly once.
+ * BaseScheduleViewModel that should be executed exactly once.
```
vjkoskela: Refactoring error?

app/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicSchedule.java
```diff
@@ -29,7 +29,7 @@
 import java.util.Optional;
 
 /**
- * Schedule for a job that repeats periodically.
+ * BaseScheduleViewModel for a job that repeats periodically.
```
vjkoskela: Refactoring error?

app/controllers/ReportController.java
```diff
@@ -47,8 +56,8 @@
     /**
      * Public constructor.
      *
-     * @param configuration Instance of Play's {@code Config}.
-     * @param reportRepository Instance of {@link ReportRepository}.
+     * @param configuration          Instance of Play's {@code Config}.
```
vjkoskela: nit: don't tab indent the param descriptions.

app/controllers/ReportController.java
```diff
@@ -101,17 +165,21 @@ public Result query(
...
-        return notFound();
-    }
-
-    /**
-     * Delete a specific report.
-     *
-     * @param id The identifier of the report.
-     * @return No content if successful, otherwise an HTTP error code.
-     */
-    public Result delete(final String id) {
```
vjkoskela: Why was delete removed?

app/controllers/ReportController.java
```diff
@@ -101,17 +165,21 @@ public Result query(
...
-        return notFound();
+        final Organization org;
+        try {
+            org = _organizationRepository.get(request());
+        } catch (final NoSuchElementException e) {
+            return notFound();
```
vjkoskela: The method above combines the org lookup with the add or update returning a 5xx if the org lookup fails. Here you are returning a 4xx if the org is not found, which can be ambiguous since you are looking for a report. I think you should separate the different error paths above, and here probably map no organization to a 5xx and no report to a 4xx. @BrandonArp ?

app/models/ebean/Report.java
```diff
@@ -196,7 +196,7 @@ public void setOrganization(final Organization value) {
 
         // The ETag should change if:
         // - the Report changes (in which case, its timestamp will change); or
-        // - the Source changes (in which case, its timestamp will change); or
+        // - the BaseSourceViewModel changes (in which case, its timestamp will change); or
```
vjkoskela: Refactoring error?

app/models/internal/impl/DefaultReport.java
```diff
@@ -60,8 +62,8 @@ public UUID getId() {
     }
 
     @Override
-    public String getETag() {
-        return _eTag;
+    public Optional<String> getETag() {
+        return Optional.ofNullable(_eTag);
```
vjkoskela: This should be an `Optional<...>` on the member by convention instead of wrapping on access. I realize this is against IDEA advice.

app/models/internal/impl/DefaultReport.java
```diff
@@ -157,7 +160,7 @@ public Builder setId(final UUID id) {
         }
 
         /**
-         * Set the report ETag. Required. Cannot be null or empty.
+         * Set the report ETag. Optional. Cannot be null or empty.
```
vjkoskela: The setter argument should be marked `@Nullable`.

vjkoskela: And the constraint "Cannot be null" should be removed.

app/models/internal/impl/PdfReportFormat.java
```diff
@@ -86,7 +87,9 @@ public int hashCode() {
         return Objects.hash(_widthInches, _heightInches);
     }
 
+    @JsonProperty("widthInches")
```
vjkoskela: Why is the internal model annotated for JSON serialization?

app/models/internal/scheduling/Job.java
vjkoskela: Just so we're clear it's not that the existing models have etag; because they don't. It seems more a matter that the etag pattern was never completely flushed out. In fact, etags appear to only exist to represent the state of a collection of objects for the organization (which I don't think are themselves fully utilized).

We should be using the version for concurrent modification control on database backed entities. However, Report does not have a version column. The version in the database case can also be used as the basis of the etag (e.g. together with id and org id; ideally, hashed in some way to obfuscate the data). However, it's debatable whether we should do that (use repository specific fields in the etag computation).

If we compute the etag from private database fields then it must be on the internal model and must be optional. The problem with this is that it can lead to brittle logic where two absent etag values are considered equal (this code review does not make the mistake; but it could be a source of problems). Alternatively, we relegate the etag to the view layer and compute it from the contents (perhaps even from the view model). This goes a bit against the current design where etags are surface from the repositories (although that may not be correct).

Given that we already compute the etags at the repository level, I think it's safe to continue that until we need to refactor the pattern. Under that plan, I think you're solution is the correct one. The only change I would like to see is to add version and use that to compute the etag. However, this change can be done in a separate code review as this review is already rather lengthy the it would address functionality already committed to master.

app/models/view/impl/ChromeScreenshotReportSource.java
```diff
@@ -0,0 +1,148 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
+import models.view.reports.ReportSource;
+
+import java.net.URI;
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * Internal model for a report source that pulls screenshots from a Chrome web browser.
```
vjkoskela: This is not an internal model

app/models/view/impl/HtmlReportFormat.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package models.view.impl;
+
+import com.arpnetworking.logback.annotations.Loggable;
+import com.google.common.base.MoreObjects;
+import models.view.reports.ReportFormat;
+
+/**
+ * An HTML report format.
+ *
```
vjkoskela: Add view model comment: `Play view models are mutable.`

app/models/view/impl/HtmlReportFormat.java
```diff
@@ -0,0 +1,52 @@
...
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+@Loggable
+public final class HtmlReportFormat implements ReportFormat {
+    private static final HtmlReportFormat INSTANCE = new HtmlReportFormat();
```
vjkoskela: Unused.

app/models/view/impl/OneOffSchedule.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2018 Dropbox, Inc.
```
vjkoskela: 2019?

app/models/view/impl/OneOffSchedule.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    public com.arpnetworking.metrics.portal.scheduling.impl.OneOffSchedule toInternal() {
+        return new com.arpnetworking.metrics.portal.scheduling.impl.OneOffSchedule.Builder()
+                .setRunAtAndAfter(_runAt)
+                .build();
+    }
+
+    /**
+     * Create a {@code OneOffSchedule} from its internal representation.
+     * @param schedule The internal model.
```
vjkoskela: nit: blank line between description and annotations

app/models/view/impl/OneOffSchedule.java
```diff
@@ -0,0 +1,60 @@
...
+     * @return The view model.
+     */
+    public static OneOffSchedule fromInternal(final com.arpnetworking.metrics.portal.scheduling.impl.OneOffSchedule schedule) {
+        final OneOffSchedule viewSchedule = new OneOffSchedule();
+        viewSchedule.setRunAt(schedule.getRunAtAndAfter());
+        return viewSchedule;
+    }
+
+    @JsonProperty("runAt")
```
vjkoskela: Is this really necessary? Our object mapper should be preferring accessor based serialization.

app/models/view/impl/PeriodicSchedule.java
```diff
@@ -0,0 +1,100 @@
...
+        final Periodicity period =
+                Periodicity.fromValue(schedule.getPeriod())
+                        .orElseThrow(() -> new IllegalArgumentException("No corresponding schedule period for " + schedule.getPeriod()));
+        viewSchedule.setPeriod(period);
+        viewSchedule.setOffset(schedule.getOffset());
+        viewSchedule.setZone(schedule.getZone());
+        return viewSchedule;
+    }
+
+    @JsonProperty("runAtAndAfter")
```
vjkoskela: Same question; these should be accessor based.

app/models/view/impl/PdfReportFormat.java
```diff
@@ -0,0 +1,101 @@
...
+        return Float.compare(that._widthInches, _widthInches) == 0
+                && Float.compare(that._heightInches, _heightInches) == 0;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(_widthInches, _heightInches);
+    }
+
+    private float _widthInches;
```
vjkoskela: These do not have the annotation mentioned before and after on other view model classes.

app/models/view/reports/EmailRecipient.java
```diff
@@ -0,0 +1,54 @@
...
+import models.internal.impl.DefaultEmailRecipient;
+
+/**
+ * View model for a recipient specified by an email address.
+ *
+ * See {@link DefaultEmailRecipient} for the equivalent internal model.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public class EmailRecipient extends Recipient {
```
vjkoskela: `final`

app/models/view/reports/Recipient.java
```diff
@@ -0,0 +1,88 @@
...
+        final models.internal.reports.Recipient.Visitor<Recipient> visitor = new models.internal.reports.Recipient.Visitor<Recipient>() {
+            @Override
+            public Recipient visit(final DefaultEmailRecipient emailRecipient) {
+                return EmailRecipient.fromInternal(emailRecipient, format);
+            }
+        };
+        return visitor.visit(recipient);
+    }
+
+    @JsonProperty("id")
```
vjkoskela: Same.

app/models/view/reports/Report.java
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * View model for a {@link models.internal.reports.Report}.
+ *
+ * Play view models are mutable.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class Report {
+    private Report() {
```
vjkoskela: The other view model constructors are not private. This will probably make testing harder, and Jackson is going to call the private ctor anyway. If you want this pattern, we should discuss at applying it across the board.

app/models/view/reports/Report.java
```diff
@@ -0,0 +1,146 @@
...
+    private static Stream<Recipient> recipientsFromEntry(
+            final Map.Entry<models.internal.reports.ReportFormat,
+            Collection<models.internal.reports.Recipient>> entry
+    ) {
+        final models.internal.reports.ReportFormat format = entry.getKey();
+        final ReportFormat viewFormat = ReportFormat.fromInternal(format);
+        return entry.getValue().stream().map(r -> Recipient.fromInternal(r, viewFormat));
+    }
+
+    @JsonProperty("id")
```
vjkoskela: Same

app/models/view/reports/ReportFormat.java
```diff
@@ -0,0 +1,67 @@
...
+ * View model of {@link models.internal.reports.ReportFormat}. Play view models are mutable.
+ *
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+@JsonTypeInfo(
+        use = JsonTypeInfo.Id.NAME,
+        include = JsonTypeInfo.As.PROPERTY,
+        property = "type")
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = HtmlReportFormat.class, name = "Html"),
```
vjkoskela: Can we make the names `HTML` and `PDF`?

app/models/view/reports/ReportSource.java
```diff
@@ -0,0 +1,58 @@
...
+ * View model of {@link models.internal.reports.ReportSource}. Play view models are mutable.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+@JsonTypeInfo(
+        use = JsonTypeInfo.Id.NAME,
+        include = JsonTypeInfo.As.PROPERTY,
+        property = "type")
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = ChromeScreenshotReportSource.class, name = "ChromeScreenshot"),
```
vjkoskela: `CHROME_SCREENSHOT`?

app/models/view/scheduling/Periodicity.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private static final Map<String, Periodicity> BY_NAME = Maps.newHashMap();
+
+    static {
+        for (Periodicity p : Periodicity.values()) {
+            BY_NAME.put(p._name, p);
+        }
+    }
+
+    private final String _name;
```
vjkoskela: Java enum's already have a "name"; we should avoid redefining that concept.

vjkoskela: Also, enums in the internal package are frequently used in the view model and storage model directly because they are enumerations (e.g. domain identifiers w/o state). Such a strategy with this enum may simplify the implementation considerably.

app/models/view/scheduling/Periodicity.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Get an instance of {@link Periodicity} that is equivalent to the given value.
+     *
+     * @param value An equivalent instance of {@link ChronoUnit}
+     * @return An optional of the {@code Periodicity} with this value, otherwise {@link Optional#empty()}.
+     */
+    public static Optional<Periodicity> fromValue(final ChronoUnit value) {
+        return BY_NAME.values()
```
vjkoskela: Why not have a reverse map?

app/models/internal/impl/PdfReportFormat.java
cwbriones: Unintentional, reverted.

app/com/arpnetworking/metrics/portal/scheduling/impl/NeverSchedule.java
cwbriones: Yes (to this and all similar changes) - reverted.

app/models/internal/impl/DefaultReport.java
cwbriones: I'm all for consistency even if it goes against convention. I'd rather that happen then have a ton of differing styles.

app/models/view/impl/OneOffSchedule.java
cwbriones: I think I had a different issue and incorrectly attributed it to this; I had forgotten that Jackson uses those for field discovery. Reverted.

app/models/view/scheduling/Periodicity.java
cwbriones: I wasn't aware of that convention - I had actually been thinking this was a lot of overhead since this class already essentially exists. I'll try refactoring this to use that then.

app/models/view/reports/ReportSource.java
cwbriones: I had kept it Pascalcase (as with `HTML` and `PDF`) because typescript constants are styled like that. I suppose I could keep it capital so that the types are the same as on other endpoints, I just thought the translation code is more readable on the java side with these annotations (vs. handrolled on the typescript side).

app/controllers/ReportController.java
cwbriones: I would rather change the other path to explicitly return 404 to match this and include a message with not found indicating the missing entity. It seems incorrect to return 5XX when the request is at fault.

app/controllers/ReportController.java
cwbriones: There isn't currently a delete method on the repository. I can add that in and plumb it through here if you'd like that in this PR, or I can handle that afterwards.

app/models/internal/scheduling/Job.java
cwbriones: It's confusing to me to state that they don't have etags when they're present as fields of the class but I understand what you're saying. I think of etags as additional functionality that wraps a type, so I think we're in agreement when you say _"etags appear to only exist to represent the state of a collection of objects for the organization"_; they're an optimization.

> Under that plan, I think you're solution is the correct one. The only change I would like to see is to add version and use that to compute the etag.

Sure, just file an Issue/task so we can track that.

app/controllers/ReportController.java
vjkoskela: After is fine.

app/models/view/reports/ReportSource.java
vjkoskela: Not a problem as long as it's not contradicting any existing usage in MP and we keep it consistent.

PR #192, opened 2019-05-07
Add a delete endpoint for reports

cwbriones:
Adds `ReportRepository#deleteReport` as well as the associated controller endpoint and typescript client functionality.

This includes two other changes to the ReportController to make its behavior a little more consistent:
* Update the controller endpoints to all return 500 when an organization is not found rather
than a mix of 404 and 500.
* Bind UUID directly to the path parameters for report endpoints to reduce manual validation within each controller action.


app/com/arpnetworking/metrics/portal/reports/ReportRepository.java
```diff
@@ -43,6 +43,16 @@
...
 
+    /**
+     * Delete a {@code Report} given its identifier.
+     *
+     * @param identifier The {@code Report} identifier.
+     * @param organization The {@code Organization} which owns the report.
+     * @throws javax.persistence.PersistenceException if the delete was unsuccessful.
+     * @return the number of reports deleted. Should be 0 or 1.
+     */
+    int deleteReport(UUID identifier, Organization organization);
```
cwbriones: Returning an int isn't the most intuitive thing IMO, I would rather this throw `EntityNotFound` or return an Optional (maybe containing the deleted report?) if unsuccessful. That seems like it would be harder to mess up in implementation / error handling. I only did it like this since I rank consistency with the existing repositories higher in my considerations.

cwbriones:
Renamed to WIP since I want to add Repository tests.

app/com/arpnetworking/metrics/portal/reports/ReportRepository.java
vjkoskela: First, you probably shouldn't throw `javax.persistence.PersistenceException` since that seems implementation specific. 

Second, I totally agree with you that this interface definition is borked. I would prefer to see it return the deleted entity although if that's not possible just return nothing and throw on failure. However, we would need to apply that refactoring all the `delete` repository interfaces for consistency. If you want to do that here, I'm fine with it, otherwise, we can file an issue and refactor it later.

app/controllers/ReportController.java
```diff
@@ -162,24 +170,55 @@ public Result query(
...
+            organization = _organizationRepository.get(request());
         } catch (final NoSuchElementException e) {
-            return notFound();
+            return internalServerError();
+        }
+        try {
+            final Optional<Report> report = _reportRepository.getReport(id, organization);
+            return report.map(r -> ok(Json.toJson(models.view.reports.Report.fromInternal(r))))
+                    .orElseGet(ReportController::notFound);
+        } catch (final PersistenceException e) {
```
vjkoskela: At the moment the repositories throw runtime exceptions for unexpected failures which fall through the controllers and result in an automatic 5xx from Play.

vjkoskela: We really shouldn't be declaring or catching runtime exceptions.

https://docs.oracle.com/javaee/7/api/javax/persistence/PersistenceException.html

app/com/arpnetworking/metrics/portal/reports/ReportRepository.java
vjkoskela: Opinion: Runtime exceptions should not be declared thrown on the signature or on the Javadoc because you should not take logical action based on runtime exceptions. There is unfortunately a lot of misclassification of exceptions in the Java environment, where handle-able exceptions are often typed as runtime exceptions because it is perceived as giving programmers more choice.

In the case of this project, the repository interfaces throw exceptions but only those they don't expect clients to handle. For example, delete should not throw if the item to be deleted does not exist (it should return 0). Delete should throw a runtime exception subclass if the delete itself fails.

There is a perfectly reasonable argument to be made that the repositories should throw checked exceptions on failure to allow clients to handle and recover appropriately. I'm not against that, we just don't have a use case for it in the project and as such it was punted because designing the right granularity of exceptions across all repository interfaces and implementations was deemed to have low ROI.

Happy to discuss this more.

app/controllers/ReportController.java
vjkoskela: See above.

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportRepository.java
```diff
@@ -176,6 +176,47 @@ public void close() {
...
+            // CHECKSTYLE.OFF: IllegalCatchCheck
+        } catch (final RuntimeException e) {
+            // CHECKSTYLE.ON: IllegalCatchCheck
+            LOGGER.error()
+                    .setMessage("Failed to delete report")
+                    .addData("uuid", identifier)
+                    .addData("organization.uuid", organization.getId())
+                    .setThrowable(e)
+                    .log();
+            throw new PersistenceException(e);
```
vjkoskela: Is there a reason we wrap the exception instead of rethrowing? (not sure what the common practice in this project is at the moment)

Also, as a general rule of thumb you should either log or throw, but not both. If you throw, the caller can decide the severity and log (or not if they act) and your log would just confuse the matter. If you swallow the exception (e.g. recover, default, etc.), you need to log something at an appropriate level to indicate what happened.

I bet that Play will log the uncaught exception from the handler.

vjkoskela: ^ Either that or remove the throw and return 0 (although I'm less of a fan of that approach)

cwbriones:
Updated to 66c869f

I decided to write the update query itself instead of relying on ebean's generation of the soft-delete, since it requires an additional read that is otherwise unnecessary. The logging and rethrow code is removed in favor of letting the exception propagate but the other repository methods are not updated to do the same in this commit. 

On the controller side, it looks like Play does automatically respond 500 on uncaught exceptions and logs them as well, as seen in `globals/ErrorHandler` so I removed some of the manual error handling in the delete endpoint.

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportRepository.java
```diff
@@ -186,35 +189,23 @@ public int deleteReport(final UUID identifier, final Organization organization)
...
+            _ebeanServer.update(models.ebean.Report.class)
+                    .set("deleted", true)
+                    .where()
+                    .eq("uuid", identifier)
+                    .eq("organization.uuid", organization.getId())
+                    .eq("deleted", false)
+                    .update();
+
+        LOGGER.debug()
+                .setMessage("Deleted report")
```
vjkoskela: This may be misleading if `deleted == 0`

PR #195, opened 2019-05-14
Add status endpoint to Metrics Portal in the ssame way as MAD and CAGG.

vjkoskela:
Produces:

```
{
"name": "Metrics Portal",
"version": "0.9.6-SNAPSHOT",
"sha": "0c270c278cfccb4878aee9ccd78d7700832cb287"
}
```


vjkoskela:
I completely agree @ddimensia; however, this matches the default path in MAD and CAGG as well. If we agree with `/version` I can make Metrics Portal that now, and MAD can be fixed with 2.x release later this year. Not sure when we'll be able to make CAGG consistent. @BrandonArp thoughts on compatibility and impact?

vjkoskela:
Yes, I agree we need to figure out the problem, although I don't see why you think this is unexpected nor should we block unrelated development. What is unexpected is that this did not fail earlier when the tests in question were introduced. 

Like any test flakiness, I strongly believe this should not block unrelated development. I will look into the CSRF problem, in fact I'd like to understand tomorrow why you think it should not be triggered, but in the meantime this diff should land unless there is something in here you think is changing that behavior (otherwise it can block anyone's diff and bring all development to a grind). 

As for the work around, with a flaky test where you would disable, this test only disable CSRF on the routes while the service is under test. It's better than an `@Ignore`.

PR #196, opened 2019-05-14
Eliminate `Recipient` polymorphism

speezepearson:
Make there be only a single Recipient class. This seems safe because there's no variation between the implementations we can foresee: all Recipients boil down to "a Recipient with `____` handle `<string>`", e.g. "a Recipient with `email` handle `speezepearson@mailinator.com`" or "a Recipient with `slack` handle `speezepearson`.

Old architecture: there is an `interface Recipient` which is implemented by `class EmailRecipient` etc.
New architecture: there is a single `class Recipient`, which has a field of type `enum RecipientType`.


app/models/internal/reports/Recipient.java
```diff
@@ -16,62 +16,125 @@
...
+import java.util.Objects;
 import java.util.UUID;
 
 /**
  * Internal model for a recipient, i.e. a destination.
  *
  * @author Christian Briones (cbriones at dropbox dot com)
  */
-public interface Recipient {
+public final class Recipient {
```
vjkoskela: I think the other internal models still have an interface and then a `DefaultFoo` implementation.

app/models/view/reports/Recipient.java
```diff
@@ -29,14 +27,7 @@
...
  */
-@JsonTypeInfo(
-        use = JsonTypeInfo.Id.NAME,
-        include = JsonTypeInfo.As.PROPERTY,
-        property = "type")
-@JsonSubTypes({
-        @JsonSubTypes.Type(value = EmailRecipient.class, name = "EMAIL"),
-})
-public abstract class Recipient {
+public class Recipient {
```
vjkoskela: Are there any changes necessary the UI with this change in polymorphism of the view model recipient?

app/models/internal/impl/DefaultRecipient.java
```diff
@@ -72,11 +75,13 @@ public int hashCode() {
...
-     * Builder implementation that constructs {@code DefaultEmailRecipient}.
+     * Builder implementation that constructs {@code DefaultRecipient}.
      */
-    public static final class Builder extends OvalBuilder<DefaultEmailRecipient> {
+    public static final class Builder extends OvalBuilder<DefaultRecipient> {
         @NotNull
         private UUID _id;
+        @NotNull
+        private RecipientType _type;
         @Email
```
vjkoskela: This should not longer be validated as an email address.

vjkoskela:
Looks great, just remove the one leftover validation constraint and I'll merge this in the morning!

PR #199, opened 2019-05-17
dont use cookies in test

BrandonArp:


vjkoskela:
I think you can also set the cookie policy on a per request basis; not sure if there's a way to change the default policy on the client.

https://hc.apache.org/httpclient-3.x/cookies.html

PR #201, opened 2019-05-20
Have `DefaultReport` actually render+send upon execution

speezepearson:
- Define how to render reports:
    - Add the `Renderer` interface to represent a mechanism for rendering a report into a particular kind of format (e.g. `ChromeWebRenderer`, which is not yet implemented).
    - Add `RenderedReport` interface to represent the result.
    - Add `Renderers` utility class to encapsulate the logic of choosing a renderer and rendering a report into many formats at once.
    - Add visitor pattern to `ReportSource` (emulating `ReportFormat`), to enable a mapping from (sourceType, formatType) -> rendererType.

- Define how to send reports:
    - Add the `Sender` interface to represent a mechanism for sending rendered reports to given recipients (e.g. `EmailSender`, which is not yet implemented).
    - Add `Senders` utility class to encapsulate the logic of choosing senders and sending reports to many recipients at once.

- Fill in the previously-noop implementation of `DefaultReport.execute()`:
    - Make it take a Guice `Injector`, so it can choose the appropriate renderers/senders.
        (This is actually a change to the `Job.execute` interface.)


app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
```diff
@@ -0,0 +1,184 @@
...
+                .stream()
+                .collect(ImmutableSetMultimap.flatteningToImmutableSetMultimap(
+                        Map.Entry::getKey,
+                        e -> e.getValue().stream()));
+        final ImmutableMultimap<Recipient, ReportFormat> recipientToFormats = formatToRecipients.inverse();
+        return renderAll(injector, formatToRecipients.keySet(), report.getSource(), scheduled)
+                .thenCompose(
+                        formatToRendered -> sendAll(injector, recipientToFormats, formatToRendered)
+                ).thenApply(
+                        nothing -> null
```
vjkoskela: Can you elaborate on what this is doing?

vjkoskela: As discussed if this is temporary and if it's easy let's just type the CompletionStage to `Void` for now.

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
```diff
@@ -0,0 +1,184 @@
...
+    /**
+     * Render and send a report.
+     *
+     * @param report The report to render and execute.
+     * @param injector A Guice injector to pull dependencies out of.
+     * @param scheduled The instant the report was scheduled for.
+     * @return A CompletionStage that completes when sending has completed for every recipient.
+     */
+    public static CompletionStage<Report.Result> execute(final Report report, final Injector injector, final Instant scheduled) {
+        verifyDependencies(report, injector);
```
vjkoskela: This seems unnecessary. It will either pass or fail every time. Imo we should not add defensive runtime code on each execution for what is a configuration problem. I expect the job will fail gracefully if it cannot load a renderer or sender when it's actually executing. I assume this is more of a all or none type failure protection, I'm not particularly concerned because it's config based not data based.

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+
+    /* package private */ static CompletionStage<ImmutableMap<ReportFormat, RenderedReport>> renderAll(
+            final Injector injector,
+            final ImmutableSet<ReportFormat> formats,
+            final ReportSource source,
+            final Instant scheduled
+    ) {
+        final Map<ReportFormat, RenderedReport> result = Maps.newHashMap(); // TODO(spencerpearson) -- ConcurrentMap?
```
vjkoskela: Yes. You are populating the map concurrently as the result of each render future is executed. Nothing guarantees synchronous access to the map.

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
```diff
@@ -0,0 +1,184 @@
...
+     */
+    private static <K, V> ImmutableMap<K, V> mask(final ImmutableMap<K, V> map, final Collection<K> keys) {
+        return keys.stream().collect(ImmutableMap.toImmutableMap(key -> key, map::get));
+    }
+
+    private ReportExecution() {}
+
+    private static final ReportSource.Visitor<String> SOURCE_TYPE_VISITOR = new ReportSource.Visitor<String>() {
+        @Override
+        public String visit(final ChromeScreenshotReportSource source) {
```
vjkoskela: We're going to be renaming this source right?

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
```diff
@@ -0,0 +1,184 @@
...
+            return "web";
+        }
+    };
+    private static final ReportFormat.Visitor<String> FORMAT_TYPE_VISITOR = new ReportFormat.Visitor<String>() {
```
vjkoskela: Should `ReportFormat` not just have a `name()` method or some such? Same question above about `ChromeScreenshotReportSource`?

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
```diff
@@ -0,0 +1,184 @@
...
+        final String keyName = getSenderKeyName(recipient);
+        final Sender sender = injector.getInstance(Key.get(Sender.class, Names.named(keyName)));
+        if (sender == null) {
+            throw new IllegalArgumentException("no Sender exists for key name '" + keyName + "'");
+        }
+        return sender;
+    }
+
+    /* package private */ static String getSenderKeyName(final Recipient recipient) {
+        return CaseFormat.UPPER_UNDERSCORE.to(CaseFormat.LOWER_UNDERSCORE, recipient.getType().name());
```
vjkoskela: I would prefer that we just had a member or method on `RecipientType` that returned or computed the key for a given type.

app/com/arpnetworking/metrics/portal/reports/Sender.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Send some {@link RenderedReport} documents to a {@link Recipient}.
+     *
+     * @param recipient The recipient to notify.
+     * @param formatToRendered The reports to send.
+     * @return A CompletionStage that completes when the transmission has completed.
+     */
+    CompletionStage<Void> send(
+            Recipient recipient,
+            ImmutableMap<ReportFormat, RenderedReport> formatToRendered
```
vjkoskela: `formatToRendered` => `formatsToSend`?

app/models/internal/impl/DefaultRecipient.java
```diff
@@ -73,6 +74,15 @@ public int hashCode() {
         return Objects.hash(_id, _address);
     }
 
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                .add("_id", _id)
+                .add("_type", _type)
+                .add("_address", _address)
```
vjkoskela: We don't include the underscores in the field names.

app/models/internal/impl/DefaultRenderedReport.java
```diff
@@ -0,0 +1,165 @@
...
+    public int hashCode() {
+        int result = Objects.hash(_scheduledFor, _generatedAt, _format);
+        result = 31 * result + Arrays.hashCode(_bytes);
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                .add("_bytes", _bytes)
```
vjkoskela: We don't include the underscores in the field names.

app/models/internal/impl/DefaultRenderedReport.java
```diff
@@ -0,0 +1,165 @@
...
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.Objects;
+
+/**
+ * A document containing the results of a particular rendering of a particular {@link models.internal.reports.Report}.
+ *
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public final class DefaultRenderedReport implements RenderedReport {
```
vjkoskela: Please annotate with `@Loggable`

app/models/internal/impl/DefaultRenderedReport.java
```diff
@@ -0,0 +1,165 @@
...
+        private byte[] _bytes;
+        @NotNull
+        private Instant _scheduledFor;
+        @NotNull
+        private Instant _generatedAt;
+        @NotNull
+        private ReportFormat _format;
+    }
+
+    private DefaultRenderedReport(final Builder builder) {
```
vjkoskela: By convention this appears above after other class methods but before the members.

app/models/internal/impl/DefaultRenderedReport.java
```diff
@@ -0,0 +1,165 @@
...
+        _bytes = builder._bytes;
```
vjkoskela: Using `byte[]` has problems because it is mutable. 

There are no good answers to this:

https://stackoverflow.com/questions/5480070/immutable-container-for-a-byte-that-supports-subsequences-like-string-substri

However, it's something we should at least document clearly, or if possible consider available alternatives. Definitely not suggesting solving the underlying problem entirely.

app/models/internal/impl/DefaultRenderedReport.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        /**
+         * Set the report bytes. Required. Cannot be null.
+         *
+         * @param bytes The report bytes.
+         * @return This instance of {@code Builder}.
+         */
+        public Builder setBytes(final byte[] bytes) {
+            _bytes = bytes.clone();
```
vjkoskela: Nice! Typically we do this in the ctor.

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
speezepearson: Without this check, `execute()` throws synchronously if it can't find all the Renderers it needs, and asynchronously if it can't find all the Senders it needs. The mixture of [a]synchronous throwing seemed gross to me, and I feel that a synchronous throw is more appropriate here than asynchronous (since the impending failure _can_ be determined quickly).

What counts as a "graceful" failure here? Aborting and logging an exception?

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
speezepearson: Yep! I was hoping to put that in another diff, since it's far-reaching.

app/models/internal/impl/DefaultRenderedReport.java
speezepearson: Hmm. In this case, due to the mutability issue you point out above, I feel that doing it here is more appropriate. (If I expected that we'd call `setBytes()` on the same instance a _whole lot_, I could be convinced that the performance cost would justify moving the `clone()` into the constructor; but since I don't expect that, I think I'd rather keep as few externally-mutable objects around as possible.)

app/com/arpnetworking/metrics/portal/reports/RecipientType.java
```diff
@@ -24,5 +24,15 @@
     /**
      * Email.
      */
-    EMAIL,
+    EMAIL("email");
+
+    public String getName() {
```
vjkoskela: An enum already has `public final String name()`, so this will be very confusing. Please be more clear.

vjkoskela: ^ Or alternatively if these are the same then why do we need a lower case version?

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
vjkoskela: If you want to throw synchronously and want to force users to deal with it then this should probably be a checked exception.

However, because there is a completion stage returned user's will need to deal with its failure no matter what. If you want to check up front, why not just return a failed completion stage? (Why throw and return a failable future?)

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
vjkoskela: No problem. Yes, no one is using this yet.

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
vjkoskela: So this seems like an odd visitor to have just to get the name of the format. Behavior in visitors seems reasonable, but this seems to be returning data that should be evident in the model. What am I missing here wrt our conversation this morning?

app/models/internal/reports/ReportSource.java
```diff
@@ -30,5 +32,39 @@
...
+         */
+        T visit(ChromeScreenshotReportSource source);
+
+        /**
+         * Convenience method equivalent to {@code format.accept(this) }.
+         *
+         * @param format The format to visit.
+         * @return The result of applying the visitor
+         */
+        default T visit(ReportSource format) {
```
vjkoskela: type-o: `format` => `source`

app/models/internal/reports/ReportSource.java
```diff
@@ -30,5 +32,39 @@
...
+     * @param <T> the return type of the visitor.
+     */
+    interface Visitor<T> {
+        /**
+         * Visit a {@link ChromeScreenshotReportSource}.
+         *
+         * @param source The source to visit.
+         * @return The result of applying the visitor.
+         */
+        T visit(ChromeScreenshotReportSource source);
```
vjkoskela: Shouldn't all `ResourceSource` subclasses have an `accept` method? (e.g. `Visitable`)

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
speezepearson: Good point! Fixed: it now only ever throws asynchronously.

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
```diff
@@ -0,0 +1,181 @@
...
+     * @param report The report to render and execute.
+     * @param injector A Guice injector to pull dependencies out of.
+     * @param scheduled The instant the report was scheduled for.
+     * @return A CompletionStage that completes when sending has completed for every recipient.
+     * @throws IllegalArgumentException if any necessary class is missing from the given Injector.
+     */
+    public static CompletionStage<Report.Result> execute(final Report report, final Injector injector, final Instant scheduled) {
+        try {
+            verifyDependencies(report, injector);
+        } catch (final ConfigException exception) {
```
vjkoskela: The documentation below indicates that an `IllegalArgumentException` is thrown.

As an aside, another way to do this would be to chain all these calls together (then the exception type doesn't matter).

app/com/arpnetworking/metrics/portal/reports/ReportExecution.java
```diff
@@ -0,0 +1,181 @@
...
+                .stream()
+                .collect(ImmutableSetMultimap.flatteningToImmutableSetMultimap(
+                        Map.Entry::getKey,
+                        e -> e.getValue().stream()));
+        final ImmutableMultimap<Recipient, ReportFormat> recipientToFormats = formatToRecipients.inverse();
+        return renderAll(injector, formatToRecipients.keySet(), report.getSource(), scheduled)
+                .thenCompose(
+                        formatToRendered -> sendAll(injector, recipientToFormats, formatToRendered)
+                ).thenApply(
+                        nothing -> new DefaultReportResult()
```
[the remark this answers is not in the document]
speezepearson: Added a `TODO` to the `Report.Result` interface. It's currently just a marker, but I agree that at some point we'll want it to be more. (Seems non-MVP, though.)

PR #202, opened 2019-05-20
Tests for ReportController and serialization

speezepearson:
Also several small tweaks:
- Add `equals`/`hashCode`/`toString` to several view-model classes for debugging / testing.
- `HtmlReportFormat` is no longer a singleton, since Jackson creates its own instances anyway.
- Make the `eTag` field on internal reports optional, since internal-models created from view-models may not have an ETag yet.


test/java/com/arpnetworking/metrics/portal/integration/controllers/ReportControllerIT.java
```diff
@@ -0,0 +1,95 @@
...
+        HttpGet getRequest = new HttpGet(WebServerHelper.getUri("/v1/reports/" + reportId));
+        try (CloseableHttpResponse response = WebServerHelper.getClient().execute(getRequest)) {
+            assertEquals(Http.Status.OK, response.getStatusLine().getStatusCode());
+            final Report returnedReport = OBJECT_MAPPER.readValue(response.getEntity().getContent(), models.view.reports.Report.class);
+            assertEquals(writtenReport, returnedReport);
+        }
+
+    }
+
+    private String loadResource(final String suffix) {
```
vjkoskela: Can we refactor this (and its copy in the AlertControllerIT) into com/arpnetworking/metrics/portal/integration/test either into the WebServerHelper or else into a new ResourceHelper?

test/java/com/arpnetworking/metrics/portal/integration/controllers/ReportControllerIT.java
```diff
@@ -0,0 +1,95 @@
...
+
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public final class ReportControllerIT {
+
+    @Test
+    public void testCreateValid() throws IOException {
+        HttpPut putRequest = new HttpPut(WebServerHelper.getUri("/v1/reports"));
+        putRequest.setHeader(new BasicHeader(HttpHeaders.CONTENT_TYPE, "application/json"));
+        putRequest.setEntity(new StringEntity(loadResource("testCreateValid")));
```
vjkoskela: The problem with putting a fixed resource is that the test cannot be run again on the same instance because the data will collide. As per comments in AlertControllerIT we should move to using randomly generated data instead of JSON fixtures. There is already support for many Ebean and internal models in TestBeanFactory and I see no reason not to extend support for view model creation.

vjkoskela: ^ Any specific fields that the test needs to set for its condition should be explicitly set in the test. This way the data that relates to the test is clearly spelled out in the test (it's very non-obvious reading JSON what the relevant part of the JSON blob is to a particular test).

test/java/com/arpnetworking/metrics/portal/integration/controllers/ReportControllerIT.java
```diff
@@ -0,0 +1,95 @@
...
+        try (CloseableHttpResponse response = WebServerHelper.getClient().execute(putRequest)) {
+            assertEquals(Http.Status.NO_CONTENT, response.getStatusLine().getStatusCode());
+        }
+
+        final Report writtenReport = OBJECT_MAPPER.readValue(loadResource("testCreateValid"), models.view.reports.Report.class);
+        final String reportId = writtenReport.getId().toString();
+        HttpGet getRequest = new HttpGet(WebServerHelper.getUri("/v1/reports/" + reportId));
+        try (CloseableHttpResponse response = WebServerHelper.getClient().execute(getRequest)) {
+            assertEquals(Http.Status.OK, response.getStatusLine().getStatusCode());
+            final Report returnedReport = OBJECT_MAPPER.readValue(response.getEntity().getContent(), models.view.reports.Report.class);
```
vjkoskela: There is already a helper in WebServerHelper to read into a JsonNode: `readContentAsJson`. Should we just have an overload that accepts a type-reference/class and invokes OM for you? This should remove some of the boiler plate from all tests.

test/java/com/arpnetworking/metrics/portal/integration/controllers/ReportControllerIT.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Integration tests for {@code ReportController}.
+
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public final class ReportControllerIT {
+
+    @Test
+    public void testCreateValid() throws IOException {
+        HttpPut putRequest = new HttpPut(WebServerHelper.getUri("/v1/reports"));
```
vjkoskela: nit: `final`?

test/java/com/arpnetworking/metrics/portal/integration/controllers/ReportControllerIT.java
```diff
@@ -0,0 +1,95 @@
...
+        putRequest.setHeader(new BasicHeader(HttpHeaders.CONTENT_TYPE, "application/json"));
+        putRequest.setEntity(new StringEntity(loadResource("testCreateValid")));
+        try (CloseableHttpResponse response = WebServerHelper.getClient().execute(putRequest)) {
+            assertEquals(Http.Status.NO_CONTENT, response.getStatusLine().getStatusCode());
+        }
+
+        final Report writtenReport = OBJECT_MAPPER.readValue(loadResource("testCreateValid"), models.view.reports.Report.class);
+        final String reportId = writtenReport.getId().toString();
+        HttpGet getRequest = new HttpGet(WebServerHelper.getUri("/v1/reports/" + reportId));
```
vjkoskela: nit: `final`?

test/java/models/view/ReportSerializationTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.time.Duration;
+import java.time.ZoneId;
+
+import static org.junit.Assert.*;
```
vjkoskela: We allow specific static imports, we still (afaik) discourage `*` imports.

test/java/models/view/ReportSerializationTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Tests for [de]serializing Report-relevant types.
+
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public class ReportSerializationTest {
```
vjkoskela: nit: `final`?

test/java/models/view/ReportSerializationTest.java
```diff
@@ -0,0 +1,99 @@
...
+    @Test
+    public void testValidReport() throws IOException, URISyntaxException {
+        final Report report = OBJECT_MAPPER.readValue(loadResource("testValidReport"), Report.class);
+        assertEquals(new URI("https://example.com"), ((ChromeScreenshotReportSource) report.getSource()).getUri());
+        assertTrue(report.getSchedule() instanceof OneOffSchedule);
+        assertEquals(1, report.getRecipients().size());
+        assertEquals("nobody@example.com", report.getRecipients().get(0).getAddress());
```
vjkoskela: Should we build the entire bean and just do .equals for validation?

app/models/view/impl/HtmlReportFormat.java
```diff
@@ -49,6 +66,6 @@ public String toString() {
      * @return The view model.
      */
     public static HtmlReportFormat fromInternal(final models.internal.impl.HtmlReportFormat format) {
-        return INSTANCE;
+        return new HtmlReportFormat();
```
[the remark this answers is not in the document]
speezepearson: It seemed misleading to me to act like this class was a singleton, given that Jackson creates a new instance whenever it deserializes a `HtmlReportFormat`. (Maybe instead there's a way to tell Jackson to return a singleton instance?)

test/java/com/arpnetworking/metrics/portal/integration/controllers/ReportControllerIT.java
speezepearson: I opted to `models.view.impl.Report.fromInternal(TestBeanFactory.createReportBuilder())`. Good? Or is there a reason to have a separate view-model-creation method?

PR #203, opened 2019-05-21
Add status endpoint with actor system information.

vjkoskela:


app/global/MainModule.java
```diff
@@ -260,6 +262,18 @@ private PeriodicMetrics providePeriodicMetrics(final MetricsFactory metricsFacto
         return periodicMetrics;
     }
 
+    @Provides
+    @com.google.inject.Singleton
+    @com.google.inject.name.Named("status-cache")
+    @SuppressFBWarnings("UPM_UNCALLED_PRIVATE_METHOD") // Invoked reflectively by Guice
+    private ActorRef provideStatusCache(
```
BrandonArp: Something isn't right here.  You're purporting to supply the status-cache actor, but actually providing the status actor.

app/controllers/MetaController.java
```diff
@@ -92,6 +93,49 @@ public Result ping() {
         return internalServerError(result);
     }
 
+    /**
+     * Endpoint implementation to retrieve service status as JSON.
+     *
+     * @return Serialized response containing service status.
+     */
+    public CompletionStage<Result> status() {
```
BrandonArp: Can we just skip all this fun and inject the actorref directly?

PR #211, opened 2019-05-27
ReportFormat.Visitor: now an abstract class (instead of interface); specialize method names

speezepearson:


app/models/internal/reports/ReportFormat.java
```diff
@@ -40,30 +40,30 @@
...
 
         /**
          * Convenience method equivalent to {@code format.accept(this) }.
          *
          * @param format The format to visit.
          * @return The result of applying the visitor
          */
-        default T visit(ReportFormat format) {
+        public final T visit(final ReportFormat format) {
             return format.accept(this);
```
vjkoskela: Sorry, is there a reason the called does not call `accept` on the object and pass the visitor?

PR #212, opened 2019-05-28
add ChromeScreenshotReportSource.{timeout,jsRunOnLoad}

speezepearson:
The ability to run JS on page-load is necessary to hook our reporting system into Grafana.

Having a timeout isn't necessary, merely prudent-seeming.

### DIFF
--- a/test/java/com/arpnetworking/metrics/portal/TestBeanFactory.java
+++ b/test/java/com/arpnetworking/metrics/portal/TestBeanFactory.java
@@ -158,7 +158,7 @@ public final class TestBeanFactory {
                                 .setId(UUID.randomUUID())
                                 .setTriggeringEventName(TEST_EVENT + UUID.randomUUID().toString())
                                 .setUri(URI.create("http://" + UUID.randomUUID().toString().replace("-", "") + ".example.com"))
-                                .setTimeout(Duration.ofMillis((long)(10000 * RANDOM.nextFloat())))
+                                .setTimeout(Duration.ofMillis((long) (10000 * RANDOM.nextFloat())))
                                 .setJsRunOnLoad(jsRunOnLoad)
                                 .setIgnoreCertificateErrors(true)
                                 .build())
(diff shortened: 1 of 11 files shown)


app/models/ebean/ChromeScreenshotReportSource.java
```diff
@@ -43,6 +44,12 @@
     @Column(name = "ignore_certificate_errors")
     private boolean ignoreCertificateErrors;
 
+    @Column(name = "timeout")
+    private Duration timeout;
```
vjkoskela: I believe there is a problem with `Duration` serialization through EBean where it truncates the seconds. 

https://github.com/ArpNetworking/metrics-portal/pull/179/files#r279255056

app/models/internal/impl/ChromeScreenshotReportSource.java
```diff
@@ -204,6 +252,10 @@ public Builder setIgnoreCertificateErrors(final Boolean ignoreCertificateErrors)
         @NotEmpty
         private String _triggeringEventName;
         @NotNull
+        private Duration _timeout;
```
vjkoskela: Probably needs to be positive as well.

test/java/com/arpnetworking/metrics/portal/TestBeanFactory.java
```diff
@@ -146,6 +146,8 @@ private TestBeanFactory() {
                                 .setId(UUID.randomUUID())
                                 .setTriggeringEventName(TEST_EVENT + UUID.randomUUID().toString())
                                 .setUri(URI.create("http://" + UUID.randomUUID().toString().replace("-", "") + ".example.com"))
+                                .setTimeout(Duration.ofSeconds(123))
+                                .setJsRunOnLoad("some_javascript()")
```
vjkoskela: Neither of these is random. For the timeout, it should probably be a reasonable value in a range (e.g. 1 sec to 10 sec). For the JS, it should probably pick from some set of valid JS constants including not setting any (using the default of empty string).

speezepearson:
> This code does not seem to actually use either of these values. The timeout I can understand. What's the use case for the JS?

It's useful to be able to manipulate the page before taking a screenshot of it, in order to remove the visual clutter of the parts that don't matter -- for our Grafana use case, this means replacing the whole page-body with the rendered iframe's body.

vjkoskela:
As discussed; use `&kiosk` on the URL to avoid any post-processing.

speezepearson:
Taking a completely different approach. Abandoning this PR.

PR #218, opened 2019-05-31
add timeout to WebPageReportSource

speezepearson:
Because what if the page never finishes loading?


speezepearson:
Actually, enforcing timeouts is probably the Renderer's job.

PR #219, opened 2019-06-03
implement EmailSender

speezepearson:


test/java/com/arpnetworking/metrics/portal/reports/EmailSenderTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+import org.simplejavamail.email.Email;
+import org.simplejavamail.mailer.Mailer;
+
+import java.time.Instant;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * Tests class <code>EmailReportSink</code>.
```
cwbriones: Fix this header (classname / author)

test/java/com/arpnetworking/metrics/portal/reports/EmailSenderTest.java
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright 2018 Dropbox, Inc.
```
cwbriones: 2019

app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
```diff
@@ -32,12 +50,84 @@
...
+    private String sendSync(
+            final Report report,
+            final Recipient recipient,
+            final ImmutableMap<ReportFormat, RenderedReport> formatsToSend,
+            final Instant scheduled
+    ) throws IOException {
+
+        final String subject = getSubject(report, scheduled);
+        EmailPopulatingBuilder builder = EmailBuilder.startingBlank()
+                .from("no-reply+amp-reporting@dropbox.com")
```
cwbriones: The from field should be configurable

app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
```diff
@@ -32,12 +50,84 @@
...
+                .addData("recipient", recipient)
+                .log();
+
+        _mailer.sendMail(builder.buildEmail());
+        return "";
+    }
+
+    private String getSubject(final Report report, final Instant scheduled) {
+        final String formattedTime = ZonedDateTime.ofInstant(scheduled, ZoneId.of("UTC")).toString();
+        return "[Report] " + report.getName() + " for " + formattedTime;
```
cwbriones: I think this should also be configurable, but that's a bit more complicated than the above since we'd probably have to have it as a format string so I'd be fine on punting on it for now.

app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
```diff
@@ -32,12 +50,84 @@
...
+    }
+
+    private EmailPopulatingBuilder addFormat(
+            final EmailPopulatingBuilder builder,
+            final RenderedReport rendered
+    ) throws IOException {
+        final String mimeType = rendered.getFormat().getMimeType();
+        final InputStream content = rendered.getBytes();
+        if (mimeType.equals("text/html")) {
```
cwbriones: You should use `MediaType.HTML_UTF_8.toString()` from Guava (unless there's a standard java one I'm unaware of)

app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
```diff
@@ -32,12 +50,84 @@
...
+                .addData("report", report)
+                .addData("recipient", recipient)
+                .log();
+
+        _mailer.sendMail(builder.buildEmail());
+        return "";
+    }
+
+    private String getSubject(final Report report, final Instant scheduled) {
+        final String formattedTime = ZonedDateTime.ofInstant(scheduled, ZoneId.of("UTC")).toString();
```
cwbriones: `ZonedDateTime.ofInstant(scheduled, ZoneOffset.UTC).toString()` is equivalent

app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
```diff
@@ -32,12 +50,84 @@
...
+     *
+     * @param mailer The Mailer to send emails through.
+     */
+    @Inject
+    public EmailSender(final Mailer mailer) {
+        _mailer = mailer;
+    }
+
+    private static byte[] readBytes(final InputStream stream) throws IOException {
+        return ByteStreams.toByteArray(stream);
```
cwbriones: (this applies to `readString` as well) I'm not sure if the stream below needs to be closed but this will *not* close it https://stackoverflow.com/a/4186462

app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
```diff
@@ -32,12 +50,84 @@
...
+        }
+
+        LOGGER.info()
+                .setMessage("sending email")
+                .addData("report", report)
+                .addData("recipient", recipient)
+                .log();
+
+        _mailer.sendMail(builder.buildEmail());
+        return "";
```
cwbriones: Why does this return anything?

app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
speezepearson: Good point, and agreed re: punting! I don't see any good string-templating libraries already in the project (though I could easily have missed something), so I've left a TODO and I'm gonna call this resolved.

app/com/arpnetworking/metrics/portal/reports/RenderedReport.java
```diff
@@ -40,7 +40,7 @@
      *
      * @return The bytes.
      */
-    InputStream getBytes();
```
speezepearson: `ByteSource` is a more natural interface here than `InputStream`, I think; I just didn't know about it when I was initially writing this interface.

app/com/arpnetworking/metrics/portal/reports/Sender.java
```diff
@@ -31,12 +33,16 @@
...
+     * @param formatsToSend The reports to send. Must be non-empty.
+     * @param scheduled The instant that the report being sent was generated for.
      * @return A CompletionStage that completes when the transmission has completed.
      */
     CompletionStage<Void> send(
+            Report report,
             Recipient recipient,
-            ImmutableMap<ReportFormat, RenderedReport> formatsToSend
+            ImmutableMap<ReportFormat, RenderedReport> formatsToSend,
+            Instant scheduled
```
speezepearson: I'm a little bit uncomfortable about how I keep needing to add arguments to interface methods: it seems bad for pluggability, since people who drop in their own `Sender` implementations will need to rewrite them whenever the interface changes.

The obvious solution is to replace all of these arguments with a single omnibus object whose class is named, like, `ArgumentsToSend`. But that seems pretty gross in its own way, for reasons I can't really put my finger on.

Thoughts on this, @vjkoskela or @BrandonArp ?

app/global/MainModule.java
```diff
@@ -160,6 +162,17 @@ private Props getHostProviderProps(final HostProviderFactory provider, final Env
         return provider.create(config.getConfig("hostProvider"), ConfigurationHelper.getType(environment, config, "hostProvider.type"));
     }
 
+    @Provides
+    @Singleton
+    @SuppressFBWarnings(value = "UPM_UNCALLED_PRIVATE_METHOD", justification = "Invoked reflectively by Guice")
+    private Mailer getEmailTransport(final Config config) {
+        final String host = config.hasPath("mail.smtp.host") ? config.getString("mail.smtp.host") : "localhost";
+        final Integer port = config.hasPath("mail.smtp.port") ? config.getInt("mail.smtp.port") : 25;
```
speezepearson: `mail.smtp.{host,port}` are the properties that simplejavamail usually takes its config from, so I cribbed those names. Am open to change.

test/java/com/arpnetworking/metrics/portal/reports/EmailSenderTest.java
```diff
@@ -0,0 +1,112 @@
...
+                .build();
+        _sender.send(
+                TestBeanFactory.createReportBuilder().setName("P75 TTI").build(),
+                TestBeanFactory.createRecipient(),
+                ImmutableMap.of(report.getFormat(), report),
+                Instant.parse("2019-01-01T00:00:00.000Z")
+        ).toCompletableFuture().get();
+
+        Mockito.verify(_mailer).sendMail(_message.capture());
+        Assert.assertEquals("[Report] P75 TTI for 2019-01-01T00:00Z", _message.getValue().getSubject());
```
speezepearson: I'm open to making the time-format more human-readable, like `Tue Jan 1 2019 12:00am UTC`. Alternatively, we could punt this until we have configurable string-templating for the subject line.

app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
```diff
@@ -32,12 +50,78 @@
...
+                .setMessage("sending email")
+                .addData("report", report)
+                .addData("recipient", recipient)
+                .log();
+
+        _mailer.sendMail(builder.buildEmail());
+    }
+
+    private String getSubject(final Report report, final Instant scheduled) {
+        final String formattedTime = ZonedDateTime.ofInstant(scheduled, ZoneOffset.UTC).toString();
```
speezepearson: I feel weird about using `UTC` here. Perhaps we need to promote timezone from being a `Schedule`-level concept to being a field on `Report`? (And maybe `Report` should eventually contain a lot of locale/UI information?)

If we're playing this on hard mode, though: suppose a report is being sent to some people in SFO and some people in NYC. What should the timestamp in the subject line be _then_?

(Or maybe we should just punt on this until we have configurable/templatable subject lines.)

app/com/arpnetworking/metrics/portal/reports/Sender.java
vjkoskela: As discussed.

test/java/com/arpnetworking/metrics/portal/reports/EmailSenderTest.java
vjkoskela: I think file an issue for using intl for the header which can include formatting of the subject.

speezepearson: Done: https://github.com/ArpNetworking/metrics-portal/issues/225

PR #224, opened 2019-06-05
have Renderer populate a builder rather than returning a full RenderedReport

speezepearson:
This avoids forcing every implementation to contain the boilerplate `new DefaultRenderedReport.Builder().setFormat(format).setGeneratedAt(Instant.now()).setScheduledFor(scheduled)`.

Implemented by:
- moving `DefaultRenderedReport.Builder` to `RenderedReport.Builder`
- add `RenderedReport.Builder` argument to `Renderer#render`
- dependency-injecting a `Clock` into `ReportExecutionContext` since it now needs to be able to compute `now()`

### DIFF
--- a/app/com/arpnetworking/metrics/portal/reports/RenderedReport.java
+++ b/app/com/arpnetworking/metrics/portal/reports/RenderedReport.java
@@ -16,13 +16,10 @@
 
 package com.arpnetworking.metrics.portal.reports;
 
-import com.arpnetworking.commons.builder.OvalBuilder;
 import models.internal.reports.ReportFormat;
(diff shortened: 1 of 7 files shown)


app/com/arpnetworking/metrics/portal/reports/RenderedReport.java
```diff
@@ -63,4 +63,47 @@
      * @return The instant.
      */
     Instant getGeneratedAt();
+
+
+    /**
+     *
+     * Builder implementation that constructs {@code DefaultReport}.
```
vjkoskela: nit: this is not an "implementation"

PR #226, opened 2019-06-06
make EmailSender take lazily instantiated Mailer

speezepearson:
I'm not sure how the email branch was green in CI -- after its merge, master has build errors.

`EmailSender` needs to have a `Mailer` dependency-injected into it, or else running the tests will try to make real-life SMTP connections. I've made it a lazily instantiated Mailer in order to avoid making connections before any reports actually need sending.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
+++ b/app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
@@ -116,12 +116,23 @@ public class EmailSender implements Sender {
      */
     @Inject
     public EmailSender(@Assisted final Config config) {
+        this(
+                MailerBuilder
+                        .withSMTPServer(
+                                config.hasPath("smtp.host") ? config.getString("smtp.host") : "localhost",
+                                config.hasPath("smtp.port") ? config.getInt("smtp.port") : 25
+                        )
+                        .buildMailer(),
+                config
+        );
+    }
+
+    /**
+     * Constructor for tests, allowing dependency injection of the {@link Mailer}.
+     */
+    /* package private */ EmailSender(final Mailer mailer, final Config config) {
         _fromAddress = config.getString("fromAddress");
-        final String host = config.hasPath("smtp.host") ? config.getString("smtp.host") : "localhost";
-        final Integer port = config.hasPath("smtp.port") ? config.getInt("smtp.port") : 25;
-        _mailer = MailerBuilder
-                .withSMTPServer(host, port)
-                .buildMailer();
+        _mailer = mailer;
     }
 
     private final Mailer _mailer;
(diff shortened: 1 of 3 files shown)


app/global/MainModule.java
```diff
@@ -160,6 +164,20 @@ private Props getHostProviderProps(final HostProviderFactory provider, final Env
         return provider.create(config.getConfig("hostProvider"), ConfigurationHelper.getType(environment, config, "hostProvider.type"));
     }
 
+    @Singleton
+    @Named("mailer")
```
speezepearson: Is this `@Named` necessary? I'm not confident in how Guice injects generics like `Supplier<Mailer>`.

cwbriones: 1. Unless you have multiple bindings,`@Named` is not necessary.

2. This has a level of indirection that's unnecessary. If you instead mark this as `@Provides Mailer` and then at the injection site request a `com.google.inject.Provider<Mailer>`, guice will automatically create a Provider instance for you and it will always refer to the same object because of the `@Singleton` scope, removing the need for memoization.

i.e.
```java
    @Provides
    @SuppressFBWarnings(value = "UPM_UNCALLED_PRIVATE_METHOD", justification = "Invoked reflectively by Guice")
    private Mailer getMailer(final Config config) {
            final String host = config.hasPath("mail.smtp.host") ? config.getString("mail.smtp.host") : "localhost";
            final Integer port = config.hasPath("mail.smtp.port") ? config.getInt("mail.smtp.port") : 25;
            return MailerBuilder
                    .withSMTPServer(host, port)
                    .buildMailer();
    }
```

cwbriones: re: 1. It looks like this is only true if you use the method in 2. If not, you would have to create the binding with a `TypeLiteral` to preserve the generic parameter.

BrandonArp: Also, you shouldn't even have to inject a Provider<Mailer>.  As long as you're not binding as an eager singleton, it should be delayed until you actually construct something that needs the Mailer.

vjkoskela: I don't think the Mailer should be injected at all. Has something changed from our discussion on this yesterday?

speezepearson: Re: eagerness vs laziness: since Senders are eagerly instantiated, if the EmailSender takes a Mailer then the Mailer must be eagerly instantiated too. (One could argue that Senders should be lazily instantiated, which wouldn't be crazy; I opted for eagerness because if the config is malformed, it seems more appropriate to throw the exception on startup.)

> I don't think the Mailer should be injected at all.

For testing, it _must_ be possible to mock out an EmailSender's Mailer; else, tests will try to make real-life SMTP connections. I'm guessing you favor a `public EmailSender(Config)` constructor which builds a Mailer and delegates to a `/*package-private*/ EmailSender(Config, Mailer)`?

PR #233, opened 2019-06-14
make rendering use TimeRange instead of the time the report was scheduled to run

speezepearson:


app/models/internal/TimeRange.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.time.Instant;
+import java.util.Objects;
+
+/**
+ * Represents an interval of time.
+ *
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public class TimeRange {
```
vjkoskela: `@Loggable`?
`final`?

PR #234, opened 2019-06-20
Added period based chunking, fixed period evaluation bug and added count aggregator to rollup query.

ddimensia:
This addresses a few issues with rollups:

1. Backfills could be expensive because multiple periods would try to be rolled up in a single saveas query.  This was addressed by chunking the saveAs queries into one query per period.

2. When the saveAs was moved to save data at the startTime there was a mismatch with how the calculations were made to determine if a rollup needed to occur.  Previously we checked against the end of the most recently closed period, now we check against the start of the most recently closed period.

3. The saveAs query could return a sizable payload for histograms.  This is because saveAs is done through the normal query mechanism and simply passes the datapoints that were saved back to the caller.  In the case of histograms this would be the entire serialized histogram.  By adding a 'count' aggregator after the saveAs we now get a single datapoint per tag combination.

### DIFF
--- a/app/com/arpnetworking/rollups/RollupGenerator.java
+++ b/app/com/arpnetworking/rollups/RollupGenerator.java
@@ -73,11 +73,11 @@ public class RollupGenerator extends AbstractActorWithTimers {
     /**
      * RollupGenerator actor constructor.
      *
-     * @param configuration    play configuration
+     * @param configuration play configuration
      * @param metricsDiscovery actor ref to metrics discovery actor
-     * @param kairosDbClient   kairosdb client
-     * @param clock            clock to use for time calculations
-     * @param metrics          periodic metrics instance
+     * @param kairosDbClient kairosdb client
+     * @param clock clock to use for time calculations
+     * @param metrics periodic metrics instance
      */
     @Inject
     public RollupGenerator(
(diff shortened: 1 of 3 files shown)


app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -70,11 +74,11 @@ public Receive createReceive() {
     /**
      * RollupGenerator actor constructor.
      *
-     * @param configuration play configuration
+     * @param configuration    play configuration
```
vjkoskela: nit: pretty sure we don't do column alignment on parameter descriptions.

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -297,45 +300,62 @@ private LastDataPointMessage buildLastDataPointResponse(
...
+            rollupPeriodStart = rollupPeriodStart.plus(period.periodCountToDuration(1));
+        }
 
         metricBuilder.setName(message.getMetricName());
         if (!message.getTags().isEmpty()) {
             metricBuilder.setGroupBy(ImmutableList.of(
                     new MetricsQuery.GroupBy.Builder()
-                    .setName("tag").addOtherArg("tags", message.getTags())
-                    .build()
+                            .setName("tag").addOtherArg("tags", message.getTags())
```
vjkoskela: nit: can you move `.addOtherArg` to a new line

app/com/arpnetworking/rollups/RollupGenerator.java
ddimensia: Fixed

app/com/arpnetworking/rollups/RollupGenerator.java
ddimensia: Fixed

PR #235, opened 2019-06-26
add missing email-sending dependency

speezepearson:
Apparently `simplejavamail` has a sneaky dependency on `net.markenwerk.utils.mail.dkim`. I pulled it in, and added a mock SMTP server library for testing, so that messages actually get sent through `simplejavamail` instead of it being mocked out the whole time. (Maybe this should be an integration test? It seems conceptually very similar to using WireMock, so I made it a unit test.)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -279,7 +279,6 @@
         <configuration>
           <ignoredDependencies combine.children="append">
             <ignoredDependency>com.typesafe.play:play-java_${scala.package.version}:jar:*</ignoredDependency>
-            <ignoredDependency>net.markenwerk:utils-mail-dkim:jar:1.1.12</ignoredDependency>
           </ignoredDependencies>
         </configuration>
       </plugin>
@@ -748,6 +747,7 @@
       <groupId>net.markenwerk</groupId>
       <artifactId>utils-mail-dkim</artifactId>
       <version>${dkimutils.version}</version>
+      <scope>runtime</scope>
     </dependency>
 
 
--- a/pom.xml
+++ b/pom.xml
@@ -277,6 +277,7 @@
         <configuration>
           <ignoredDependencies combine.children="append">
             <ignoredDependency>com.typesafe.play:play-java_${scala.package.version}:jar:*</ignoredDependency>
+            <ignoredDependency>net.markenwerk:utils-mail-dkim:jar:1.1.12</ignoredDependency>
           </ignoredDependencies>
         </configuration>
       </plugin>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -741,6 +742,12 @@
       <artifactId>simple-java-mail</artifactId>
       <version>5.0.0</version>
     </dependency>
+    <dependency>
+      <groupId>net.markenwerk</groupId>
+      <artifactId>utils-mail-dkim</artifactId>
+      <version>1.1.12</version>
```
vjkoskela: The versions should be in properties at the top sorted alphabetically and into non-test/test. Same for above for `simple-java-mail` and below for `dumbster`.

pom.xml
```diff
@@ -741,6 +742,12 @@
       <artifactId>simple-java-mail</artifactId>
       <version>5.0.0</version>
```
vjkoskela: Is there a reason we used a 16 month old version of `simple-java-mail`? It looks like `5.1.7` was released in May-19.

https://search.maven.org/search?q=g:org.simplejavamail%20AND%20a:simple-java-mail&core=gav

vjkoskela: (there could be a good reason; just curious)

pom.xml
```diff
@@ -277,6 +277,7 @@
         <configuration>
           <ignoredDependencies combine.children="append">
             <ignoredDependency>com.typesafe.play:play-java_${scala.package.version}:jar:*</ignoredDependency>
+            <ignoredDependency>net.markenwerk:utils-mail-dkim:jar:1.1.12</ignoredDependency>
```
vjkoskela: So I suspect the reason you are doing this is because there are no non-test references to this package. Why not simply make the direct dependency on this package scoped to `test` (and remove the exception here)?

pom.xml
vjkoskela: Why is this not a `test` scope dependency?

pom.xml
```diff
@@ -1136,6 +1143,12 @@
       <version>${thrift.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.kirviq</groupId>
+      <artifactId>dumbster</artifactId>
+      <version>1.7.1</version>
```
vjkoskela: Version property reference.

pom.xml
speezepearson: No good reason! Just copy-pasting.

PR #236, opened 2019-06-27
Fixed issue with rollup override (_!) not returning tags.

ddimensia:
In grafana, if you tried to use the rollup override suffix you'd get no tags back as it wasn't a valid metric name.  This modifies the proxy to remove the _! suffix if it sees it on a queryMetricTags request.

### DIFF
--- a/app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
+++ b/app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
@@ -57,8 +57,7 @@ public class KairosDbServiceImpl implements KairosDbService {
 
 
     /**
-     * Public constructor.
-     *
+     * Public constructor. *
      * @param kairosDbClient Client to use to make requests to backend kairosdb
      * @param metricsFactory MetricsFactory instance for recording service metrics
      */
@@ -69,7 +68,16 @@ public class KairosDbServiceImpl implements KairosDbService {
 
     @Override
     public CompletionStage<MetricsQueryResponse> queryMetricTags(final MetricsQuery query) {
-        throw new UnsupportedOperationException();
+        final Metrics metrics = _metricsFactory.create();
+        final Timer timer = metrics.createTimer("kairosService/queryMetricTags/request");
+        // Filter out rollup metric overrides and forward the query
+        return filterRollupOverrides(query)
+                .thenCompose(_kairosDbClient::queryMetricTags)
+                .whenComplete((result, error) -> {
+                    timer.stop();
+                    metrics.incrementCounter("kairosService/queryMetricTags/success", error == null ? 1 : 0);
+                    metrics.close();
+                });
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
```diff
@@ -57,8 +57,7 @@
 
 
     /**
-     * Public constructor.
-     *
+     * Public constructor. *
```
vjkoskela: nit: the star (`*`) goes on the next line ;)

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
```diff
@@ -69,7 +68,16 @@ public KairosDbServiceImpl(final KairosDbClient kairosDbClient, final MetricsFac
...
     public CompletionStage<MetricsQueryResponse> queryMetricTags(final MetricsQuery query) {
-        throw new UnsupportedOperationException();
+        final Metrics metrics = _metricsFactory.create();
+        final Timer timer = metrics.createTimer("kairosService/queryMetricTags/request");
+        // Filter out rollup metric overrides and forward the query
+        return filterRollupOverrides(query)
+                .thenCompose(_kairosDbClient::queryMetricTags)
+                .whenComplete((result, error) -> {
+                    timer.stop();
+                    metrics.incrementCounter("kairosService/queryMetricTags/success", error == null ? 1 : 0);
```
vjkoskela: I think we usually (or try?) to use snake case in metric names. 

(looks like in this file we haven't -- not a big deal esp if we have dashboards already)

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
```diff
@@ -210,7 +218,7 @@ private static MetricsQuery useAvailableRollups(
         newQueryBuilder.setMetrics(originalQuery.getMetrics().stream().map(metric -> {
             // Check to see if there are any rollups for this metrics
             final String metricName = metric.getName();
-            if (metricName.endsWith("_!")) {
+            if (metricName.endsWith(ROLLUP_OVERRIDE)) {
                 metrics.incrementCounter("kairosService/useRollups/bypass", 1);
```
vjkoskela: Should we log a `0` against `bypass` in the `else` branch?

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
ddimensia: Fixed

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
ddimensia: Fixed

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
ddimensia: We already have dashboards so not changing at this point.

PR #237, opened 2019-06-29
mark DefaultRecipient and DefaultReportResult loggable

speezepearson:
JSON log output from Docker:
Before:
```json
    "recipient": {
      "_id": "7fb58d1f",
      "_class": "models.internal.impl.DefaultRecipient"
    }
```
After:
```
    "recipient": {
      "_id": "4e3247b",
      "_class": "models.internal.impl.DefaultRecipient",
      "address": "a@b.com",
      "id": "1516321a-2e59-4e8a-8da5-71c8bb504580",
      "type": "EMAIL"
    }
```

### DIFF
--- a/app/com/arpnetworking/metrics/portal/reports/RecipientType.java
+++ b/app/com/arpnetworking/metrics/portal/reports/RecipientType.java
@@ -15,14 +15,11 @@
  */
 package com.arpnetworking.metrics.portal.reports;
 
-import com.arpnetworking.logback.annotations.Loggable;
-
 /**
  * Mechanisms by which reports can be sent to people.
  *
  * @author Spencer Pearson (spencerpearson at dropbox dot com)
  */
-@Loggable
 public enum RecipientType {
     /**
      * Email.
--- a/app/com/arpnetworking/metrics/portal/reports/RecipientType.java
+++ b/app/com/arpnetworking/metrics/portal/reports/RecipientType.java
@@ -15,11 +15,14 @@
  */
 package com.arpnetworking.metrics.portal.reports;
 
+import com.arpnetworking.logback.annotations.Loggable;
+
 /**
  * Mechanisms by which reports can be sent to people.
  *
  * @author Spencer Pearson (spencerpearson at dropbox dot com)
  */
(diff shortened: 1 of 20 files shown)


app/com/arpnetworking/metrics/portal/reports/RecipientType.java
```diff
@@ -15,11 +15,14 @@
...
 package com.arpnetworking.metrics.portal.reports;
 
+import com.arpnetworking.logback.annotations.Loggable;
+
 /**
  * Mechanisms by which reports can be sent to people.
  *
  * @author Spencer Pearson (spencerpearson at dropbox dot com)
  */
+@Loggable
```
vjkoskela: Can you confirm this actually does something? -- I'd love to see the before/after output for this. My suspicion is that it may be a no-op and that this is a bug in the logger factory.

app/models/internal/reports/Recipient.java
```diff
@@ -25,6 +26,7 @@
  *
  * @author Christian Briones (cbriones at dropbox dot com)
  */
+@Loggable
```
vjkoskela: Typically, we put the annotation on the implementation because it will serialize the implementation instance not the interface. I don't think this is wrong, in fact it's probably more right because it _should_ constrain the log output to just the methods on the interface. 

If it's not too much trouble, could you post output of logging an implementation instance of `Recipient` where the implementation has a public getter on it (e.g. `public String getFoo() { return "foo"; }`)?

app/com/arpnetworking/metrics/portal/reports/RecipientType.java
speezepearson: I only added this because I was overeager! Tested by hand -- turns out that enums serialize just fine into strings, as you can see up above.

app/models/internal/reports/Recipient.java
speezepearson: Rather than opening this possibly-time-consuming can of worms, I just put the annotation on `DefaultRecipient`. I can open an issue for this if you want?

PR #241, opened 2019-07-03
make UI capable of handling Grafana reports

speezepearson:


app/assets/javascripts/classes/reports/EditReportViewModel.ts
```diff
@@ -197,6 +217,8 @@ class EditSourceViewModel extends BaseSourceViewModel {
               "<ul class='list-group'>"+
               "<li class='list-group-item'><b>Browser rendered</b> - A URL is loaded in the browser and the rendered " +
               "contents of the page are taken as the generated report.</li>" +
+              "<li class='list-group-item'><b>Grafana</b> - Like browser-rendered, but tweaked to pull data from a Grafana Report panel" +
+              " (which are loaded asynchronously after page-load, hence needing the specialization).</li>" +
```
cwbriones: I'm starting to think a better distinction would be to have an empty event field for Browser Rendered, leaving "Grafana" exactly the same but with an event to wait for specified.

app/assets/javascripts/classes/reports/EditReportViewModel.ts
```diff
@@ -175,14 +175,34 @@ class EditRecipientViewModel extends BaseRecipientViewModel {
...
+                    ...requiredFields,
+                    uri: this.url(),
+                    title: this.title(),
+                    ignoreCertificateErrors: this.ignoreCertificateErrors(),
+                    triggeringEventName: this.eventName(),
+                };
+            case SourceType.GRAFANA:
+                return {
+                    ...requiredFields,
+                    webPageReportSource: {
```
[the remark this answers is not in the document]
speezepearson: Yeah, somewhat. Throughout the stack (except at the EBean level), a GrafanaReportSource just contains a WebPageReportSource. (I went back and forth on that -- happy to discuss my reasoning in person / over Slack.) The nested JSON structure both mirrors that structure and is somewhat easier to parse as a result.

app/assets/javascripts/classes/reports/EditReportViewModel.ts
speezepearson: (discussed in person; I believe we agree that these need to be separate, at least for now)

PR #244, opened 2019-07-08
make Chrome-interaction code use CompletionStages rather than callback registration

speezepearson:


app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsServiceWrapper.java
```diff
@@ -66,18 +69,26 @@ public Object evaluate(final String js) {
...
     }
 
     @Override
-    public void onLoad(final Runnable callback) {
-        _dts.getPage().enable();
-        _dts.getPage().onLoadEventFired(e -> callback.run());
+    public CompletionStage<Void> nowOrOnEvent(final String eventName, final Supplier<Boolean> ready) {
+        final CompletableFuture<Void> result = new CompletableFuture<>();
+        waitForEvent(eventName).thenAccept(foo -> result.complete(null));
+        if (ready.get()) {
```
vjkoskela: I'm confused; wont this block?

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get--

PR #249, opened 2019-07-10
log errors from job execution

speezepearson:


speezepearson:
Oh, we already do this, just at the `debug` level. That's why I didn't see it.

PR #258, opened 2019-07-23
add JsonAny{Getter,Setter} to Metric for arbitrary forwards-compatibility

speezepearson:


app/com/arpnetworking/kairos/client/models/Metric.java
```diff
@@ -177,6 +202,9 @@ public Builder setOrder(@Nullable final Order value) {
         private Integer _limit;
 
         private Order _order;
+
+        @NotNull
+        private Map<String, Object> _extraFields = Maps.newHashMap();
```
vjkoskela: If you want to avoid the copy above, you could use an `ImmutableMap.Builder`. When used with JSON this is safe, but if used by hand it does not allow overwriting keys (and instead throws an exception).

speezepearson: Thanks for the tip! Since this isn't a hot path, I think I'll leave it as-is out of laziness.

PR #269, opened 2019-07-24
make DevToolsService#close() idempotent

speezepearson:


app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsServiceWrapper.java
```diff
@@ -131,6 +133,9 @@ public Object evaluate(final String js) {
 
     @Override
     public void close() {
+        if (_closed.getAndSet(true)) {
+            return;
```
vjkoskela: Description states any other calls are illegal after close, but does not behave any differently. Should we throw ISE if closed on any method call?

PR #270, opened 2019-07-24
change *RendererIT to *RendererTest; and refactor out common time range

speezepearson:


test/java/com/arpnetworking/metrics/portal/reports/impl/chrome/BaseChromeTest.java
```diff
@@ -40,11 +42,12 @@
  *
  * @author Spencer Pearson (spencerpearson at dropbox dot com)
  */
-public class BaseChromeIT {
+public class BaseChromeTest {
```
vjkoskela: Should not end with `IT` or `Test` if this class does not actually include tests. Further, it should probably be marked `abstract` if it is intended for extension/code sharing only.

PR #272, opened 2019-07-30
Use an independent actor for holding rollups to be performed.

ddimensia:
Adds a cluster sharded actor that holds rollups to be performed, with an actor for actually executing those rollups.

Currently the actors are not all fully configured, this will also be followed with a PR to remove the rollup execution code out of the RollupGenerator actor.  That actor will instead send RollupDefinitions to the new RollupManager actor.

### DIFF
--- a/app/com/arpnetworking/rollups/RollupDefinition.java
+++ b/app/com/arpnetworking/rollups/RollupDefinition.java
@@ -1,0 +1,177 @@
+/*
+ * Copyright 2019 Dropbox Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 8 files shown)


app/com/arpnetworking/rollups/RollupDefinition.java
```diff
@@ -0,0 +1,177 @@
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(_sourceMetricName, _destinationMetricName, _period, _startTime, _endTime, _groupByTags);
+    }
+
+    /**
+     * {@code RollupDefinition} builder static inner class.
+     */
+    public static final class Builder extends OvalBuilder<RollupDefinition> {
+        private String _sourceMetricName;
```
vjkoskela: Are there no validation constraints on these?

PR #273, opened 2019-07-30
Make Chrome devtools actions cancellable

speezepearson:
This diff is probably best viewed commit-by-commit.

PR #274, opened 2019-07-31
make Chrome-based renderers respect report timeouts

speezepearson:
This diff, on its own, doesn't actually free up the resources being consumed, it just marks a CompletableFuture as cancelled. #273 will make that cancellation cascade into freeing up resources.


app/com/arpnetworking/metrics/portal/reports/impl/chrome/BaseScreenshotRenderer.java
```diff
@@ -98,19 +102,24 @@
...
+                });
+
+        result.whenComplete((x, e) -> {
+            LOGGER.debug()
+                    .setMessage("rendering completed")
+                    .addData("source", source)
+                    .addData("format", format)
+                    .addData("timeRange", timeRange)
+                    .addData("result", x)
+                    .addData("exception", e)
```
vjkoskela: You should add the exception with `.setThrowable`.

PR #277, opened 2019-08-01
Wired in RollupExecutor and made RollupGenerator delegate to RollupManager.

ddimensia:
This wires up the RollupExecutor actor in the MainModule and changes RollupGenerator to send RollupDefinitions to RollupManager rather than execute the rollup queries itself.


vjkoskela:
Looks good, no way to really test it by reading the code. Only question is whether we really want roll-ups on by default. Generally, features in MP are opt-in.

PR #278, opened 2019-09-09
add OriginConfig class

speezepearson:
An instance of this class defines policies to be implemented for a particular origin, e.g. what paths should be requestable/navigable-to or what headers should be added to requests to that origin.

Potentially controversial: I've added Jackson annotations to this class so that it's deserializable from JSON. I'm hoping to use Jackson to deserialize the `Map<String, OriginConfig>` in the application.conf file, because manually validating fields' presence/types is a pain. I'd be willing-but-not-thrilled to budge on this.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/reports/impl/chrome/OriginConfig.java
+++ b/app/com/arpnetworking/metrics/portal/reports/impl/chrome/OriginConfig.java
@@ -17,8 +17,6 @@
 package com.arpnetworking.metrics.portal.reports.impl.chrome;
 
 import com.arpnetworking.commons.builder.OvalBuilder;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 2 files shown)


app/com/arpnetworking/metrics/portal/reports/impl/chrome/OriginConfig.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.Objects;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+/**
+ * Describes special actions that should be done for some origin (i.e. scheme+host+port).
+ *
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+@JsonDeserialize(builder = OriginConfig.Builder.class)
```
vjkoskela: This should not be necessary if you are serializing/deserializing via the injected ObjectMapper.

app/com/arpnetworking/metrics/portal/reports/impl/chrome/OriginConfig.java
```diff
@@ -0,0 +1,158 @@
...
+                .add("allowedNavigationPaths", _allowedNavigationPaths)
+                .add("allowedRequestPaths", _allowedRequestPaths)
+                .add("additionalHeaders", _additionalHeaders)
+                .toString();
+    }
+
+    /**
+     * Builder implementation that constructs {@code OriginConfig}.
+     */
+    @JsonPOJOBuilder(withPrefix = "set")
```
vjkoskela: Same here.

PR #279, opened 2019-09-10
Enforce origin restrictions in DevToolsService (stacked on #278)

speezepearson:


app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsServiceWrapper.java
```diff
@@ -63,6 +68,39 @@
...
+    }
+
+    private void configureRequestInterception() {
+        final Network network = _dts.getNetwork();
+        network.setRequestInterception(ImmutableList.of(new RequestPattern()));
+        network.onRequestIntercepted(event -> {
+            final URI uri = URI.create(event.getRequest().getUrl());
+            final String origin = uri.getScheme() + "://" + uri.getAuthority();
+            final OriginConfig originConfig = _originConfigs.get(origin);
+            if (originConfig == null || !originConfig.isRequestAllowed(uri.getPath())) {
```
cwbriones: nit. I think all of this logic could be replaced with `isNavigationAllowed`

PR #280, opened 2019-09-10
plumb OriginConfigs from configfile into devtools

speezepearson:


app/com/arpnetworking/play/configuration/ConfigurationHelper.java
```diff
@@ -111,6 +113,17 @@ public static FiniteDuration getFiniteDuration(final Config configuration, final
...
 
+    /**
+     * Resolves and converts a piece of a {@link Config} to a JSON string.
+     *
+     * @param config The config to be rendered as JSON.
+     * @param path The path to the sub-object to render as JSON.
+     * @return The JSON representation of that sub-object.
+     */
+    public static String toJson(final Config config, final String path) {
+        return config.withFallback(ConfigFactory.systemProperties()).resolve().getObject(path).render(ConfigRenderOptions.concise());
```
BrandonArp: I'm concerned about your fallback here.  All of the fallbacks should already be applied when the config was initialized.

speezepearson: Ah, thanks! I can fix that -- I just cargo-culted wrong from MAD.

app/com/arpnetworking/metrics/portal/reports/impl/chrome/DefaultDevToolsFactory.java
```diff
@@ -81,6 +85,14 @@ public DefaultDevToolsFactory(final Config config) {
         _chromeArgs = ImmutableMap.copyOf(config.getObject("args").unwrapped());
         _executor = createExecutorService(config.hasPath("executor") ? config.getConfig("executor") : ConfigFactory.empty());
         _service = Suppliers.memoize(this::createService);
+        try {
+            _originConfigs = ObjectMapperFactory.createInstance().readValue(
```
BrandonArp: Missed this before: inject an ObjectMapper here rather than creating one.

app/com/arpnetworking/play/configuration/ConfigurationHelper.java
```diff
@@ -111,6 +113,17 @@ public static FiniteDuration getFiniteDuration(final Config configuration, final
...
 
+    /**
+     * Resolves and converts a piece of a {@link Config} to a JSON string.
+     *
+     * @param config The config to be rendered as JSON.
+     * @param path The path to the sub-object to render as JSON.
+     * @return The JSON representation of that sub-object.
+     */
+    public static String toJson(final Config config, final String path) {
+        return config.resolve().getObject(path).render(ConfigRenderOptions.concise());
```
BrandonArp: I don't think you need to .resolve() anymore.  It should've already been done for you.

PR #285, opened 2019-09-11
allow (require) whitelisting of legal report email recipients

speezepearson:


app/com/arpnetworking/metrics/portal/reports/impl/EmailSender.java
```diff
@@ -113,18 +121,24 @@ private EmailPopulatingBuilder addFormat(
...
 
     /**
      * Constructor for tests, allowing dependency injection of the {@link Mailer}.
      */
-    /* package private */ EmailSender(final Mailer mailer, final Config config) {
+    /* package private */ EmailSender(final Mailer mailer, final Config config, final ObjectMapper mapper) {
         _fromAddress = config.getString("fromAddress");
         _mailer = mailer;
+        try {
+            _allowedRecipients = mapper.readValue(ConfigurationHelper.toJson(config, "allowedRecipients"), ALLOWED_RECIPIENTS_TYPE);
```
cwbriones: nit. make this a top-level constant

test/java/com/arpnetworking/metrics/portal/reports/impl/EmailSenderTest.java
```diff
@@ -100,17 +107,35 @@ public void testSendFailsIfExceptionThrown() throws MailException, ExecutionExce
...
         try {
             _sender.send(
-                    TestBeanFactory.createRecipient(),
+                    TestBeanFactory.createRecipientBuilder().build(),
                     ImmutableMap.of(new HtmlReportFormat.Builder().build(), report)
             ).toCompletableFuture().get();
         } catch (final ExecutionException e) {
-            if (e.getCause() instanceof MailException) {
-                throw (MailException) e.getCause();
+            final Throwable cause = e.getCause();
```
cwbriones: I think this would be clearer with a JUnit rule and an `thrown.expectCause(isInstanceOf(MailException.class))` call

see https://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html#expectCause(org.hamcrest.Matcher)

test/java/com/arpnetworking/metrics/portal/reports/impl/EmailSenderTest.java
```diff
@@ -100,17 +107,35 @@ public void testSendFailsIfExceptionThrown() throws MailException, ExecutionExce
...
     }
 
+    @Test
+    public void testSendRejectsIllegalRecipients() {
+        final ImmutableList<String> legalAddresses = ImmutableList.of("alice@example.com", "bob@example.com");
+        for (final String address : legalAddresses) {
+            _sender.send(TestBeanFactory.createRecipientBuilder().setAddress(address).build(), ImmutableMap.of());
+        }
+
+        final ImmutableList<String> illegalAddresses = ImmutableList.of("", "@example.com", "charlie@example.com.ru");
```
cwbriones: nit. Can you split this into its own test case? the top portion is more `testSendAllowsWhitelistedRecipients`

It _may_ also be useful to use JUnit rules here as well so that you can see all recipients that failed
https://junit.org/junit4/javadoc/latest/org/junit/rules/ErrorCollector.html

PR #291, opened 2019-09-14
generalize QueryProblem -> Problem

speezepearson:
...and refactor some associated methods out of MetricsController, so that they can eventually get reused for Reports.

### DIFF
--- a/app/com/arpnetworking/play/metrics/ProblemHelper.java
+++ b/app/com/arpnetworking/play/metrics/ProblemHelper.java
@@ -26,7 +26,7 @@ import play.mvc.Http;
 import java.util.List;
 
 /**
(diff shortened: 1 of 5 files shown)


app/com/arpnetworking/play/metrics/ProblemHelper.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.common.collect.ImmutableList;
+import models.internal.Problem;
+import play.Environment;
+import play.libs.Json;
+import play.mvc.Http;
+
+import java.util.List;
+
+/**
+ * TODO(spencerpearson).
```
vjkoskela: Please complete.

PR #292, opened 2019-09-16
internationalize reporting error messages

speezepearson:
- tweak the report-editing UI's alert box so that it can display multiple messages (e.g. for reports that have multiple validation errors).
- change several validation methods from `void ... throws ...` to `ImmutableList<Problem> ...`

Net effect:
Screenshot 2019-09-16 14 15 37


app/assets/javascripts/classes/reports/EditReportViewModel.ts
```diff
@@ -77,7 +77,7 @@ class EditReportViewModel {
                 this.recipients.push(model);
             });
         }).fail(() => {
-            this.alertMessage("Report not found");
+            this.alertMessages(["Report not found"]);
```
vjkoskela: Ideally on a 404 for the report we would return the error information from the server instead of hard coding the language specific text in the client.

speezepearson:
> a comment inline
✅ 
>  and an issue
✅

PR #294, opened 2019-09-17
make default configuration very permissive

speezepearson:
It will now send email to anyone, and allow navigation/requests to any web page.

### DIFF
--- a/test/java/com/arpnetworking/metrics/portal/reports/impl/chrome/PerOriginConfigsTest.java
+++ b/test/java/com/arpnetworking/metrics/portal/reports/impl/chrome/PerOriginConfigsTest.java
@@ -52,12 +52,20 @@ public class PerOriginConfigsTest {
         );
     }
 
+
     @Test
     public void testDefaults() {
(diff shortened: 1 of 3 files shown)


test/java/com/arpnetworking/metrics/portal/reports/impl/chrome/PerOriginConfigsTest.java
```diff
@@ -78,31 +86,14 @@ public void testDispatch() {
         assertFalse(config.isNavigationAllowed("https://other.com/"));
     }
 
-    @Test
-    public void testIsNavigationAllowed() {
```
cwbriones: isn't this test still useful?

speezepearson: It is! But it's also an accidental copy-paste of [a test for a related component](https://github.com/speezepearson/metrics-portal/blob/permissive/test/java/com/arpnetworking/metrics/portal/reports/impl/chrome/OriginConfigTest.java#L85).

PR #302, opened 2019-09-24
fix fields not getting proxied to Kairos query endpoint

### LINKED ISSUE #257: Put @JsonAnyGetter and @JsonAnySetter on all KairosDB JSON models for proxy's forwards-compatibility
When we added per-metric `{start,end}_absolute` fields to Kairos, we didn't update MPortal's proxy's data-model to match, so those fields are getting dropped.

(How will this interact with the Builder-Jackson integration? Dunno.)

speezepearson:
Also, add de/serialization tests for various model classes (and add equals()/hashCode()/toString() in service thereof)

Closes #257.


app/com/arpnetworking/kairos/client/models/Aggregator.java
```diff
@@ -78,6 +80,40 @@ public String getName() {
         return _otherArgs;
     }
 
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                .add("_name", _name)
```
vjkoskela: nit: (it's gotten inconsistent but) preference is to use field names with the underscore prefix for the name.

test/java/com/arpnetworking/kairos/client/models/AggregatorTest.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.IOException;
+
+import static org.junit.Assert.fail;
+
+/**
+ * Tests for {@link Aggregator}.
+ *
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public class AggregatorTest {
```
vjkoskela: nit: `final`

test/java/com/arpnetworking/kairos/client/models/testing/SerializationTestUtils.java
```diff
@@ -0,0 +1,84 @@
...
+        final Object complete = OBJECT_MAPPER.convertValue(node, clazz);
+        for (final String field : ImmutableList.copyOf(node.fieldNames())) {
+            try {
+                final Object diminished = OBJECT_MAPPER.convertValue(node.deepCopy().without(field), clazz);
+                Assert.assertNotEquals(
+                        String.format("removing field %s did not change deserialized representation", field),
+                        complete,
+                        diminished
+                );
+            } catch (final IllegalArgumentException e) {
```
vjkoskela: Shouldn't this be converted to a `RuntimeException` and rethrown?

test/java/com/arpnetworking/kairos/client/models/SamplingTest.java
```diff
@@ -0,0 +1,56 @@
...
+    private static final String CLASS_NAME = SamplingTest.class.getSimpleName();
+
+    @Test
+    public void testTranslationLosesNothing() throws Exception {
+        final ObjectNode node = OBJECT_MAPPER.readValue(readResource("testTranslationLosesNothing"), ObjectNode.class);
+        SerializationTestUtils.assertTranslationLosesNothing(node, Sampling.class);
+    }
+
+
+    private String readResource(final String resourceSuffix) {
```
vjkoskela: This method seems the same for all these classes. Instead of passing in a `CLASS_NAME` you could just pass in `Class<?>` and then obtain its package path and name to create the file path.

e.g.
```
getClass().getResource(this.getClass().getSimpleName() + ".json");
```

test/java/com/arpnetworking/kairos/client/models/testing/SerializationTestUtils.java
speezepearson: Nope! `convertValue` throws `IllegalArgumentException` if conversion fails, e.g. if the Builder fails to validate, e.g. if we've removed a required field.

test/java/com/arpnetworking/kairos/client/models/SamplingTest.java
speezepearson: Good point! This led to a significant overhaul where (a) I noticed that I'd already written a helper for this, and (b) I saw one of Gil's tests using `JSONAssert` and realized that that library could greatly simplify the `SerializationTestUtils` class.

PR #304, opened 2019-09-30
Tag Namer Filtering Part 1

vjkoskela:
Add support for excluding tags (by name). Implement first part of this filtering for the list tag names endpoint (http://[host]:[port]/api/v1/tagnames).


vjkoskela:
There are many possible extensions to supporting tag filtering, such as regex based filters, per metric filters, white and black filters (e.g. drop all these, except these). 

The immediate use case for this is to support architectures where we explicitly add tags in order to support multiple re-aggregation clusters. In such a case we need to remove the tag globally. 

Thanks for the input!

PR #305, opened 2019-10-01
upgrade to using Chrome devtools Fetch API instead of Network

speezepearson:
This will make reporting work with newer versions of Chrome. (Thank goodness, the Chrome devtools library recently had a version bump that included the new devtools API.)

Verified locally with `mvn verify` and by creating a tab in non-headless mode and manually driving it to sites in order to verify that the rules specified in the application.conf were enforced. Hopefully it passes on Travis too.


app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsServiceWrapper.java
```diff
@@ -113,23 +140,26 @@ private void configureRequestInterception() {
...
-                0.4,
-                "",
-                true,
-                "",
-                "",
-                true
-        )));
+        return supplyInExecutor(() -> Base64.getDecoder().decode(
+                _dts.getPage().printToPDF(
+                    false,
```
vjkoskela: nit: It would be helpful to add a comment after each literal argument as to the parameter name to make this more readable.

PR #307, opened 2019-10-09
Added RollupForwarder to use as an indirection level for between the cluster pool and RollupManager

ddimensia:
This addresses a problem with the individual RollupExecutor actors not being able to get work from their local RollupManager due to it being behind a ConsistentHashPool.  Instead of the pool instantiating the RollupManager it now instantiates a RollupForwarder which in turn references the local RollupManager.  This allows the RollupManager to be created as an eager singleton and passed to the RollupExecutors so they can talk to the local instance.

### DIFF
--- a/app/com/arpnetworking/rollups/RollupForwarder.java
+++ b/app/com/arpnetworking/rollups/RollupForwarder.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2019 Dropbox Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


app/com/arpnetworking/rollups/RollupForwarder.java
```diff
@@ -0,0 +1,50 @@
...
+package com.arpnetworking.rollups;
+
+import akka.actor.AbstractActor;
+import akka.actor.ActorRef;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+
+/**
+ * @author Gilligan Markham (gmarkham at dropbox dot com)
```
vjkoskela: nit: incomplete javadoc

PR #308, opened 2019-10-10
Tag exclusions cleanup

vjkoskela:


app/com/arpnetworking/kairos/client/models/SamplingUnit.java
```diff
@@ -53,4 +57,18 @@
...
+     * NOTE: KairosDb accepts either upper or lower case representation. This
+     * model is otherwise a pass-through but converts all enumeration values
+     * to lower case.
+     *
+     * @return json encoded value
+     */
+    @JsonValue
+    public String toJson() {
+        return name().toLowerCase(Locale.getDefault());
+    }
```
speezepearson: nit: If Kairos isn't picky about casing, why do we need this method?

test/java/com/arpnetworking/testing/SerializationTestUtils.java
```diff
@@ -29,7 +37,40 @@
 public final class SerializationTestUtils {
 
     // TODO(spencerpearson): allow the ObjectMapper to get passed in, if ever necessary
-    private static final ObjectMapper OBJECT_MAPPER = ObjectMapperFactory.getInstance();
+    // IMPORTANT: The configuration here must match that in MainModule for testing REST APIs
```
speezepearson: Doesn't have to be part of this diff, but: would it be feasible to refactor the ObjectMapper-instantiation out of MainModule into something that the test could import?

test/java/com/arpnetworking/kairos/client/models/MetricsQueryTest.java
```diff
@@ -33,4 +33,28 @@ public void testTranslationLosesNothing() throws Exception {
...
+    @Test
+    public void testStartRelative() throws Exception {
+        SerializationTestUtils.assertTranslationLosesNothing(
+                ResourceHelper.loadResource(getClass(), "testStartRelative"),
+                MetricsQuery.class
+        );
+    }
+
+    @Test
+    public void tesEndRelative() throws Exception {
```
speezepearson: nit: `test`

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
```diff
@@ -224,72 +202,66 @@ private static MetricsQuery useAvailableRollups(
...
                     }
-
-                    return metric;
-                }
-            }
-        })
-                .collect(ImmutableList.toImmutableList()));
-
-
-        return newQueryBuilder.build();
```
speezepearson: I skimmed this and verified that it looked like you'd basically just indented all this as a side effect of switching to ThreadLocalBuilder. Let me know if you actually changed the logic inside.

app/com/arpnetworking/kairos/client/models/SamplingUnit.java
vjkoskela: You are correct, we don't. However, based on the documentation (which admittedly is not great) most of the examples were lower case. So given the decision of LC+UC=>UC or LW+UC=>LC I elected the latter which seemed more consistent with the documentation.

test/java/com/arpnetworking/testing/SerializationTestUtils.java
vjkoskela: I give it a shot on the next one.

test/java/com/arpnetworking/kairos/client/models/MetricsQueryTest.java
vjkoskela: Fixed!

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
vjkoskela: There were other formatting and naming changes to this block aside from the indentation; however, there were no logical changes.

Pro Tip: 
* Click the gear icon at the top of the review
* Check `Hide whitespace changes`

PR #310, opened 2019-10-14
Add tag exclusions to query metrics and response.

vjkoskela:
Removes filter tags based on configured excluded tags from query response unless explicitly requested in the metrics query. Next step is to evaluate what if anything needs to be done for group-by.

### DIFF
--- a/test/java/com/arpnetworking/metrics/portal/integration/controllers/KairosDbProxyControllerIT.java
+++ b/test/java/com/arpnetworking/metrics/portal/integration/controllers/KairosDbProxyControllerIT.java
@@ -33,6 +33,7 @@ import com.arpnetworking.testing.SerializationTestUtils;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMultimap;
 import org.apache.http.HttpHeaders;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpGet;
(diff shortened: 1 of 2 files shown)


test/java/com/arpnetworking/metrics/portal/integration/controllers/KairosDbProxyControllerIT.java
```diff
@@ -167,8 +168,74 @@ public void testTagQuery() throws ExecutionException, InterruptedException {
...
 
+    @Test
+    public void testQueryMetricsExcludeFilterTagInResponse() throws ExecutionException, InterruptedException {
+        _kairosDbClientDirect.addDataPoints(ImmutableList.of(
+                new MetricDataPoints.Builder()
+                        .setName(_metricName)
+                        // NOTE: KDB requires every metric to have at least one tag!
+                        .setTags(ImmutableMap.of(
+                                "host", "host.example.com",
+                                "ignored", "foo"))
```
speezepearson: I'm not sure I understand this comment (or the purpose of this test) -- if you didn't set the `host` tag on the datapoint, would we throw because all the remaining tags are excluded?

vjkoskela: So the comment is a warning that all metrics added to KDB must have at least one tag set or the REST API of KDB will reject them. That was surprising to me, so I started calling it out in places. In this particular test, I am adding two tags so it's not really a concern.

Further, tags are _never_ filtered on  data add. In fact, all the data add calls use the `_kairosDbClientDirect` which bypasses our logic in MP and hits KDB directly. In fact, MP is a read-only proxy for KDB and cannot be used to write data to KDB.

The purpose of these two tests is that if data is written to KDB with a tag that we want to ignore, that tag is _NOT_ returned implicitly when a query for that metric is issued (KDB returns tags for metrics if there is just one value for the tag in the query result set). However, that tag SHOULD be returned if the query explicitly sets the tag on the filter even if it is configured to be ignored.

speezepearson: Ah, I see, because `assertMetricsQuery` asserts that there's only one tag. Got it.

PR #311, opened 2019-10-17
Filtering & Group By

vjkoskela:
Complete data model for group by and add tests ensuring that tag filtering does not impact group by.

### DIFF
--- a/extra-checkstyle-suppressions.xml
+++ b/extra-checkstyle-suppressions.xml
@@ -23,6 +23,6 @@
     <suppress checks=".*" files="target/routes/main/controllers/routes.java"/>
 
     <!-- Exclude large nested external data models -->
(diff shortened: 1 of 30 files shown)


extra-checkstyle-suppressions.xml
```diff
@@ -21,4 +21,8 @@
 <suppressions>
     <!-- Exclude auto-generated code -->
     <suppress checks=".*" files="target/routes/main/controllers/routes.java"/>
+
+    <!-- Exclude large nested external data models -->
+    <suppress checks=".*" files="app/com/arpnetworking/kairos/client/models/MetricsQuery.java"/>
```
speezepearson: nit: why turn off all checks instead of just... `FileLength`, or whichever (presumably few) checks are actually going off?

vjkoskela: Good catch! That's just a copy-pasta error.

PR #312, opened 2019-10-24
add configuration option to use deprecated DevTools request-interception protocol

speezepearson:
- add enum for whether to use old or new DevTools protocol
- refactor request-interception code into that enum (copying old code from 8c0fe5cd25ac3f28d1b337af5ce2155893f70b95)
- add configuration option from app-config to control which strategy to use

### DIFF
--- a/app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsNetworkConfigurationProtocol.java
+++ b/app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsNetworkConfigurationProtocol.java
@@ -26,11 +26,13 @@ import com.github.kklisura.cdt.protocol.types.network.RequestPattern;
 import com.github.kklisura.cdt.services.ChromeDevToolsService;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import java.util.Collection;
 import java.util.Map;
 import java.util.function.BiConsumer;
 
+@SuppressFBWarnings("SE_BAD_FIELD") // False positive: https://github.com/spotbugs/spotbugs/issues/740
 /* package private */ enum DevToolsNetworkConfigurationProtocol {
 
     @Deprecated
--- a/app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsNetworkConfigurationProtocol.java
+++ b/app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsNetworkConfigurationProtocol.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2019 Dropbox, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.arpnetworking.metrics.portal.reports.impl.chrome;
+
+import com.arpnetworking.steno.Logger;
+import com.arpnetworking.steno.LoggerFactory;
+import com.github.kklisura.cdt.protocol.commands.Fetch;
+import com.github.kklisura.cdt.protocol.commands.Network;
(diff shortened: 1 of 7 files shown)


app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsNetworkConfigurationProtocol.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    public void configure(final ChromeDevToolsService dts, final PerOriginConfigs originConfigs) {
+        _configure.accept(dts, originConfigs);
+    }
+
+    private final BiConsumer<ChromeDevToolsService, PerOriginConfigs> _configure;
+
+    @SuppressWarnings("deprecation")
+    private static void configureWithNetwork(final ChromeDevToolsService dts, final PerOriginConfigs originConfigs) {
```
speezepearson: (this comes from https://github.com/ArpNetworking/metrics-portal/blob/8c0fe5cd25ac3f28d1b337af5ce2155893f70b95/app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsServiceWrapper.java#L74 )

app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsNetworkConfigurationProtocol.java
```diff
@@ -0,0 +1,137 @@
...
+                    url,
+                    event.getRequest().getMethod(),
+                    event.getRequest().getPostData(),
+                    headers,
+                    null
+            );
+        });
+    }
+
+    private static void configureWithFetch(final ChromeDevToolsService dts, final PerOriginConfigs originConfigs) {
```
speezepearson: (this comes from https://github.com/ArpNetworking/metrics-portal/blob/6cf37e6eb6ce1ed96420479b29b6dec1acaf0fde/app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsServiceWrapper.java#L103 )

test/java/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsServiceWrapperTest.java
```diff
@@ -142,66 +132,4 @@ public void testFiltersNavigation() {
         }
     }
 
-    @Test
-    public void testFiltersRequests() {
```
speezepearson: (these tests were extracted into `DevToolsNetworkConfigurationProtocolTest`)

app/com/arpnetworking/metrics/portal/reports/impl/chrome/DevToolsNetworkConfigurationProtocol.java
```diff
@@ -0,0 +1,137 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.function.BiConsumer;
+
+/* package private */ enum DevToolsNetworkConfigurationProtocol {
+
+    DEPRECATED_NETWORK(DevToolsNetworkConfigurationProtocol::configureWithNetwork),
```
vjkoskela: You should comment as to which versions each work with. Also, instead of prefix naming `NETWORK` with `DEPRECATED_` please just mark it deprecated with `@deprecated`.

app/com/arpnetworking/metrics/portal/reports/ReportExecutionContext.java
```diff
@@ -174,6 +176,16 @@
                                         .setTimeRange(timeRange),
                                 report.getTimeout()
                         )
+                        .whenComplete((res, err) -> {
```
vjkoskela: Is there a way to just handle the error instead of the success/failure?

app/com/arpnetworking/metrics/portal/reports/ReportExecutionContext.java
```diff
@@ -201,6 +213,16 @@
                 .stream()
                 .map(entry -> getSender(entry.getKey())
                         .send(entry.getKey(), mask(formatToRendered, entry.getValue()))
+                        .whenComplete((result, err) -> {
```
vjkoskela: Same

app/com/arpnetworking/metrics/portal/reports/ReportExecutionContext.java
```diff
@@ -174,6 +176,16 @@
                                         .setTimeRange(timeRange),
                                 report.getTimeout()
                         )
+                        .whenComplete((res, err) -> {
+                            if (err != null) {
+                                LOGGER.warn()
```
vjkoskela: Does the error from this future get swallowed here? It seems like we should leave the logging to whomever is waiting on the future's result (e.g. don't log and throw).

PR #318, opened 2019-10-31
fix exception when a report has multiple executions; and subtly change semantics for scheduling

speezepearson:
Currently, `DatabaseReportRepository.getJobLastRun()` will throw if a job has multiple completed executions. This fixes that.

Along the way, I decided that the semantics for that method were wrong, and it should return
`job.executions().filter(hasCompleted).map(scheduledFor).max()`
rather than
`job.executions().filter(hasCompleted).map(completedAt).max()`
, in order to catch up in weird edge cases where one run doesn't complete until after it is next scheduled to start.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/scheduling/JobRepository.java
+++ b/app/com/arpnetworking/metrics/portal/scheduling/JobRepository.java
@@ -53,14 +53,14 @@ public interface JobRepository<T> {
     Optional<Job<T>> getJob(UUID id, Organization organization);
 
     /**
-     * Get the last time that a job with a given UUID was run.
+     * Get the last time an execution for a job was scheduled to run.
      *
      * @param id The id assigned to the Job by a previous call to {@code add}.
      * @param organization The organization owning the job.
-     * @return The last time that that job was executed.
+     * @return The last time that an execution for that job was scheduled to run.
      * @throws NoSuchElementException if no job has the given UUID.
      */
-    Optional<Instant> getJobLastRun(UUID id, Organization organization) throws NoSuchElementException;
+    Optional<Instant> getJobLastExecutionScheduled(UUID id, Organization organization) throws NoSuchElementException;
 
     /**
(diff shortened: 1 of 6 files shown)


app/com/arpnetworking/metrics/portal/scheduling/JobRepository.java
```diff
@@ -53,14 +53,16 @@
...
+     * More precisely, in pseudocode, {@code job.executions().filter(hasCompleted).map(scheduledFor).max()}.
      *
      * @param id The id assigned to the Job by a previous call to {@code add}.
      * @param organization The organization owning the job.
-     * @return The last time that that job was executed.
+     * @return The last time that any completed execution for the job was scheduled for.
      * @throws NoSuchElementException if no job has the given UUID.
      */
-    Optional<Instant> getJobLastRun(UUID id, Organization organization) throws NoSuchElementException;
+    Optional<Instant> getJobLastExecutionScheduled(UUID id, Organization organization) throws NoSuchElementException;
```
vjkoskela: Naming: Seems somewhat ambiguous in the sense that it is for the last completed job; although I get how the other was misleading as well, that it returns the scheduled date and not the run (start/end) date/time.

Perhaps: `getJobLastRunScheduledAt` -- not a fan of "run", but I want the name to convey that its the scheduled date time for a job that has completed (success or failure); `getMostRecentJobCompletedScheduledDate` (wow that's a mouthful) ... `getLastJobCompletedScheduledDate` ... sigh naming sucks.

speezepearson: Yeah, I think that this concept is inherently high-Kolmogorov-complexity, so any name must be either verbose or ambiguous. I tend to prefer "verbose" to "ambiguous," and would lean towards `getLastScheduledTimeWhereExecutionCompleted`, but I could also do `getMostRecentJobCompletedScheduledDate` as you suggest.

PR #319, opened 2019-10-31
remove concept of "triggering event" from web-based reports

speezepearson:
It's never even used in the code currently; and the current behavior of "simple web pages get screenshotted on `load`, Grafana pages get screenshotted on `reportrendered`" seems perfectly adequate for all foreseeable use cases.


vjkoskela:
Just to be clear that `reportrendered` is a built-in for the Chrome dev tools?

speezepearson:
> Just to be clear that `reportrendered` is a built-in for the Chrome dev tools?

i.e. that it's hardcoded? [Exactly right.](https://github.com/ArpNetworking/metrics-portal/blob/edb986e0edb4a8597892761424c1aa0e24d0bbaa/app/com/arpnetworking/metrics/portal/reports/impl/chrome/grafana/BaseGrafanaScreenshotRenderer.java#L82)

PR #328, opened 2019-12-03
allow configuration of what report format types are displayed in the UI.

speezepearson:
Follows the same pattern as source-type configuration.

### DIFF
--- a/app/assets/javascripts/classes/reports/Models.ts
+++ b/app/assets/javascripts/classes/reports/Models.ts
@@ -26,7 +26,18 @@ export enum ScheduleRepetition {
     MONTHLY,
 }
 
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/classes/reports/Models.ts
```diff
@@ -26,7 +26,18 @@ export enum ScheduleRepetition {
     MONTHLY,
 }
 
+function mimeTypeToReportFormat(mimeType: string): ReportFormat | undefined {
+    if (/^application\/pdf\b/.test(mimeType)) {
+        return ReportFormat.HTML;
```
orborde: ITYM PDF?

PR #331, opened 2020-02-24
Nullability Fix for Gaps Aggregator

vjkoskela:
Make data point optional to support gaps aggregator. Remove internal model conversion instead of propagating change down deprecated path.

### DIFF
--- a/app/com/arpnetworking/kairos/client/models/DataPoint.java
+++ b/app/com/arpnetworking/kairos/client/models/DataPoint.java
@@ -21,13 +21,12 @@ import com.fasterxml.jackson.annotation.JsonValue;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import models.internal.TimeSeriesResult;
-import models.internal.impl.DefaultTimeSeriesResult;
 import net.sf.oval.constraint.NotNull;
 
 import java.time.Instant;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 /**
  * Model class for a data point in a kairosdb metrics query.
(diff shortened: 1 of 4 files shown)


app/com/arpnetworking/kairos/client/models/DataPoint.java
```diff
@@ -89,11 +76,11 @@ public String toString() {
 
     private DataPoint(final Builder builder) {
         _time = builder._time;
-        _value = builder._value;
+        _value = Optional.ofNullable(builder._value);
     }
 
     private final Instant _time;
-    private final Object _value;
+    private final Optional<Object> _value;
```
cwbriones: From what [I've read in the past](https://blog.joda.org/2014/11/optional-in-java-se-8.html), the recommendation was to only use Optional.ofNullable in your accessor methods and store a nullable field.

So this would become
```suggestion
@Nullable
private final Object _value;
```

and the above
```java
    public Optional<Object> getValue() {
        return Optional.ofNullable(_value);
    }
```

BrandonArp: One of the main reasons for this recommendation is that Optional is not Serializable.  Since we use Jackson, we actually serialize as nullable in json and this won't bite us.  Using Optional like this is what we usually do, even though it's not Java standard.

cwbriones: Makes sense. 

I think another point was that it could contribute to increased GC pressure if you have many short-lived objects with optional fields. But since neither of those cases apply here and theres precedent 🤷‍♂️

vjkoskela: To quote a colleague, "we should just use Kotlin".

cwbriones: I wonder who could have said that? 🤔

PR #338, opened 2020-03-10
[rollups] support per-period backfill configuration

cwbriones:
Modify the configuration to allow for different backfill lengths based on the period, e.g. backfill daily for 7 days and hourly for 168 hours in order to have both units backfilled to the same point in the past.

If a backfill length is not configured for a period, we explicitly avoid generating any rollup jobs for that period.


app/com/arpnetworking/rollups/RollupExecutor.java
```diff
@@ -17,7 +17,7 @@
 
 import akka.actor.AbstractActorWithTimers;
 import akka.actor.ActorRef;
-import akka.pattern.PatternsCS;
+import akka.pattern.Patterns;
```
cwbriones: `PatternsCS` is deprecated in favor of `Patterns`

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -51,11 +55,33 @@
...
+ * <p>
+ * The {@code RollupGenerator} will periodically retrieve metrics eligible for rollup and enqueue jobs up
+ * to the maximum number of backfill periods, if any are configured.
  *
  * @author Gilligan Markham (gmarkham at dropbox dot com)
+ *
  */
 public class RollupGenerator extends AbstractActorWithTimers {
+
+    /*
```
cwbriones: Not sure where the right place for this is but I would really like some form of in-line documentation explaining the state machine like this. Ideally I would still be able to have it processed by javadoc for formatting, for now I've opted for this unattached comment as I've done here.

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -185,41 +222,45 @@ private void handleTagNamesMessage(final TagNamesMessage message) {
...
-                final Instant startOfRecentClosedPeriod = period.recentStartTime(_clock.instant());
+                final Instant oldestBackfillPoint = period.recentEndTime(_clock.instant())
+                        .minus(period.periodCountToDuration(maxBackFillPeriods));
 
-                // We either want to start at the oldest backfill point or the start of the period after the last datapoint.
+                // We either want to start at the oldest backfill point or the start of the period
+                // after the last datapoint since it contains data for the period that follows it.
                 Instant rollupPeriodStart = lastDataPoint.isBefore(oldestBackfillPoint)
-                        ? oldestBackfillPoint : lastDataPoint.plus(period.periodCountToDuration(1));
+                        ? oldestBackfillPoint : period.recentEndTime(lastDataPoint).plus(period.periodCountToDuration(1));
```
cwbriones: This line had a bug where it didn't align the last data point to the most recent period. I fixed that and update the test.

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -281,22 +322,36 @@ private LastDataPointMessage buildLastDataPointResponse(
         if (failure != null) {
             builder.setFailure(failure);
         } else {
```
cwbriones: I changed the following section to use `streams/map/flatMap` because the chain of `get()` calls to nullable return values gave me anxiety

PR #340, opened 2020-03-18
Create a whitelist for rollup usage in proxied KairosDB queries

cwbriones:
When evaluating queries aggregated over hourly/daily windows, metrics portal will currently use the largest rollup series available. There have been some correctness issues with daily rollups and so this change allows us to selectively enable the series based on the sampling unit. This way we can continue to use hourly rollups, which would still be significantly better for queries on larger timescales.


conf/portal.application.conf
```diff
@@ -151,6 +153,14 @@ kairosdb {
 kairosdb.proxy {
   filterRollups = true
   excludedTagNames = []
+
+  # For enabling/disabling usage of a particular rollup in a query
+  rollups.blacklist = [
```
cwbriones: I figured this belongs in the `proxy` config rather than the `rollups` config, since really it's all used on the proxy side when querying kairos.

The labels for periods lead to a somewhat annoying inconsistency (`hours` vs `hourly` in the rollup config), but that's because I wanted to continue to use `SamplingUnit` from the proxy code instead of importing `RollupPeriod`.

app/com/arpnetworking/kairos/config/MetricsQueryConfigImpl.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+public class MetricsQueryConfigImpl implements MetricsQueryConfig {
+
+    /**
+     * Construct a new {@code MetricsQueryImpl} by parsing a {@link Config}.
+     *
+     * @param configuration Play configuration to load from
+     */
+    @Inject
+    public MetricsQueryConfigImpl(final Config configuration) {
```
cwbriones: I'd ordinarily want to separate out the configuration here but that would require a lot more boilerplate, especially since `buildBlacklistEntry` below has to crawl the config list. It also seems like configuration is directly injected elsewhere in the codebase.

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
```diff
@@ -234,56 +237,64 @@ private static MetricsQuery useAvailableRollups(
                         return Metric.Builder.<Metric, Metric.Builder>clone(metric)
                                 .setName(metricName.substring(0, metricName.length() - 2))
                                 .build();
+                    }
```
cwbriones: The following section, up until the `maxUnit.isPresent()` code, is me repeatedly applying Intellij's "remove redundant else" refactor

app/com/arpnetworking/kairos/config/MetricsQueryConfigImpl.java
```diff
@@ -0,0 +1,88 @@
...
+                .map(MetricsQueryConfigImpl::buildBlacklistEntry)
+                .collect(ImmutableList.toImmutableList());
+    }
+
+    @Override
+    public Set<SamplingUnit> getQueryEnabledRollups(final String metricName) {
+        final Set<SamplingUnit> disabledUnits = queryBlacklistEntry(metricName)
+                .map(RollupQueryBlacklistEntry::getPeriods)
+                .orElse(ImmutableSet.of());
+        if (disabledUnits.isEmpty()) {
```
vjkoskela: If the periods key is absent then we can imply ALL, but an explicit empty array seems very confusing way to denote ALL and excludes the possibly of communicating NONE.

conf/portal.application.conf
vjkoskela: I think it is surprising that an empty array leads to suppressing all rollup usage. At the very least the documentation should make that clear. Ideally, periods should then be optional, and its absence would imply ALL, but an empty array would be a no-op.

vjkoskela:
So because we cannot compute daily roll-ups we want to disable their usage. I guess that makes sense, but we should probably stop computing them too (for the time being) if we can't get it to scale. I assume we already have that config option available and that this is just to stop using the existing (often) partial daily roll-ups.

conf/portal.application.conf
cwbriones: Perhaps there's a misunderstanding or I can make this clearer? The presence of any value in this list can only disable rollups, the absence cannot do the same.

**Example 1:** hourly rollups are disabled for `some_metric`
```
rollups.blacklist = [
      {
         pattern = "^some_metric$",
         periods = ["hours"],
      }
  ]
]
```

**Example 2:** A pattern matches `some_metric`, but the disabled periods list is empty so nothing is disabled. (This is a strange case, but legal)
```
rollups.blacklist = [
      {
         pattern = "^some_metric$",
         periods = [],
      }
  ]
]
```

**Example 3:** Empty config. *No* rollups are disabled for any metric
```
rollups.blacklist = []
```

I would consider making Example 2 invalid, since it doesn't convey any useful information and could be misleading.

vjkoskela: Sorry, so it's not possible to disable all roll-up periods for a metric; e.g.

```
rollups.blacklist = [
      {
         pattern = "^some_metric$",
      }
  ]
]
```

PR #341, opened 2020-03-25
hierarchical rollup periods

cwbriones:
Hierarchical in the chronological sense - dailys will be computed from the hourly rollups, for example.

The major changes here are to 
1. Plumb through both the source and destination metrics into `LastDataPointMessage` and query for both to detect how far back we can roll up.
2. Change the source metric such that `metric_1d` has source `metric_1h`, and more generally such that each rollup's source becomes the next smallest rollup.

As a consequence rollups are no longer independent and so it could be the case that if the hourly is behind, the daily will be behind as well.


app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -310,60 +348,72 @@ private void handleNoMoreMetricsMessage(final NoMoreMetrics message) {
...
-                builder.setLastDataPointTime(lastDataPointTime);
-            }
+                        .collect(ImmutableMap.toImmutableMap(
+                                MetricsQueryResponse.QueryResult::getName,
+                                Function.identity()
+                        ));
+
+        // Query results should *only* contain the source and destination metric.
+        if (queryResults.size() != 2 || !queryResults.containsKey(sourceMetricName) || !queryResults.containsKey(rollupMetricName)) {
+            return builder.setFailure(new Exception("Unexpected query results.")).build();
```
vjkoskela: You should really attach the query results to this exception. Possibly a new exception type?

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -310,60 +348,72 @@ private void handleNoMoreMetricsMessage(final NoMoreMetrics message) {
...
         }
+
+        // Set source time, if any.
+        Optional.ofNullable(queryResults.get(sourceMetricName))
```
vjkoskela: The value should not be null after the check above. Right? If we're going to enforce the preconditions, we should simplify the resulting code.

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -73,9 +74,10 @@
      * For each metric name received, perform the following:
      *
      *     1. Retrieve the tag names for this metric (TagNamesMessage)
-     *     2. Query for the last data point given the maximum backfill and set of tags (LastDataPointMessage)
-     *     3. If this datapoint is in the past, generate a backfill job for the period
-     *     furthest in the past, and enqueue it by sending to the RollupManager. (FinishRollupMessage)
+     *     2. For each period size:
+     *         2a. Query for the last data point given the maximum backfill and set of tags (LastDataPointMessage)
+     *         2b. If this datapoint is in the past, generate a backfill job for the period
+     *         furthest in the past, and enqueue it by sending to the RollupManager. (FinishRollupMessage)
```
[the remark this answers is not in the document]
cwbriones: Nothing right now, the assumption is there will never be mixed data since a second job would just overwrite the same datapoints.

That's all captured in a separate ticket.

app/com/arpnetworking/rollups/LastDataPointMessage.java
```diff
@@ -50,17 +58,21 @@ public RollupPeriod getPeriod() {
...
-    private final String _metricName;
+    private final String _sourceMetricName;
+    private final String _rollupMetricName;
     private final RollupPeriod _period;
     private final ImmutableSet<String> _tags;
-    private final Instant _lastDataPointTime;
-    private static final long serialVersionUID = 5745882770658263619L;
+    private final Instant _sourceLastDataPointTime;
+    private final Instant _rollupLastDataPointTime;
+    private static final long serialVersionUID = 2800761302248621189L;
```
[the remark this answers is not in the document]
BrandonArp: Uh.... I have a cluster using rollups....

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -232,53 +250,73 @@ private void handleLastDataPointMessage(final LastDataPointMessage message) {
...
+            final Instant startOfLastEligiblePeriod =
+                message.getSourceLastDataPointTime()
+                        .map(period::recentStartTime)
+                        .orElse(Instant.EPOCH);
 
             // If the most recent period aligned start time is after the most recent datapoint then
             // we need to run the rollup, otherwise we can skip this and just send a finish message.
-            if (startOfRecentClosedPeriod.isAfter(lastDataPoint)) {
-                final String rollupMetricName = metricName + period.getSuffix();
+            if (startOfLastEligiblePeriod.isAfter(lastRollupDataPoint)) {
```
speezepearson: (Does the hourly rollup for 10:00-11:00 land on 10:00 or 11:00? There's opportunity for a fencepost error here.)

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -232,53 +250,73 @@ private void handleLastDataPointMessage(final LastDataPointMessage message) {
...
 
                 final Instant oldestBackfillPoint = period.recentEndTime(_clock.instant())
                         .minus(period.periodCountToDuration(maxBackFillPeriods));
 
                 // We either want to start at the oldest backfill point or the start of the period
                 // after the last datapoint since it contains data for the period that follows it.
-                Instant rollupPeriodStart = lastDataPoint.isBefore(oldestBackfillPoint)
-                        ? oldestBackfillPoint : period.recentEndTime(lastDataPoint).plus(period.periodCountToDuration(1));
+                Instant rollupPeriodStart = lastRollupDataPoint.isBefore(oldestBackfillPoint)
+                        ? oldestBackfillPoint : period.recentEndTime(lastRollupDataPoint).plus(period.periodCountToDuration(1));
```
speezepearson: > `period.recentEndTime(lastRollupDataPoint)`

Why is the `period.recentEndTime()` necessary? It [looks to me](https://github.com/ArpNetworking/metrics-portal/blob/master/app/com/arpnetworking/rollups/RollupPeriod.java#L50) like `recentEndTime` just aligns a time, but... `lastRollupDataPoint` should already _be_ aligned, right?

cwbriones: That code is defensive, but you're correct – if it's coming from the rollup it should already be aligned.

app/com/arpnetworking/rollups/LastDataPointMessage.java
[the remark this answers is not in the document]
cwbriones: I could revert the field rename which should keep it compatible but I think the addition of `sourceMetricName` will still be a problem because it cannot be null in a valid message.

I'll consider adding a new message as suggested.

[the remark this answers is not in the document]
cwbriones: I duplicated/added a new message type. @vjkoskela and I had discussed this, worst-case the new/old clusters won't understand each other's messages but the job will be picked up on the next retry for the rollup. I opted for this instead of creating a delegating handler because it would not be possible for a new cluster to handle older messages since they will be missing required fields.

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -203,27 +204,44 @@ private void handleTagNamesMessage(final TagNamesMessage message) {
         final long startTime = System.nanoTime();
         for (final RollupPeriod period : RollupPeriod.values()) {
             final int backfillPeriods = _maxBackFillByPeriod.getOrDefault(period, 0);
+            final String rollupMetricName = metricName + period.getSuffix();
+            // Get the source of the rollup data, which is either the next smallest rollup metric or
+            // the original metric, if this is the smallest possible rollup.
+            final String sourceMetricName = period.previous()
```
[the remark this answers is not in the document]
cwbriones: Updated as discussed: daily will then pull from the original (minutely) metric.

app/com/arpnetworking/rollups/RollupGenerator.java
cwbriones: 10:00 - we align with the start time. So in this case, if the start of the next eligible period is after the last rollup time, that means this period is unaccounted for and can be rolled up.

app/com/arpnetworking/rollups/LastDataPointMessage.java
```diff
@@ -29,7 +29,9 @@
  * empty.
  *
  * @author Gilligan Markham (gmarkham at dropbox dot com)
+ * @deprecated Use {@link LastDataPointsMessage} instead.
```
vjkoskela: Please remove this in a subsequent diff.

app/com/arpnetworking/rollups/RollupPeriod.java
```diff
@@ -92,6 +95,23 @@ public SamplingUnit getSamplingUnit() {
...
+    public Optional<RollupPeriod> nextSmallest() {
+        final int i = this.ordinal();
+        if (i == 0) {
+            return Optional.empty();
+        }
+        return Optional.of(VALUES.get((i - 1) % VALUES.size()));
+    }
+
+    // values() will create a new array on every call to nextSmallest without this.
+    private static final List<RollupPeriod> VALUES = ImmutableList.copyOf(values());
```
vjkoskela: I mean if you want to get picky, we can go from O(N) to O(1) by computing the pair-wise "next smallest" in a `static {}` block. This is a regular pattern we've used with enumerations.

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -204,26 +204,52 @@ private void handleTagNamesMessage(final TagNamesMessage message) {
...
+                ).to(getSelf());
             }
         }
     }
 
-    private void handleLastDataPointMessage(final LastDataPointMessage message) {
+    private String getSourceMetricName(final String metricName, final RollupPeriod period) {
+        Optional<RollupPeriod> sourcePeriod = period.nextSmallest();
+        while (sourcePeriod.isPresent() && !sourcePeriod.map(p -> _maxBackFillByPeriod.getOrDefault(p, 0) > 0).orElse(true)) {
+            sourcePeriod = sourcePeriod.flatMap(RollupPeriod::nextSmallest);
```
vjkoskela: Is "next smallest" going to try all values less than the roll-up or only all _possible_ values less than the roll-up?

It will only work if it tries all the values, which for now is probably ok, but please leave a comment that this is potentially inefficient and also potentially incorrect.

Specifically, this is a GCD type problem which is a tree. For example, 1H can be computed from 1M, 15M, 30M roll-ups, but cannot be computed from 25M and 45M roll-ups. So walking down from 1H (60M) would walk consider impossible values like 45M and 25M and if not recognized as impossible could try to roll-up from them! (which may result in bad data)

The generalization of this is:
* Each _possible_ roll-up declares its ancestors (e.g. in the enum):
```
1H => [30M, 15M, 1M]
45M => [15M, 1M]
30M => [15M, 1M]
15M => [1M]
```
* Then the specified roll-up configuration (e.g. 1H, 30M) must look at the enumeration to determine what are possible ancestors for each specified roll-up. In this case those would be:
```
1H = [30M, 1M]
30M = [1M]
```
(the 15M possibility drops off because it's not configured!).

Anyway, I don't expect this to be solved now. However, @speezepearson should be aware of it and @cwbriones can you file an issue and add a comment to the roll-up enum describing this to remind us not to a possible roll-up that exposes this behavior.

cwbriones: I agree with the generalization but I also wrote this with the expectation that we guarantee each period size is divisible by each smaller period. So if we did have additional rollups for example I'd only expect something like `15m, 30m, 3h, 6h, 12h`. Of course, if that property does not hold then of course this will not work but I really don't see the value-add in complicating it for ourselves by doing it otherwise.

Regardless, I'll leave a comment in `RollupPeriod` since this code will not need to change even if we did add additional periods like this.

app/com/arpnetworking/rollups/RollupPeriod.java
cwbriones: `nextSmallest` is already O(1) as implemented.

I think you're suggesting it an alternative would be a lookup table or additional field per-instance?

vjkoskela: Yep

app/com/arpnetworking/rollups/RollupGenerator.java
vjkoskela: Totally fair.

Assuming we have such well behaved supported roll-ups but someone only enables 30min and 1h? Does it know not to try to use 15min?

cwbriones: Yes.

Maybe I should leave a comment, but this will walk backwards down "nextSmallest" until it finds the first one that's enabled.

So it would go - 
1. computing hourly
    * next smallest is 30m. it's enabled. use 30m
2. computing 30m, 
    * next smallest is 15m. it's disabled.
    * next smallest is 1m. use 1m.

cwbriones: [I included a test for this in the most recent push](https://github.com/ArpNetworking/metrics-portal/blob/54f1d26dc31db25b2acc3242ab17e4ce572fe3c7/test/java/com/arpnetworking/rollups/RollupGeneratorTest.java#L234)

cwbriones: I've thought about it more and I've found a case that hits the issue you've discussed.

Consider series with period `30m 15m 10m 1m`, and suppose we want to compute the `30m` rollup. All periods are enabled *except 15m*.

The current code will walk: 30m -> 15m (disabled) -> 5m and choose the `5m` rollup to use as the source instead of the `10m` because 10m does not evenly divide 15. This is non-optimal since it does divide 30.

This behavior doesn't occur in the `1h 45m 30m 15m` example.

PR #344, opened 2020-04-02
Minor tweaks to reporting user interface.

vjkoskela:


vjkoskela:
Changes include:
* Clickable + dismissable popover help (instead of hover help)
* Configured source help only
* Increased report completion logging level
* Updated instructions for running MP given port change

PR #348, opened 2020-04-08
Allow control of various aspects of report configuration.

vjkoskela:
Extends the configuration, client-side injected features, as well as report ui code to support configuration of:

* Sources
* Formats
* Recipients
* Intervals

All are enabled by default, although reporting itself is off by default.

### DIFF
--- a/app/assets/javascripts/classes/reports/EditReportViewModel.ts
+++ b/app/assets/javascripts/classes/reports/EditReportViewModel.ts
@@ -264,8 +264,7 @@ class EditSourceViewModel extends BaseSourceViewModel {
                 "the browser and the page content is taken as the rendered report. This <i>only</i> supports static " +
                 "content.</li>",
         [SourceType.GRAFANA]: "<li class='list-group-item'><b>Grafana</b> - Specialization of the Web Page source " +
-                "for capturing page content from Grafana based reports which may contain dynamic content. Additionally, " +
-                "these sources pass the scheduled date as the </li>",
+                "for capturing page content from a Grafana based report which may contain dynamic content.</li>",
     };
     readonly availableSourceTypes: {value: SourceType, text: string}[] = availableSourceTypes.map(
         sourceType => ({value: sourceType, text: EditSourceViewModel.sourceTypeDisplayNames[sourceType]})
--- a/app/assets/javascripts/classes/reports/EditReportViewModel.ts
+++ b/app/assets/javascripts/classes/reports/EditReportViewModel.ts
@@ -23,6 +23,9 @@ import csrf from '../Csrf';
 
 import {
     availableSourceTypes,
+    availableReportFormats,
+    availableRecipientTypes,
+    availableReportIntervals,
     BaseRecipientViewModel,
     BaseScheduleViewModel,
     BaseSourceViewModel,
@@ -140,14 +143,6 @@ class EditReportViewModel {
         }
     }
 
-    readonly availableRecipientTypes = [
-        {value: RecipientType.EMAIL,  text: "Email"},
-    ];
-
-    readonly helpMessages = {
-        timeout: "Time that can be spent rendering/sending the report before forcibly halting execution. HH:MM:SS or ISO-8601.",
-    }
-
     private static parseProblems(responseJson: string): string[] {
         try {
             return JSON.parse(responseJson).errors;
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/classes/reports/EditReportViewModel.ts
```diff
@@ -232,14 +256,16 @@ class EditSourceViewModel extends BaseSourceViewModel {
...
-        [SourceType.WEB_PAGE]: "<li class='list-group-item'><b>Browser rendered</b> - A URL is loaded in the browser and the rendered " +
-                "contents of the page are taken as the generated report.</li>",
-        [SourceType.GRAFANA]: "<li class='list-group-item'><b>Grafana</b> - Like browser-rendered, but tweaked to pull data from a Grafana Report panel" +
-                " (which are loaded asynchronously after page-load, hence needing the specialization).</li>",
+        [SourceType.WEB_PAGE]: "<li class='list-group-item'><b>Browser rendered</b> - The specified URL is loaded in " +
+                "the browser and the page content is taken as the rendered report. This <i>only</i> supports static " +
+                "content.</li>",
+        [SourceType.GRAFANA]: "<li class='list-group-item'><b>Grafana</b> - Specialization of the Web Page source " +
+                "for capturing page content from Grafana based reports which may contain dynamic content. Additionally, " +
+                "these sources pass the scheduled date as the </li>",
```
cwbriones, line 268: looks like this sentence got cut short

public/html/reports/ReportsViewModel.html
```diff
@@ -46,7 +46,6 @@
                             <div class="popover-content hidden" data-bind="attr: {'data-id': 'source-'+id}">
                                 <dl class="container col-md-12 dl-horizontal">
                                     <dt>url:</dt><dd data-bind="text: source.url"></dd>
-                                    <dt>event:</dt><dd data-bind="text: source.eventName"></dd>
```
cwbriones, line 49: I'm guessing this was removed because the event name was only exposed while we tweaked the report job, and now that it's working users should never need to change that.

Can you confirm this is the case?

app/models/internal/impl/DefaultFeatures.java
```diff
@@ -69,8 +72,23 @@ public boolean isReportsEnabled() {
     }
 
     @Override
-    public ImmutableList<String> getSourceTypes() {
-        return _sourceTypes.stream().map(SourceType::name).collect(ImmutableList.toImmutableList());
+    public ImmutableList<SourceType> getReportingSourceTypes() {
```
cwbriones, line 75: It seems odd that all of these methods return concrete types but I've seen this elsewhere in the repo, specifically with `ImmutableList` so 🤷‍♂

public/html/reports/ReportsViewModel.html
vjkoskela: This was actually removed altogether as a feature in a previous commit and the user interface code was leftover. 

Background: We initially thought that we could build a generic interactive web page source that would monitor for a javascript event to mark completion and that would trigger render. However, it became apparent that Grafana required many more customizations to render nicely. Therefore, we nixed the plans for a general interactive web source and just created a static web source and a specialization of that (in many ways) for Grafana. We can still add a general interactive web source in the future, but there's no motivating use case for it at the moment.

app/models/internal/impl/DefaultFeatures.java
vjkoskela: Why is this odd that they return types (an enum) instead of strings? I would have thought returning types was the norm and returning strings was strange. Can you elaborate?

app/assets/javascripts/classes/reports/EditReportViewModel.ts
vjkoskela: Good catch. I was under the impression that we set from/to on the report that we generated based on the scheduled date, but it seems that we do not. I've cataloged this work in a document where I am tracking Reporting v1.1 work items.

app/models/internal/impl/DefaultFeatures.java
cwbriones: I meant that it returns `ImmutableList` vs `List`

vjkoskela: That's fair. This is the convention we adopted for the ArpNet/ISM projects since imo it was clearer what the intention was and we wanted to be clear on where data was mutable.

PR #349, opened 2020-04-09
add ?from=...&to=... to Grafana report URIs

speezepearson:


app/com/arpnetworking/metrics/portal/reports/impl/chrome/grafana/BaseGrafanaScreenshotRenderer.java
```diff
@@ -67,8 +69,15 @@ public boolean getIgnoreCertificateErrors(final GrafanaReportPanelReportSource s
     }
 
     @Override
-    public URI getUri(final GrafanaReportPanelReportSource source) {
-        return source.getWebPageReportSource().getUri();
+    public URI getUri(final GrafanaReportPanelReportSource source, final TimeRange timeRange) {
+        try {
+            return new URIBuilder(source.getWebPageReportSource().getUri())
+                    .setParameter("from", Long.toString(timeRange.getStart().getEpochSecond()))
```
vjkoskela, line 75: Unfortunately, I'm not sure that Grafana does the "right thing" with epoch from/to values.

e.g.
`?from=1586787676&to=1586809276`

1586787676 = Monday, April 13, 2020 2:21:16 PM (GMT)
1586809276 = Monday, April 13, 2020 8:21:16 PM  (GMT)

It looks to me that the values are in fact millisecond epochs.

vjkoskela: image

vjkoskela: Also, this dashboard was configured as "UTC". However, I have seen "bugs" where developers treat epoch as "local". It may be worthwhile testing that if a report were configured as local timezone, that supplying the epoch values still causes it to render the same time interval as if the report were configured UTC (e.g. epoch interpretation is locale agnostic -- if it's not, this quirk will need to be addressed in our plugin ... somehow ... or at least documented)

app/com/arpnetworking/metrics/portal/reports/impl/chrome/BaseScreenshotRenderer.java
```diff
@@ -107,7 +109,7 @@
                 .addData("format", format)
                 .addData("timeRange", timeRange)
                 .log();
-        final CompletableFuture<Void> navigate = dts.navigate(getUri(source).toString());
+        final CompletableFuture<Void> navigate = dts.navigate(getUri(source, timeRange).toString());
```
vjkoskela, line 112: Just want to confirm that the `timeRange` here is the fixed target and independent of any scheduling and jitter. (e.g. a daily midnight to midnight report with a 1h delay will still have a timerange of midnight to midnight)

PR #351, opened 2020-04-14
fix typo in comment / variable name

speezepearson:
We're computing this with a `min()`, but it's called `max`.


cwbriones:
It's called max unit because it's the largest rollup unit we can use without making any assumptions about the particular data being rolled up. 

That's how I think of it at least, but I guess I can see how that is confusing.

speezepearson:
Obsoleted by #352

PR #352, opened 2020-04-15
Fix query-rewriting for rollups

speezepearson:
The logic for determining whether a query can be optimized to use rollups / which rollup to use is broken, and it will sometimes use a rollup even when it shouldn't.

## Demonstration
This is a problem in Dropbox's production environment: [Grafana](http://drl/q34th87h4378og)
Screenshot 2020-04-14 21 31 00
- There are two queries, identical except that one uses the `_!` feature to prevent use of rollups.
- Clearly, they should return the same value, because rollup-usage should be a pure optimization, invisible to the user. But they don't.

### Explanation
- The query has two aggregators: `count(1m)` and `count(1h)`. The first aggregator, though, is not sampling-aligned, which leads to [it being ignored](https://github.com/ArpNetworking/metrics-portal/blob/e8b8df96e566963f2dbd94c59a539fe5a5731091/app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java#L266) when MPortal is checking whether it can use rollups, so MPortal only sees the 1h aggregator and mistakenly thinks it can use rollups.
- The **correct** answer for this query is "one datapoint per hour, with value 60." (Reasoning: this is a very busy metric, with data every minute, so the `count(1m)` produces one datapoint (with irrelevant value) per minute, so the `count(1h)` should count 60 datapoints per hour.)  The `_!` ("force non-use of rollups") query gets this right.
- The **incorrect** answer is "one datapoint per hour, with value 1." This occurs when MPortal mistakenly uses the rollup and therefore the `count(1m)` produces only 1 datapoint per hour instead of 1 per minute. The non-`_!` query gets this wrong.

## Fix
Explained by a comment in the new `getMaxUsableRollupUnit` method.

### DIFF
--- a/app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
+++ b/app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
@@ -24,6 +24,7 @@ import com.arpnetworking.kairos.client.models.MetricNamesResponse;
 import com.arpnetworking.kairos.client.models.MetricTags;
 import com.arpnetworking.kairos.client.models.MetricsQuery;
 import com.arpnetworking.kairos.client.models.MetricsQueryResponse;
+import com.arpnetworking.kairos.client.models.Sampling;
 import com.arpnetworking.kairos.client.models.SamplingUnit;
 import com.arpnetworking.kairos.client.models.TagNamesResponse;
 import com.arpnetworking.kairos.client.models.TagsQuery;
@@ -31,6 +32,8 @@ import com.arpnetworking.kairos.config.MetricsQueryConfig;
 import com.arpnetworking.metrics.Metrics;
 import com.arpnetworking.metrics.MetricsFactory;
 import com.arpnetworking.metrics.Timer;
+import com.arpnetworking.steno.Logger;
+import com.arpnetworking.steno.LoggerFactory;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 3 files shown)


app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
```diff
@@ -219,7 +221,7 @@
         return response;
     }
 
-    private static MetricsQuery useAvailableRollups(
+    /* package private */ static MetricsQuery useAvailableRollups(
```
cwbriones, line 224: package private for tests?

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
```diff
@@ -299,6 +297,44 @@ private static MetricsQuery useAvailableRollups(
...
+         * If the query has no sampling aggregators, adding a `merge` aggregator clearly changes the results.
+         */
+        for (final Aggregator aggregator : metric.getAggregators()) {
+            final Optional<Sampling> sampling = aggregator.getSampling();
+            if (!sampling.isPresent()) {
+                // Empirically, non-sampling aggregators commute with `merge`, so we can ignore them
+                continue;
+            }
+            if (!aggregator.getAlignSampling().orElse(false)) {
+                // The first sampling aggregator is not aligned; inserting any aligned aggregator before it will change semantics
```
cwbriones, line 322: nit. `is not sampling aligned` to be precise

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
```diff
@@ -299,6 +297,44 @@ private static MetricsQuery useAvailableRollups(
...
+     * @param metric The metric-query that we might want to rewrite to make use of rollups.
+     * @return The largest {@link SamplingUnit} that the query is insensitive to aggregation over
+     *     (or {@code empty}, if the query is sensitive to all pre-aggregation).
+     */
+    /* package private */ static Optional<SamplingUnit> getMaxUsableRollupUnit(final Metric metric) {
+        /* By querying a rollup (say, the hourly rollup), we're effectively putting a sampling-aligned `merge(1h)`
+         *   in front of the metric's aggregators.
+         * For this to not affect the results, the first aggregator must be sampling-aligned, and must aggregate over
+         *   some integer number of hours.
+         * (Empirically, non-sampling aggregators (e.g. `div`, `filter`) commute with `merge`, so we can ignore them.)
```
cwbriones, line 312: nit. For this comment and all of the below, `sampling aggregator` should be replaced with `range aggregator` to be consistent with the KairosDB terminology

i.e. only range aggregators have a `sampling` argument https://kairosdb.github.io/docs/build/html/restapi/Aggregators.html#range-aggregator

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
```diff
@@ -262,15 +264,11 @@ private static MetricsQuery useAvailableRollups(
                     //
                     // For any given query, we can at best use the rollup based on the smallest
                     // sampling unit, if any.
```
cwbriones, line 266: Can you remove / rewrite this comment block since it's no longer true?

I would opt for just removal since you've already explained what makes a unit "usable" in the helper method itself

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
speezepearson: Correct! (I'm open to other ways of making it unit-testable too, if you have any you'd prefer.)

app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java
speezepearson: Ah, yes, thank you! I did feel fuzzy on that terminology.

PR #354, opened 2020-04-15
add instructions for how to use Postgres init script

speezepearson:


main/postgres/initdb.d/README.markdown
```diff
@@ -0,0 +1,14 @@
+This directory helps you prepare your Postgres instance for MPortal to run.
```
vjkoskela, line 1: To be clear this directory was created to initialize the Docker Postgresql instance from scratch. It was not originally intended as documentation, but does double duty in that sense.

main/postgres/initdb.d/README.markdown
```diff
@@ -0,0 +1,14 @@
...
+
+Context: when MPortal runs for the first time, it will create all the Postgres tables it needs;
+but to do that, it assumes that certain roles/"schemas" are already set up in Postgres.
+
+Here are instructions for how to take a fresh Postgres instance and make it ready for MPortal to run for the first time:
+
+- First, create two users (probably by logging in as the root user via PGAdmin), named `metrics_dba` and `akka_dba`.
+- `export POSTGRES_USER=postgres` (or whatever your root user's name is)
+- `export POSTGRES_DB=????` (TODO(spencerpearson): what is this? Our production Postgres seems to have databases named `postgres` and `rdsadmin`, plus `metrics` which these scripts seem to create.)
+- `export POSTGRES_PASSWORD=<your secret postgres-root-user password>`
```
vjkoskela, line 11: From a security point of view, you should never do `export` on any password.

main/postgres/initdb.d/README.markdown
```diff
@@ -0,0 +1,14 @@
+This directory helps you prepare your Postgres instance for MPortal to run.
+
+Context: when MPortal runs for the first time, it will create all the Postgres tables it needs;
+but to do that, it assumes that certain roles/"schemas" are already set up in Postgres.
+
+Here are instructions for how to take a fresh Postgres instance and make it ready for MPortal to run for the first time:
+
+- First, create two users (probably by logging in as the root user via PGAdmin), named `metrics_dba` and `akka_dba`.
+- `export POSTGRES_USER=postgres` (or whatever your root user's name is)
+- `export POSTGRES_DB=????` (TODO(spencerpearson): what is this? Our production Postgres seems to have databases named `postgres` and `rdsadmin`, plus `metrics` which these scripts seem to create.)
```
vjkoskela, line 10: * `rdsadmin` is created by AWS on your RDS instance
* `postgres` is the default Postgresql database

We could have easily used the default database, but elected not to, and instead create a new one called `metrics` and within it two schemas: `akka` and `portal`. The former being used by Metrics Portal's distributed actor system (when such features are enabled) and the latter as SQL data source for features such as host registry, reports, alerts (soon).

This is not that different from what happens locally (intentionally). The Docker PG instance starts-up and creates an empty `postgres` database as defined by the environment variable set on the docker image in the `pom.xml` file. Then the init script is run which creates the primary root account, and then additional users and roles and the metrics database.

main/postgres/initdb.d/README.markdown
```diff
@@ -0,0 +1,14 @@
...
+- `export POSTGRES_PASSWORD=<your secret postgres-root-user password>`
+- locally `chmod 700 ./init.sh` and edit the `export PGPASSWORD="..."` lines to contain the actual passwords for the metrics_dba and akka_dba users
+- `./init.sh`
```
vjkoskela, line 13: I ran the SQL by hand when I setup the production database so that I could see the results of each step (and considering the cost of failure). I don't know what real DBAs prefer to do, but personally I would not point a script like this at a production database instance!

main/postgres/initdb.d/README.markdown
speezepearson: Mm, yes, I see that this is a careless line for two reasons: the password will appear in your `.bash_history`, and later subprocesses from the same shell could scrape it out of the environment.

If I replaced this with
```
read -s -p 'Password: ' POSTGRES_PASSWORD
(export POSTGRES_PASSWORD; ./init.sh)
```
that would resolve the two concerns I listed.

Is there a threat model I'm missing? I see that, on Linux, a user's processes can inspect each other's environments via `/proc/$PID/environ`, and something similar is probably possible on OS X. But MPortal [already expects to have passwords in its environment](https://github.com/ArpNetworking/metrics-portal/blob/a13645a89655603233444970748bdedec12e8feb/conf/postgresql.application.conf#L32) (and [`init.sh`](https://github.com/ArpNetworking/metrics-portal/blob/a13645a89655603233444970748bdedec12e8feb/main/postgres/initdb.d/init.sh#L19) appears to export passwords as well) -- is that an antipattern that I should avoid copying? If so, why?

BrandonArp: Exporting passwords leaves them in the shell they were exported in.  I think exporting password variables in an interactive shell is a worse idea than doing it in a script where the password is gone at the end of the script's execution.  As far as not putting the password on the command line, that's a good thing. Keeping passwords out of bash history is a good thing.  Ville and I have gone back and forth about putting passwords in env vars at all for exactly the reason you mentioned.  (as well as others like some programs will dump env on errors, include it in debugging, interpolate them in strings, etc)  I'm less concerned about admin scripts like this that are only run on demand (and rarely) than I am for long-running processes like services.

PR #358, opened 2020-04-20
Added 400 response when a metrics query doesn't contain aggregators for all queried metrics.

ddimensia:
This prevents queries from being proxied through to kairos if no aggregators exist on the requested metrics.  This prevents large histogram responses on high-cardinality metrics when someone first sets up a grafana query.  Ideally this should be blocked in the grafana plugin also.

### DIFF
--- a/app/controllers/KairosDbProxyController.java
+++ b/app/controllers/KairosDbProxyController.java
@@ -158,7 +158,7 @@ public class KairosDbProxyController extends Controller {
     public CompletionStage<Result> queryMetrics() {
         try {
             final MetricsQuery metricsQuery = _mapper.treeToValue(request().body().asJson(), MetricsQuery.class);
-            if (! metricsQuery.getMetrics().stream().allMatch(metric -> metric.getAggregators().size() >= 1)) {
+            if (!metricsQuery.getMetrics().stream().allMatch(metric -> metric.getAggregators().size() >= 1)) {
                 return CompletableFuture.completedFuture(
                         Results.badRequest("All queried metrics must have at least one aggregator"));
             }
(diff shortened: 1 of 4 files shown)


app/controllers/KairosDbProxyController.java
```diff
@@ -157,10 +157,15 @@ public KairosDbProxyController(
      */
     public CompletionStage<Result> queryMetrics() {
         try {
-        final MetricsQuery metricsQuery = _mapper.treeToValue(request().body().asJson(), MetricsQuery.class);
-        return _kairosService.queryMetrics(metricsQuery)
-                .<JsonNode>thenApply(_mapper::valueToTree)
-                .thenApply(Results::ok);
+            final MetricsQuery metricsQuery = _mapper.treeToValue(request().body().asJson(), MetricsQuery.class);
+            if (!metricsQuery.getMetrics().stream().allMatch(metric -> metric.getAggregators().size() >= 1)) {
```
cwbriones, line 161: I would invert this condition to say `there is a metric whose aggregator list is empty` i.e.

```java
metricsQuery.getMetrics().stream().anyMatch(metric -> metric.getAggregators().isEmpty())
```
I think it makes intent clearer by removing a `!` and the comparison

BrandonArp:
I'm not a big fan of just 400'ing any request without an aggregator.

ddimensia:
@BrandonArp How about if I make it configurable.

test/java/controllers/KairosDbProxyControllerTest.java
```diff
@@ -0,0 +1,197 @@
...
+                .header("Content-Type", "application/json")
+                .bodyJson(OBJECT_MAPPER.<JsonNode>valueToTree(builder.build()));
+
+        when(_mockKairosDbClient.queryMetrics(any())).thenReturn(
+                CompletableFuture.completedFuture(new MetricsQueryResponse.Builder()
+                        .setQueries(ImmutableList.of()).build())
+        );
+
+        // ***
+        // Test failure case where one metric doesn't have an aggregator
```
cwbriones, line 187: nit. I wouldn't necessarily call this a failure case any more since it'd work fine with this `requiresAggregators` configuration turned off but I follow

PR #359, opened 2020-04-22
add some debug-messages to KairosDB client for querying datapoints

speezepearson:


vjkoskela:
How about a more general solution like a request/reply logging filter?

```
public final class LoggingFilter extends Filter {

    @Override
    public CompletionStage<Result> apply(
            final Function<Http.RequestHeader, CompletionStage<Result>> nextFilter,
            final Http.RequestHeader requestHeader) {
        final long startTime = System.currentTimeMillis();
        return nextFilter.apply(requestHeader).thenApply(result -> {
            final long endTime = System.currentTimeMillis();
            final long elapsedTime = endTime - startTime;

            LOGGER.debug()
                    .setMessage("Request reply")
                    .addData("request.headers", requestHeader.headers())
                    .addData("request.method", requestHeader.method())
                    .addData("request.path", requestHeader.path())
                    .addData("request.query", requestHeader.queryString())
                    .addData("response.header", result.headers())
                    .addData("response.status", result.status())
                    .addData("response.time", elapsedTime)
                    .log();

            return result;
        });
    }

    /**
     * Public constructor.
     *
     * @param materializer Instance of {@code Materializer}.
     */
    @Inject
    public LoggingFilter(final Materializer materializer) {
        super(materializer);
    }

    private static final Logger LOGGER = LoggerFactory.getLogger(LoggingFilter.class);
}
```

Is there a reason that we're logging the time and success/failure instead of emitting these as metrics?

PR #360, opened 2020-04-25
interfaces for internal JobExecution

cwbriones:
Introduce a new type of repository, `JobExecutionRepository`, that pulls out all execution related methods from the `JobRepository` interface.

In the case of Alerts, these two repos will not be backed by the same implementation, and in general that is not a strict requirement and so it makes sense to consider them separately (although in the case of reports, both will still just use the same SQL Ebean impls underneath).

Relatedly, some alert APIs will need to retrieve the state of the most recent execution and so we need an internal model to represent these. We did not have this defined previously because reports are write-only at the moment.

Perhaps somewhat contentiously, I've added another ADT-like hiearchy in `JobExecution` rather than stuff all fields as top-level `Optional`s. This prevents you from, for example, attempting to call `getResult` from an execution that hasn't completed.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/scheduling/JobExecutionRepository.java
+++ b/app/com/arpnetworking/metrics/portal/scheduling/JobExecutionRepository.java
@@ -19,7 +19,6 @@ import models.internal.Organization;
 import models.internal.scheduling.Job;
 import models.internal.scheduling.JobExecution;
 
-import java.io.IOException;
 import java.time.Instant;
 import java.util.NoSuchElementException;
 import java.util.Optional;
@@ -49,10 +48,9 @@ public interface JobExecutionRepository<T> {
      * @param jobId The UUID of the job that completed.
      * @param organization The organization owning the job.
      * @throws NoSuchElementException if no job has the given UUID.
-     * @throws IOException if deserializing the job result fails.
      * @return The last successful execution.
      */
-    Optional<JobExecution.Success<T>> getLastSuccess(UUID jobId, Organization organization) throws NoSuchElementException, IOException;
+    Optional<JobExecution.Success<T>> getLastSuccess(UUID jobId, Organization organization) throws NoSuchElementException;
 
     /**
      * Get the last completed execution, regardless of if it succeeded.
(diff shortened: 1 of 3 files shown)


app/models/internal/scheduling/JobExecution.java
```diff
@@ -0,0 +1,435 @@
...
+         */
+        protected abstract B self();
+    }
+
+    /**
+     * An execution that has been started but not yet completed.
+     *
+     * @param <T> The result type of this execution.
+     */
+    public static final class Started<T> extends JobExecution<T> {
```
speezepearson, line 166: Does this need to be generic?

app/com/arpnetworking/metrics/portal/scheduling/JobExecutionRepository.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Notify the repository that a job finished executing successfully.
+     *
+     * @param jobId The UUID of the job that completed.
+     * @param organization The organization owning the job.
+     * @param scheduled The time that the completed job-run was scheduled for.
+     * @param result The result that the job computed.
+     */
+    void jobSucceeded(UUID jobId, Organization organization, Instant scheduled, T result);
```
speezepearson, line 82: If you wanted to get _crazy_, you could make this take a `JobExecution.Started<T>` as proof that the job has already started.

Recent unpleasant experiences have mellowed me on the whole "use the typechecker to verify your program's correctness" thing, though, and I'll happily admit that that's _probably_ going overboard.

app/models/internal/scheduling/JobExecution.java
```diff
@@ -0,0 +1,435 @@
...
+            }
+        }
+    }
+
+    /**
+     * An execution that completed with an error result.
+     *
+     * @param <T> The result type of this execution.
+     */
+    public static final class Failure<T> extends JobExecution<T> {
```
speezepearson, line 331: Does _this_ need to be generic?

app/models/internal/scheduling/JobExecution.java
cwbriones: Yep - the type parameter is essentially a phantom type here. It's needed to be able to uniformly handle each variant of `JobExecution<T>` for a particular `T`. 

Otherwise you could assign a `JobExecution.Started<Integer>` to a `JobExecution<String>` which doesn't seem correct regardless of if the object actually contains a `T` or not.

A more pragmatic reason is that even if I wanted to leave it out the compiler would complain that the type variable `T` is not defined anywhere to be specified in the `extends` clause.

app/models/internal/scheduling/JobExecution.java
cwbriones: same as above, yup.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutionRepository.java
cwbriones: I agree - while that would be nice it would put more of a burden on the user of the API than I'd like, especially since this is all on top of the database so we'd essentially be _requiring_ an initial lookup if you started with just a Job ID.

It seems easier / more idiomatic to handle that precondition within the repository implementation and throw if its violated.

PR #361, opened 2020-04-25
Implement JobExecutionRepository for Reports

cwbriones:
This pulls out all relevant execution logic from `DatabaseReportRepository` into `DatabaseReportExecutionRepository`.

Depends on #360.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
+++ b/app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
@@ -1,0 +1,297 @@
+/*
+ * Copyright 2018 Dropbox, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.arpnetworking.metrics.portal.reports.impl;
+
+import com.arpnetworking.metrics.portal.scheduling.JobExecutionRepository;
+import com.arpnetworking.steno.Logger;
+import com.arpnetworking.steno.LoggerFactory;
+import io.ebean.EbeanServer;
+import io.ebean.Transaction;
+import models.ebean.ReportExecution;
+import models.internal.Organization;
+import models.internal.reports.Report;
+import models.internal.scheduling.JobExecution;
+
+import java.time.Instant;
+import java.util.NoSuchElementException;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.persistence.EntityNotFoundException;
+import javax.persistence.PersistenceException;
+
+/**
+ * Implementation of {@link JobExecutionRepository} for {@link Report} jobs using a SQL database.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class DatabaseReportExecutionRepository implements JobExecutionRepository<Report.Result> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DatabaseReportExecutionRepository.class);
+
+    private final AtomicBoolean _isOpen = new AtomicBoolean(false);
+    private final EbeanServer _ebeanServer;
+
+    /**
+     * Public constructor.
+     *
+     * @param ebeanServer Play's {@code EbeanServer} for this repository.
+     */
+    @Inject
+    public DatabaseReportExecutionRepository(@Named("metrics_portal") final EbeanServer ebeanServer) {
+        _ebeanServer = ebeanServer;
+    }
+
+    @Override
+    public void open() {
+        assertIsOpen(false);
+        LOGGER.debug().setMessage("Opening DatabaseReportExecutionRepository").log();
+        _isOpen.set(true);
+    }
+
+    @Override
+    public void close() {
+        assertIsOpen();
+        LOGGER.debug().setMessage("Closing DatabaseReportExecutionRepository").log();
+        _isOpen.set(false);
+    }
+
+
+    /**
+     * Get the most recently scheduled execution, if any.
+     *
+     * This could possibly return an execution that's pending completion.
+     *
+     * @param jobId The UUID of the job that completed.
+     * @param organization The organization owning the job.
+     * @throws NoSuchElementException if no job has the given UUID.
+     * @return The last successful execution.
+     */
+    public Optional<JobExecution<Report.Result>> getLastScheduled(final UUID jobId, final Organization organization) throws NoSuchElementException {
+        assertIsOpen();
+        return _ebeanServer.find(ReportExecution.class)
+                .orderBy()
+                .desc("scheduled")
+                .where()
+                .eq("report.uuid", jobId)
+                .eq("report.organization.uuid", organization.getId())
+                .setMaxRows(1)
+                .findOneOrEmpty()
+                .map(this::toInternalModel);
+    }
+
+    @Override
+    public Optional<JobExecution.Success<Report.Result>> getLastSuccess(final UUID jobId, final Organization organization) throws NoSuchElementException {
+        assertIsOpen();
+        final Optional<ReportExecution> row = _ebeanServer.find(ReportExecution.class)
+                .orderBy()
+                .desc("completed_at")
+                .where()
+                .eq("report.uuid", jobId)
+                .eq("report.organization.uuid", organization.getId())
+                .eq("state", ReportExecution.State.SUCCESS)
+                .setMaxRows(1)
+                .findOneOrEmpty();
+        if (row.isPresent()) {
+            final JobExecution<Report.Result> execution = toInternalModel(row.get());
+            if (execution instanceof JobExecution.Success) {
+                return Optional.of((JobExecution.Success<Report.Result>) execution);
+            }
+            throw new IllegalStateException("execution returned was not a success");
+        }
+        return Optional.empty();
+    }
+
+    @Override
+    public Optional<JobExecution<Report.Result>> getLastCompleted(final UUID jobId, final Organization organization) throws NoSuchElementException {
+        assertIsOpen();
+        return _ebeanServer.find(ReportExecution.class)
+                .orderBy()
+                .desc("completed_at")
+                .where()
+                .eq("report.uuid", jobId)
+                .eq("report.organization.uuid", organization.getId())
+                .in("state", ReportExecution.State.SUCCESS, ReportExecution.State.FAILURE)
+                .setMaxRows(1)
+                .findOneOrEmpty()
+                .map(this::toInternalModel);
+    }
+
+    @Override
+    public void jobStarted(final UUID reportId, final Organization organization, final Instant scheduled) {
+        assertIsOpen();
+        updateExecutionState(
+                reportId,
+                organization,
+                scheduled,
+                ReportExecution.State.STARTED,
+                null,
+                null
+        );
+    }
+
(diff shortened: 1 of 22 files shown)


app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
```diff
@@ -0,0 +1,311 @@
+/*
+ * Copyright 2018 Dropbox, Inc.
```
cwbriones, line 2: Just caught that this says 2018, will fix.

app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -133,10 +153,12 @@ public void preStart() throws Exception {
                 .build();
     }
 
+    // CHECKSTYLE.OFF: ParameterNumber - This is only called in one location.
```
cwbriones, line 156: @speezepearson is there a reason this whole method is static? If this were not static, I could refactor all of this by just passing a method reference `this::runAntiEntropy` as the runnable rather than the returned value from `createAntiEntopyRunnable`

I'm guessing it might have to do with the actors possibly running on different hosts?

speezepearson: I think you can just make this an instance method. I believe I was under the mistaken impression that a periodic timer's messages would pile up, if the actor took more than 1 period to action them.

app/models/ebean/ReportExecution.java
```diff
@@ -37,7 +38,7 @@
 
 /**
  * An execution event for a {@link Report}.
- *
+ * <p>
```
speezepearson, line 41: nit: missing `</p>`?

app/models/ebean/ReportExecution.java
```diff
@@ -76,7 +74,7 @@
     @Nullable
     @DbJsonB
     @Column(name = "error")
-    private Map<String, String> error;
+    private Error error;
```
speezepearson, line 77: Is this data-type change... legal? Don't we need a migration, at least?
(I see your new Error classes below, but I don't think I understand how this all fits together. Help?)

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
```diff
@@ -0,0 +1,311 @@
...
+        final ReportExecution.State state = beanModel.getState();
+        switch (state) {
+            case STARTED:
+                return new JobExecution.Started.Builder<Report.Result>()
+                        .setJobId(beanModel.getReport().getUuid())
+                        .setScheduled(beanModel.getScheduled())
+                        .setStartedAt(beanModel.getStartedAt())
+                        .build();
+            case FAILURE:
+                @Nullable final Throwable throwable = beanModel.getError() == null ? null : beanModel.getError().getThrowable();
```
vjkoskela, line 280: Above it seems that error is required (when transitioning to failure), but here it seems optional. Did I miss something?

vjkoskela: (is this constraint represented in the model?)

app/models/ebean/Report.java
```diff
@@ -194,6 +196,24 @@ public void setOrganization(final Organization value) {
...
+     */
+    public static Optional<Report> findByUUID(
+            final EbeanServer ebeanServer,
+            final models.ebean.Organization organization,
+            final UUID reportId) {
+        return ebeanServer.find(models.ebean.Report.class)
+                .where()
+                .eq("uuid", reportId)
+                .eq("organization.uuid", organization.getUuid())
+                .findOneOrEmpty();
```
vjkoskela, line 215: note: here you don't do the max rows.

app/models/ebean/ReportExecution.java
vjkoskela: This isn't referring to:

https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html

Is it?

I would also like to know how this is backwards compatible and what motivated the change.

app/models/ebean/ReportExecution.java
```diff
@@ -208,5 +202,61 @@ public int hashCode() {
...
+     */
+    @JsonTypeInfo(
+            use = JsonTypeInfo.Id.NAME,
+            defaultImpl = ErrorString.class,
+            property = "type"
+    )
+    @JsonSubTypes({
+            @JsonSubTypes.Type(value = ErrorString.class, name = "ERROR_STRING"),
+    })
+    public abstract static class Error {
```
vjkoskela, line 217: Oh. No.

https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html

This would be a major source of confusion based on the name.

app/models/ebean/ReportExecution.java
cwbriones: This is backwards compatible and I included a test for it. I changed it because while revisiting the existing structure I wanted something more structured than `Map<String, String>`

app/models/ebean/ReportExecution.java
cwbriones: I thought about this too and I was going to name it something like `ExecutionError` but then I figured that would be redundant because this is an inner class.

As in, I didn't think it bought much to have a different name because it will already be used as `ReportExecution.Error`.

app/models/ebean/ReportExecution.java
cwbriones: The javadoc compiler automatically inserts closing paragraph tags

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
```diff
@@ -0,0 +1,311 @@
...
+    public Optional<JobExecution<Report.Result>> getLastScheduled(final UUID jobId, final Organization organization)
+            throws NoSuchElementException {
+        assertIsOpen();
+        return _ebeanServer.find(ReportExecution.class)
+                .orderBy()
+                .desc("scheduled")
+                .where()
+                .eq("report.uuid", jobId)
+                .eq("report.organization.uuid", organization.getId())
+                .setMaxRows(1)
```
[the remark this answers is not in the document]
cwbriones: Yes but I'm not sure if there's an alternative here. I want the most recently scheduled and since there are many executions for (most) jobs then removing the maxRows would cause it to throw most of the time.

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
```diff
@@ -0,0 +1,311 @@
...
+                        reportId.toString(),
+                        organization.getId()
+                );
+                throw new EntityNotFoundException(message);
+            }
+
+            final Optional<ReportExecution> existingExecution =
+                    _ebeanServer.createQuery(ReportExecution.class)
+                            .where()
+                            .eq("report.uuid", reportId)
```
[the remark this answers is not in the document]
cwbriones: I changed this to instead take the `report` object itself since I have one; it will then match the organization and reportId together

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
```diff
@@ -0,0 +1,311 @@
...
+            switch (state) {
+                case STARTED:
+                    newOrUpdatedExecution.setStartedAt(Instant.now());
+                    newOrUpdatedExecution.setCompletedAt(null);
+                    break;
+                case FAILURE:
+                    if (error == null) {
+                        throw new IllegalArgumentException("Error is null for a failed execution.");
+                    }
+                    newOrUpdatedExecution.setError(new ReportExecution.ErrorString(Throwables.getStackTraceAsString(error)));
```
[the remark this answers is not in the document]
cwbriones: this includes the message already

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
```diff
@@ -0,0 +1,311 @@
...
+        } catch (final RuntimeException e) {
+            // CHECKSTYLE.ON: IllegalCatchCheck
+            LOGGER.error()
+                    .setMessage("Failed to update report executions")
+                    .addData("report.uuid", reportId)
+                    .addData("scheduled", scheduled)
+                    .addData("state", state)
+                    .setThrowable(e)
+                    .log();
+            throw new PersistenceException("Failed to update report executions", e);
```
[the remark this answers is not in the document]
cwbriones: At that point should I just pull this out into its own exception type? My reasoning is that by including the error details in the log and re-thrown exception, I'm essentially handling the error twice (log + rethrow).

If I pull it out I can then just re-throw and rely on the outer handler to log this information.

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
cwbriones: It's nullable here because it is nullable in the data model. The constraint is validated in the internal model's builder.

app/models/ebean/Report.java
cwbriones: The difference is here I don't expect there to be more than one row with a given report/organization id, so it should throw.

The above code only specifies the report, but a report can have multiple executions. Since we want the latest one, we add the limit because otherwise we'd be grabbing them all (and triggering an exception in `findOneOrEmpty`.

app/models/ebean/ReportExecution.java
vjkoskela: I *believe* the original point of this was to be that unstructured so that it could store more than just an exception. For example, we could store progress, messages from dependencies, etc..

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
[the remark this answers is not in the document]
vjkoskela: You could set it to 2.

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
vjkoskela: Yikes. We have a stricter internal model than data model? I guess it's not dangerous, just odd.

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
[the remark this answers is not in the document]
vjkoskela: Sorry, missed that. We should never log and throw. One or the other; if you throw let the caller decide.

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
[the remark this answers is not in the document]
vjkoskela: Oh! Nevermind.

app/models/ebean/ReportExecution.java
vjkoskela: Please add a comment to `ReportExecution.Errror` to rename the class.

app/models/ebean/ReportExecution.java
cwbriones: I figure that with this change you'd still be able to throw in new fields in a new structure that would be identified by the `JsonTypeInfo`. Yes it would not be as flexible for throwing in these other fields but we could still handle those other cases with additional metadata by adding a new class to hold them.

I suppose it's all trade-offs. If there's no obvious benefit here then I can just revert this.

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportExecutionRepository.java
cwbriones: I don't think it would be safe to mark the fields as NonNull anyway since they're essentially coming from the outside world. I consider it defensive at an IO boundary. There's nothing stopping someone from populating a row with a nonsensical state from the command line, for example.

app/models/ebean/ReportExecution.java
cwbriones: I'll just remove/revert it. There's not as obvious as a gain as I had thought from moving this to a class instead of a Map, and removing it avoids these problems entirely.

PR #362, opened 2020-04-28
fix KairosDB query-logging to actually include the query

speezepearson:
Log line before (gotten by tailing the logs after `mvn docker:start` with a tweaked Logback config to enable debug-level logging, then hitting the container with a KairosDB query request):
```
{
    ...
    "query": {
      "_id": "356c4af6",
      "_class": "com.arpnetworking.kairos.client.models.MetricsQuery",
      "cache_time": 0
    }
    ...
}
```

Log line after:
```
{
    ...
    "query": {
      "metrics": [
        {
          "aggregators": [
            {
              "sampling": {
                "unit": "minutes",
                "value": 1
              },
              "name": "percentile",
              "align_sampling": true,
              "align_start_time": true,
              "percentile": "0.999"
            }
          ],
          "limit": 0,
          "tags": {
            "build_channel": [
              "beta",
              "stable"
            ]
          },
          "name": "some_metric/foo/bar",
          "group_by": [
            {
              "name": "tag",
              "tags": [
                "platform",
                "build_channel"
              ]
            }
          ]
        }
      ],
      "start_absolute": 1586772000000,
      "end_absolute": 1586772000000,
      "cache_time": 0
    }
    ...
}
```

### DIFF
--- a/app/com/arpnetworking/kairos/client/KairosDbClientImpl.java
+++ b/app/com/arpnetworking/kairos/client/KairosDbClientImpl.java
@@ -39,6 +39,7 @@ import com.arpnetworking.steno.LoggerFactory;
 import com.fasterxml.jackson.annotation.JacksonInject;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.type.TypeFactory;
 import com.google.common.collect.ImmutableList;
@@ -48,6 +49,8 @@ import scala.concurrent.duration.FiniteDuration;
 
 import java.io.IOException;
 import java.net.URI;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.UUID;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


cwbriones:
Can this also be resolved by adding a `@Loggable` to the MetricsQuery class? Then you wouldn't be serializing at the callsite.

speezepearson:
> Can this also be resolved by adding a `@Loggable` to the MetricsQuery class? Then you wouldn't be serializing at the callsite.

Considered that. Certainly `@Loggable` is more idiomatic in this codebase. What sent me down the other path was:
- It wouldn't just be making `MetricsQuery` `@Loggable`, but also `Metric`, `Sampling`, `RelativeDateTime`, and `Aggregator`; and for consistency, probably everything else in the package. And if we added a new class we'd have to remember to make it `@Loggable` too, or it would end up opaque in the logs.
- The query is _already_ getting turned into JSON; by logging that JSON, we guarantee that we log an exact and complete description of the request getting sent to KairosDB.

I don't feel strongly, though, and if you think `@Loggable` is the way to go I'll update the PR accordingly.

vjkoskela:
So in general, all model classes should be marked @Loggable as long as they contain data that is safe to log; e.g. no credentials, ssns, credit cards, etc.

speezepearson:
> So in general, all model classes should be marked @loggable as long as they contain data that is safe to log; e.g. no credentials, ssns, credit cards, etc.

Okay! I applied that fix in #363.

app/com/arpnetworking/kairos/client/KairosDbClientImpl.java
```diff
@@ -61,26 +64,25 @@
...
+                .withEntity(ContentTypes.APPLICATION_JSON, queryJson.toString());
+        final Instant startTime = Instant.now();
+        return fireRequest(request, MetricsQueryResponse.class)
+                .whenComplete((response, error) -> {
+                    LOGGER.trace()
+                            .setMessage("finished queryMetrics")
+                            .addData("queryUuid", queryUuid)
+                            .addData("query", queryJson)
+                            .addData("duration", Duration.between(startTime, Instant.now()))
+                            .addData("success", error == null)
```
vjkoskela, line 83: Sorry, shouldn't you call `setThrowable` with `error` so the actual error is output as well?

app/com/arpnetworking/kairos/client/KairosDbClientImpl.java
```diff
@@ -61,26 +64,25 @@
...
+        final HttpRequest request = HttpRequest.POST(createUri(METRICS_QUERY_PATH).toString())
+                .withEntity(ContentTypes.APPLICATION_JSON, queryJson.toString());
+        final Instant startTime = Instant.now();
+        return fireRequest(request, MetricsQueryResponse.class)
+                .whenComplete((response, error) -> {
+                    LOGGER.trace()
+                            .setMessage("finished queryMetrics")
+                            .addData("queryUuid", queryUuid)
+                            .addData("query", queryJson)
+                            .addData("duration", Duration.between(startTime, Instant.now()))
```
vjkoskela, line 82: No problem logging it, but it seems like it should be emitted to metrics as well.

app/com/arpnetworking/kairos/client/KairosDbClientImpl.java
speezepearson: Ah, forgot about that method! Added.

app/com/arpnetworking/kairos/client/KairosDbClientImpl.java
speezepearson: Added!

PR #364, opened 2020-04-28
Rollups: add filter-by-tag field(s) to RollupDefinition

speezepearson:
This consists of:

- Adding the `_filterTags` field to tell the RollupExecutors which tags to slice on. This field isn't used yet, it's just left at its default empty value, but that will change.

- Generalizing `_groupByTags` (used to populate the rollup-query's `group_by`) to include tag _values_ as well as names. (new name `_allMetricTags`)

Together, these will facilitate partition-and-retry logic: when a job fails, we can pick a tag from `_allMetricTags` that we haven't yet filtered by, and add various values to `_filterTags` for various sub-jobs.

### DIFF
--- a/app/com/arpnetworking/rollups/RollupDefinition.java
+++ b/app/com/arpnetworking/rollups/RollupDefinition.java
@@ -176,9 +176,9 @@ public final class RollupDefinition implements Serializable, ConsistentHashingRo
         }
 
         /**
-         * Sets the {@code _allMetricTags} and returns a reference to this Builder so that the methods can be chained together.
+         * Sets the {@code _filterTags} and returns a reference to this Builder so that the methods can be chained together.
          *
-         * @param value the {@code _allMetricTags} to set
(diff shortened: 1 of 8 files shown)


app/com/arpnetworking/rollups/RollupDefinition.java
```diff
@@ -164,13 +176,24 @@ public Builder setStartTime(final Instant value) {
         }
 
         /**
-         * Sets the {@code _groupByTags} and returns a reference to this Builder so that the methods can be chained together.
+         * Sets the {@code _allMetricTags} and returns a reference to this Builder so that the methods can be chained together.
+         *
+         * @param value the {@code _allMetricTags} to set
```
orborde, line 181: Should these docstrings say `_filterTags` instead of `_allMetricTags`?

speezepearson: Quite right, thanks!

PR #365, opened 2020-04-29
use ThreadLocalBuilder where possible

speezepearson:
This should make our logs somewhat quieter.

### DIFF
--- a/app/com/arpnetworking/rollups/RollupGenerator.java
+++ b/app/com/arpnetworking/rollups/RollupGenerator.java
@@ -19,6 +19,7 @@ import akka.actor.AbstractActorWithTimers;
 import akka.actor.ActorRef;
 import akka.japi.pf.ReceiveBuilder;
 import akka.pattern.Patterns;
+import com.arpnetworking.commons.builder.ThreadLocalBuilder;
 import com.arpnetworking.kairos.client.KairosDbClient;
 import com.arpnetworking.kairos.client.models.Aggregator;
 import com.arpnetworking.kairos.client.models.DataPoint;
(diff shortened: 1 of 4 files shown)


app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -376,13 +376,16 @@ private LastDataPointsMessage buildLastDataPointResponse(
             final ImmutableMultimap<String, String> tags,
             final MetricsQueryResponse response,
             @Nullable final Throwable failure) {
-        final LastDataPointsMessage.Builder builder = new LastDataPointsMessage.Builder()
-                .setSourceMetricName(sourceMetricName)
-                .setRollupMetricName(rollupMetricName)
-                .setPeriod(period)
-                .setTags(tags);
+        LastDataPointsMessage result = ThreadLocalBuilder.build(
```
cwbriones, line 379: Could you lift this builder into a parameter?
Then at the callsite for `buildLastDataPointResponse` you can use `ThreadLocalBuilder.build` like so:
```java
ThreadLocalBuilder.build(LastDataPointMessage.Builder.class, builder ->
    buildLastDataPointResponse(
        builder,
        sourceMetricName,
        rollupMetricName,
        period,
        tags,
        response,
        failure)
)
```
and none of the method body needs to change other than the builder instantiation. It makes me sad to see the `Optional` chaining turned into `isPresent` 😢

vjkoskela:
If a Builder implements (extends) ThreadLocalBuilder it is required to adhere to an additional contract which is to provide a reset method which must set the state of the builder back to what it was when it was instantiated.

Bad bugs can result if there are issues with the reset method. I have a long-standing task to create a reflective test that verifies the behavior of all reset methods. Alternatively, we were considering code weaving to generate the reset methods at compile time based on the initializations of the fields; however, there are some complexities.

I have not looked at the code, but thank you for doing this! Let me know if anyone is interested in more details about TLB.

PR #366, opened 2020-04-29
Make rollups split and retry on failure

speezepearson:
Now, when a RollupExecutor fails, it reports that failure to the RollupManager; if the failure looks like a timeout or something[1], then the RollupManager splits the failed job into thinner slices by filtering on a new tag, and re-enqueues those slices for other executors to pick up.

[1] "looks like a timeout or something" is an ad-hoc check that I'm not thrilled with. Empirically, it tends to 502, but it also sometimes 500s.
```
$ zgrep -h 'Failed to execute rollup query' /REDACTED-*.gz | jq -c '{"message": (.data.message), "cause": (.exception.data.cause | {type, status: (try .data.httpStatus), message})}' | sort | uniq -c | sort -n
      1 {"message":"Failed to execute rollup query.","cause":{"type":"com.arpnetworking.kairos.client.KairosDbRequestException","status":500,"message":"{\"errors\":[\"org.kairosdb.core.exception.DatastoreException: com.datastax.driver.core.exceptions.OperationTimedOutException: [/REDACTED:9042] Timed out waiting for server response\"]}"}}
      1 {"message":"Failed to execute rollup query.","cause":{"type":"com.arpnetworking.kairos.client.KairosDbRequestException","status":500,"message":"{\"errors\":[\"org.kairosdb.core.exception.DatastoreException: com.datastax.driver.core.exceptions.OperationTimedOutException: [/REDACTED:9042] Timed out waiting for server response\"]}"}}
      1 {"message":"Failed to execute rollup query.","cause":{"type":"com.arpnetworking.kairos.client.KairosDbRequestException","status":500,"message":"{\"errors\":[\"org.kairosdb.core.exception.DatastoreException: Failed to read key index\"]}"}}
      1 {"message":"Failed to execute rollup query.","cause":{"type":"com.arpnetworking.kairos.client.KairosDbRequestException","status":500,"message":"{\"errors\":[\"org.kairosdb.core.exception.DatastoreException: java.util.concurrent.CancellationException: Task was cancelled.\"]}"}}
      2 {"message":"Failed to execute rollup query.","cause":{"type":"com.arpnetworking.kairos.client.KairosDbRequestException","status":500,"message":"{\"errors\":[\"org.kairosdb.core.exception.DatastoreException: com.datastax.driver.core.exceptions.ReadTimeoutException: Cassandra timeout during read query at consistency ONE (1 responses were required but only 0 replica responded)\"]}"}}
      9 {"message":"Failed to execute rollup query.","cause":{"type":"com.arpnetworking.kairos.client.KairosDbRequestException","status":500,"message":"{\"errors\":[\"org.kairosdb.core.exception.DatastoreException: com.datastax.driver.core.exceptions.TransportException: [/REDACTED:9042] Connection has been closed\"]}"}}
    103 {"message":"Failed to execute rollup query.","cause":{"type":"com.arpnetworking.kairos.client.KairosDbRequestException","status":502,"message":"<html>\r\n<head><title>502 Bad Gateway</title></head>\r\n<body>REDACTED</body>\r\n</html>\r\n"}}
```

### DIFF
--- a/app/com/arpnetworking/rollups/RollupManager.java
+++ b/app/com/arpnetworking/rollups/RollupManager.java
@@ -103,9 +103,9 @@ public class RollupManager extends AbstractActorWithTimers {
             return;
         }
 
-        final ImmutableSet<RollupDefinition> subjobs;
+        final ImmutableSet<RollupDefinition> children;
         try {
-            subjobs = RollupPartitioningUtils.splitJob(message.getRollupDefinition());
+            children = RollupPartitioningUtils.splitJob(message.getRollupDefinition());
             _periodicMetrics.recordCounter("rollup/manager/executor_finished/unsplittable", 0);
         } catch (final RollupPartitioningUtils.CannotSplitException e) {
             _periodicMetrics.recordCounter("rollup/manager/executor_finished/unsplittable", 1);
@@ -117,7 +117,13 @@ public class RollupManager extends AbstractActorWithTimers {
             return;
         }
 
-        subjobs.forEach(subjob -> getSelf().tell(subjob, getSelf()));
+        LOGGER.info()
+                .setMessage("splitting and retrying job")
+                .addData("parent", message.getRollupDefinition())
+                .addData("children", children)
+                .setThrowable(failure.get())
+                .log();
+        children.forEach(child -> getSelf().tell(child, getSelf()));
     }
 
     private Optional<RollupDefinition> getNextRollup() {
--- a/app/com/arpnetworking/rollups/RollupManager.java
+++ b/app/com/arpnetworking/rollups/RollupManager.java
@@ -63,6 +63,10 @@ public class RollupManager extends AbstractActorWithTimers {
                         RollupDefinition.class,
                         work -> _rollupDefinitions.add(work))
                 .match(
+                        RollupExecutor.FinishRollupMessage.class,
+                        this::rollupFinished
+                )
+                .match(
                         RollupFetch.class,
                         work -> {
                             _periodicMetrics.recordCounter("rollup/manager/fetch", 1);
@@ -75,6 +79,9 @@ public class RollupManager extends AbstractActorWithTimers {
 
                         })
                 .build();
+    }
+
+    private void rollupFinished(final RollupExecutor.FinishRollupMessage message) {
     }
 
     private Optional<RollupDefinition> getNextRollup() {
--- a/app/com/arpnetworking/rollups/RollupManager.java
+++ b/app/com/arpnetworking/rollups/RollupManager.java
@@ -52,6 +52,8 @@ public class RollupManager extends AbstractActorWithTimers {
      * Metrics discovery constructor.
      *
      * @param periodicMetrics periodic metrics client
+     * @param metricsFactory metrics factory
+     * @param partitioner {@link RollupPartitioner} to split up failed jobs
      */
     @Inject
     public RollupManager(final PeriodicMetrics periodicMetrics, final MetricsFactory metricsFactory, final RollupPartitioner partitioner) {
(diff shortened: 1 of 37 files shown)


app/com/arpnetworking/rollups/RollupManager.java
```diff
@@ -77,6 +85,41 @@ public Receive createReceive() {
...
+        _periodicMetrics.recordCounter("rollup/manager/executor_finished/unretriable", isRetryable ? 0 : 1);
+        if (!isRetryable) {
+            LOGGER.warn()
+                    .setMessage("giving up after non-retryable error")
+                    .addData("rollupDefinition", message.getRollupDefinition())
+                    .setThrowable(failure.get())
+                    .log();
+            return;
+        }
+
```
ddimensia, line 105: Can we at a minimum log which metrics we're retrying/splitting.  Ideally we'd have a metric by metricname but if we're concerned about cardinality then I'm okay with just logging.  The idea is that this may help inform us on where we would want to look if we switched to pre-splitting via configuration or some other mechanism.

app/com/arpnetworking/rollups/RollupPartitioningUtils.java
```diff
@@ -0,0 +1,112 @@
...
+     * @return a set of cheaper jobs which partition the parent's work among them
+     * @throws CannotSplitException if there is no way to split the given job into multiple cheaper ones
+     */
+    public static ImmutableSet<RollupDefinition> splitJob(final RollupDefinition job) throws CannotSplitException {
+        final ImmutableMap<String, String> filterTags = job.getFilterTags();
+        final ImmutableMultimap<String, String> allTags = job.getAllMetricTags();
+        final Optional<String> nextFilterTag = allTags.keySet().stream()
+                .filter(tag -> !filterTags.containsKey(tag)) // already filtered on
+                .filter(tag -> allTags.get(tag).size() > 1) // no point splitting on a single-value tag
+                .max(Comparator.comparing(
```
ddimensia, line 53: I'm okay with this as a first pass but I think we may want to change to a prioritized list of system tags as the first pass for splitting.  Just something to think about.  The reason I'm unsure about this in the first place is that I know the _os_version tag is likely to surface here for a period of time and it will result in many queries when something like _region might be sufficient to get a meaningful split.

ddimensia:
I guess one other thing that may be useful is a max time for retrying, more specifically, I shouldn't keep splitting and retrying 24 hours later for a daily rollup.  I don't really think that's likely but might be a reasonable safety valve.

app/com/arpnetworking/rollups/RollupPartitioningUtils.java
speezepearson: Yeah, I share your concerns. A couple alternatives that come to mind:
- "tag whose number of values is closest to 10"
- ["tag whose log-number of values is closest to log-10"](https://www.desmos.com/calculator/vpr78hofys)

...but I thought those were probably too fancy.

My impression is that `_os_version`, like `browser_name` for web-perf metrics, has O(100) values, almost all of which are negligible. So, if we split on `_os_version`, we'll fire off O(100) queries that finish ~immediately (because there's little to no data), plus a handful of expensive queries.

I guess that might get expensive: it looks like querying a big row-key table, which each query needs to do, [takes O(0.1s)](https://grafana.pp.dropbox.com/d/000001001/spencers-playground?tab=queries&panelId=43&edit&fullscreen&orgId=1&from=now-2d&to=now).

So, yeah, I'll change this to "log-number of values closest to log-10."

speezepearson: (I'm marking this resolved, but happy to change it if you agree with the "too fancy" criticism and see a better way)

ddimensia:
Not sure of the state of this PR but it looks like ci is still unhappy and the branch needs updating.

speezepearson:
> I guess one other thing that may be useful is a max time for retrying, more specifically, I shouldn't keep splitting and retrying 24 hours later for a daily rollup. I don't really think that's likely but might be a reasonable safety valve.

That seems reasonable! Mind if I do that as a fast-follow?

PR #368, opened 2020-05-01
Cleanup anti-entropy, generate standard methods for JobExecution

cwbriones:
* Remove static methods for createAntiEntropyRunnable and just use a method reference instead.
* Generate `hashCode`, `equals`, and `toString` for all JobExecution variants.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
+++ b/app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
@@ -153,37 +153,28 @@ public final class JobCoordinator<T> extends AbstractPersistentActorWithTimers {
                 .build();
     }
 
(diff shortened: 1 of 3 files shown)


app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -256,7 +225,7 @@ public Receive createReceive() {
 
                     getContext().getSystem().scheduler().scheduleOnce(
                             scala.concurrent.duration.Duration.Zero(),
-                            createAntiEntropyRunnable(),
+                            this::runAntiEntropy,
                             getContext().dispatcher());
```
speezepearson, line 229: I think this could whole block even be reduced to `runAntiEntropy()`. I think there's no reason for this to be spun off into the background, given that the `_currentlyExecuting` flag ensures that only one of these anti-entropy scans can be running at a time.

PR #370, opened 2020-05-05
Rollups: add latency metrics to RollupManager

speezepearson:


app/com/arpnetworking/rollups/RollupManager.java
```diff
@@ -94,9 +96,18 @@ public Receive createReceive() {
...
+        final RollupDefinition definition = message.getRollupDefinition();
+        final double latencyNs = (double) Duration.between(definition.getStartTime(), Instant.now()).toNanos();
+
         try (Metrics metrics = _metricsFactory.create()) {
-            metrics.addAnnotation("rollup_metric", message.getRollupDefinition().getDestinationMetricName());
+            metrics.addAnnotation("rollup_metric", definition.getDestinationMetricName());
             metrics.incrementCounter("rollup/manager/executor_finished", 1);
+            metrics.setGauge("rollup/manager/executor_finished/latency_sec", latencyNs / 1e9);
+            metrics.setGauge(
+                    "rollup/manager/executor_finished/latency_periods",
```
orborde, line 107: What is this metric, semantically?

speezepearson: (typed up a response, realized that this metric could be improved; updating...)

speezepearson: No longer useful!

app/com/arpnetworking/rollups/RollupManager.java
```diff
@@ -94,9 +96,13 @@ public Receive createReceive() {
...
 
     private void executorFinished(final RollupExecutor.FinishRollupMessage message) {
+        final RollupDefinition definition = message.getRollupDefinition();
+        final double latencyNs = (double) Duration.between(definition.getEndTime(), Instant.now()).toNanos();
+
         try (Metrics metrics = _metricsFactory.create()) {
-            metrics.addAnnotation("rollup_metric", message.getRollupDefinition().getDestinationMetricName());
+            metrics.addAnnotation("rollup_metric", definition.getDestinationMetricName());
             metrics.incrementCounter("rollup/manager/executor_finished", 1);
+            metrics.setGauge("rollup/manager/executor_finished/latency_sec", latencyNs / 1e9);
```
orborde, line 105: Wouldn't you want to record a value into a histogram here? Setting a gauge seems weird - the only data you'll ever record is the most recent execution.

(maybe I have the wrong mental model of how these gauge metrics work?)

speezepearson: I seem to recall that counter/timer/gauge metrics are all treated the same under the hood. Which one you use merely hints at which aggregation operations are meaningful to perform on it (e.g. "sum" on a gauge is meaningless).

(@vjkoskela, tell me if I'm wrong?)

vjkoskela: I can't. You are correct.

Gauge/Counter/Timer is mainly a client-side construct; how to measure something (e.g. timers are measured with stopwatches, counters with incrementing/decrementing, and gauges are just direct setting of an observed sample).

The system still supports which aggregators are applied by statistic type, although we are moving away from this (and don't use it at DBX) where all data is represented as an augmented histogram (which provides all numerical statistics).

PR #373, opened 2020-05-05
Enable DatabaseReportExecutionRepository for postgres config

cwbriones:
It's currently not bound to anything by default so the injector throws from within CachedJob as it tries to fetch the repository impl.


speezepearson:
Ah, and there's no test that would have caught this because there isn't yet any kind of Job that runs in the integration tests?

cwbriones:
> Ah, and there's no test that would have caught this because there isn't yet any kind of Job that runs in the integration tests?

Correct. Or at least, I believe that's the case.

PR #374, opened 2020-05-05
rollups: retry: dig into exception cause to find KDB exception

speezepearson:
(In practice, it looks like the failures passed in here are CompletionExceptions.)


cwbriones:
Seems reasonable enough. All the cases I'm thinking of that might cause problems seem contrived at best (e.g. `KairosDbRequestException` having a cause that is itself a `KairosDbRequestException`, or a causal loop). And the recursion shouldn't be an issue unless we have some absurd chaining going on.

PR #377, opened 2020-05-07
Extract Database execution helper

cwbriones:
Followup to #361.

Now that the JobExecution logic is pulled out of the Repository code, we can extract the shared logic around managing execution state into a helper class (`DatabaseExecutionHelper`) to be consumed by both the `ReportExecutionRepository` and `AlertExecutionRepository`.

The shared attributes for the execution bean were also lifted into a `BaseExecution`.


app/models/ebean/BaseExecution.java
```diff
@@ -0,0 +1,169 @@
...
+     * @param jobId The job ID
+     */
+    public abstract void setJobId(UUID jobId);
+
+    /**
+     * Get the result of this execution, if any.
+     *
+     * @return The result, or null if this execution has not completed.
+     */
+    public abstract @Nullable T getResult();
```
cwbriones, line 85: Perhaps I'll leave a comment in the code itself, but this needed to be abstract so that the subclass is forced to declare the concrete type as a member. If we instead declared it as a `T` in this class, Jackson wouldn't be able to deserialize it.

app/com/arpnetworking/metrics/portal/scheduling/impl/DatabaseExecutionHelper.java
```diff
@@ -0,0 +1,225 @@
...
+
+    /**
+     * Public constructor.
+     *
+     * @param logger The logger for the repository.
+     * @param ebeanServer An ebean server.
+     * @param adapter The execution adapter for the repository.
+     */
+    public DatabaseExecutionHelper(
+            final Logger logger,
```
cwbriones, line 58: Not sure if it's frowned upon to pass in a logger like this, but it seemed like the only way to share the code in `updateExecutionState` without declaring multiple methods (at which point I don't think it'd be worth it).

PR #378, opened 2020-05-08
Proof-of-concept rollup-consistency-checker

speezepearson:
In broad strokes:
- Each MPortal host has: a queue specifying which rollup-datapoints need to be inspected; and some number of consistency-checker actors that poll that queue.
- A consistency-checker actor runs in a loop: fetch a task from the queue; ask Kairos how many samples there are for that rollup and its source material in the relevant time-range; log any discrepancy; and repeat.
- I also added an API endpoint to trigger the consistency-checker, to facilitate test-in-prod.

I chose to make the queue non-persistent: it's simpler, and it's not a catastrophe if a host dies and we lose track of which datapoints we were considering checking.


app/com/arpnetworking/rollups/QueueActor.java
```diff
@@ -0,0 +1,255 @@
...
+import java.util.Optional;
+import java.util.Queue;
+
+/**
+ * An actor with a bounded queue.
+ *
+ * @param <T>
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public class QueueActor<T extends Serializable> extends AbstractActor {
```
ddimensia, line 36: Should this be a de-duping queue?

app/com/arpnetworking/rollups/QueueActor.java
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * {@link QueueActor} actor constructor.
+     *
+     * @param maxSize the maximum size of the queue
+     */
+    public QueueActor(final Optional<Long> maxSize) {
+        _maxSize = maxSize;
+    }
+
+    private final Optional<Long> _maxSize;
```
ddimensia, line 86: I see Ville has infected you with putting your members in the wrong place. ;)

app/global/MainModule.java
```diff
@@ -804,4 +814,58 @@ public ActorRef get() {
         private final Config _configuration;
         private final boolean _enabled;
     }
+
+    private static final class RollupConsistencyCheckerQueueProvider implements Provider<ActorRef> {
```
ddimensia, line 818: Doesn't this need to be either a cluster singleton or a sharded actor?

app/com/arpnetworking/rollups/QueueActor.java
speezepearson: Hey, I was just copying _your_ `RollupGenerator`! :P What's the right place?

app/global/MainModule.java
speezepearson: I am no Akka expert, but I thought that, as written, this would result in each host having its own queue, independent of the other hosts' queues. (And its own consistency-checkers, independent of the other hosts' checkers.) Yeah, this will result in unequal load distribution, but that seems fine, given how little (of MP's) resources the consistency-checking operation consumes.

app/com/arpnetworking/rollups/QueueActor.java
speezepearson: A good point! I've generalized this to a `CollectionActor`, which in practice is backed by a set.

app/com/arpnetworking/rollups/QueueActor.java
ddimensia: I think Ville relaxed this but the Java defacto standard is to put members at the top of the class followed by constructor and then public and private methods.

app/global/MainModule.java
ddimensia: Okay, just wanted to make sure that was a conscious decision.

app/com/arpnetworking/rollups/QueueActor.java
speezepearson: Sounds good! Reorganized.

app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -0,0 +1,579 @@
...
+                        .whenComplete((response, failure) -> requestWork()),
+                getContext().dispatcher()
+        ).to(getSelf());
+    }
+
+    private void sampleCountsReceived(final SampleCounts sampleCounts) {
+        final Task task = sampleCounts.getTask();
+        final Optional<Throwable> failure = sampleCounts.getFailure();
+
+        final Metrics metrics = _metricsFactory.create();
```
vjkoskela, line 125: This should probably be in a try with resources to ensure it's closed even on exception.

app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -0,0 +1,579 @@
...
+            return;
+        }
+
+        final double nOriginalSamples = sampleCounts.getSourceSampleCount();
+        final double nSamplesDropped = nOriginalSamples - sampleCounts.getRollupSampleCount();
+        final double fractionalDataLoss = nSamplesDropped / nOriginalSamples;
+        metrics.setGauge("rollup/consistency_checker/fractional_data_loss", fractionalDataLoss);
+        final LogBuilder logBuilder =
+                // TODO(spencerpearson): probably make this level-thresholding configurable?
+                fractionalDataLoss == 0 ? LOGGER.trace()
```
vjkoskela, line 149: Maybe `Double.compare`?

app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -0,0 +1,579 @@
...
+
+         */
+    }
+
+    /* package private */ static SampleCounts parseSampleCounts(
+            final Task task,
+            final MetricsQueryResponse response
+    ) throws MalformedSampleCountResponse {
+        final Map<String, Long> countsByMetric = Maps.newHashMap();
+        if (response.getQueries().size() != 2) {
```
vjkoskela, line 182: Can you add a comment to explain why you look for size 2 but emit an exception saying you needed exactly 1?

app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -0,0 +1,579 @@
...
+            if (queryResult.getValues().size() != 1) {
+                throw new MalformedSampleCountResponse("expected exactly 1 value, got " + queryResult.getValues().size(), response);
+            }
+            final Optional<Object> value = queryResult.getValues().get(0).getValue();
+            if (!value.isPresent()) {
+                throw new MalformedSampleCountResponse("sample count has null value", response);
+            }
+            final long longValue;
+            try {
+                longValue = Long.parseLong(value.get().toString());
```
vjkoskela, line 199: Is the value always a long? Will this parse scientific notation if it's a very large double not representable as a long?

app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -0,0 +1,579 @@
...
+        }
+        final Long sourceCount = countsByMetric.get(task.getSourceMetricName());
+        final Long rollupCount = countsByMetric.get(task.getRollupMetricName());
+        if (sourceCount == null || rollupCount == null) {
+            throw new MalformedSampleCountResponse(
+                    String.format("expected keys %s and %s", task.getSourceMetricName(), task.getRollupMetricName()),
+                    response
+            );
+        }
+        return new SampleCounts.Builder()
```
vjkoskela, line 213: This is TLB:

https://github.com/ArpNetworking/metrics-portal/blob/8a8b63dd9483ef74c0ae70f38280e4e58f25d688/app/com/arpnetworking/rollups/FailableMessage.java#L59

app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -0,0 +1,579 @@
...
+                .setTask(task)
+                .setSourceSampleCount(sourceCount)
+                .setRollupSampleCount(rollupCount)
+                .build();
+    }
+
+    private MetricsQuery buildCountComparisonQuery(final Task task) {
+        final Consumer<Metric.Builder> setCommonFields = builder -> builder
+                .setAggregators(ImmutableList.of(
+                        new Aggregator.Builder()
```
vjkoskela, line 223: Is `Aggregator` TLB?

app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -0,0 +1,579 @@
...
+                            b.setName(task.getSourceMetricName());
+                        }),
+                        ThreadLocalBuilder.build(Metric.Builder.class, b -> {
+                            setCommonFields.accept(b);
+                            b.setName(task.getRollupMetricName());
+                        })
+                )).build();
+    }
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ConsistencyChecker.class);
```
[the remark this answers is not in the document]
vjkoskela: Until this transitions to ISM and it's parent pom, either convention is acceptable. The enforcement of better Java aligned styles is happening as projects are migrated to ISM.

e.g. https://github.com/ArpNetworking/metrics-client-java/pull/59

app/global/MainModule.java
vjkoskela: I am also curious about this from the API point of view.

The controller allows users to submit tasks, but there is no way to track them or know their result because each host is independent. What's the plan here going forward? (e.g. what does this on-demand/user-initiated use case going to look like?)

^ Not saying it has to be addressed now, just want to understand what that plan is and how independent actors per host fits into it.

app/com/arpnetworking/rollups/ConsistencyChecker.java
speezepearson: Copy-paste error! Fixed.

app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -0,0 +1,579 @@
...
+                : fractionalDataLoss < 0.1 ? LOGGER.warn()
+                : LOGGER.error();
+        logBuilder.setMessage((fractionalDataLoss == 0 ? "no " : "") + "data lost in rollup")
+                .addData("task", task)
+                .addData("nOriginalSamples", nOriginalSamples)
+                .addData("nSamplesDropped", nSamplesDropped)
+                .addData("fractionalDataLoss", fractionalDataLoss)
+                .log();
+
+        /* TODO(spencerpearson, OBS-1176): re-trigger re-execution of bad datapoints, something like
```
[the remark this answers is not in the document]
speezepearson: Yeah, upon reflection, I agree with you about this TODO.

In general, though, I think it's valuable to have comments in the code pointing out potential defects or low-hanging optimization fruit. If you're debugging some problem, you'll see comments that might be relevant, but you don't generally read the entire issue tracker.

Things I could do:
- use a word other than `TODO`
- file issues for all of the TODOs that I added, and cross-reference the comments with the issues

Thoughts?

app/global/MainModule.java
speezepearson: I currently don't intend for there to _be_ a user experience; I only added an API endpoint for debugging purposes.

At some point in the future, we might decide to give the consistency-checker a real UI so that people can submit tasks without going through us; and when that happens, we can add persistence or clustering or whatever. And maybe that switchover will cause some transient issues and result in some tasks getting dropped, but that's okay, because only _after_ that point do we need to start offering reliability guarantees. Until then, it's A-okay for deployments to randomly drop some tasks: the consistency-checker doesn't need to be resilient to occasional transient disruptions.

Tldr: whatever features we add to the consistency-checker in the future, its core is _basically_ going to look like what this diff lays out: some pool of tasks that executors pull from. And because it doesn't need any particularly strong reliability guarantees, it's unusually easy to change.

app/com/arpnetworking/kairos/client/models/TagsQuery.java
```diff
@@ -42,7 +43,7 @@
  * @author Ville Koskela (ville dot koskela at inscopemetrics dot io)
  */
 @Loggable
-public final class TagsQuery {
+public final class TagsQuery implements Serializable {
```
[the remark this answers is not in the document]
speezepearson: I wasn't a fan of this either, and I've backed out this change. I had, earlier, wanted to stick a `MetricsQueryResponse` in an `Exception` that was part of a `FailableMessage`, but some refactors have now made that unnecessary!

app/controllers/RollupController.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public Result enqueueConsistencyCheck() {
+        final ConsistencyChecker.Task task;
+        try {
+            task = _mapper.treeToValue(request().body().asJson(), ConsistencyChecker.Task.class);
+        } catch (final IOException err) {
+            return badRequest(err.getMessage());
+        }
+        _consistencyCheckerQueue.tell(
+                new CollectionActor.Add<>(task),
```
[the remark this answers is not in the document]
speezepearson: Yes! Done.

app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -0,0 +1,579 @@
...
+                getContext().dispatcher()
+        ).to(getSelf());
+    }
+
+    private void sampleCountsReceived(final SampleCounts sampleCounts) {
+        final Task task = sampleCounts.getTask();
+        final Optional<Throwable> failure = sampleCounts.getFailure();
+
+        final Metrics metrics = _metricsFactory.create();
+        metrics.addAnnotation("trigger", task.getTrigger().name());
```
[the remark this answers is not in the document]
speezepearson: Thanks for the tip! https://github.com/ArpNetworking/metrics-portal/pull/383

app/com/arpnetworking/rollups/ConsistencyChecker.java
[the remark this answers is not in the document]
cwbriones: I'd rule out using something other than `TODO` by virtue of the fact that it wouldn't be greppable or even universally understood. As far as debuggability - if it's mainly informational, then I think it's fine to just leave it as an ordinary comment.

If it's not a "why" answering comment or something that can be used to assess the state of the code as-is, then my main concern is timeline visibility and what value the TODO brings there.

How long is this comment going to live in the code? IMO if the answer is...
1. **I'm not sure. this is a nice-to-have or future improvement but it's not scheduled.** I don't think this case should be in the code. If we're not sure when we'll get to it, then this comment will likely rot.
2. **I'm working on this in an immediate follow-up PR** or **There's a ticket associated with this that I _must_ do in the next month or two** - Sure, those seem fine to leave in. I think even the former seems OK without a ticket so long as the timeline is appropiately conveyed.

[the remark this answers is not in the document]
speezepearson: So noted, and (I claim) resolved!

app/com/arpnetworking/rollups/CollectionActor.java
```diff
@@ -0,0 +1,261 @@
...
+import java.util.Optional;
+
+/**
+ * An actor with a bounded-size collection.
+ *
+ * @param <T> the type of item being held
+ * @param <C> the type of collection to hold items in
+ * @author Spencer Pearson (spencerpearson at dropbox dot com)
+ */
+public class CollectionActor<T extends Serializable, C extends Collection<T>> extends AbstractActor {
```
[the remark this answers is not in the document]
speezepearson: I also don't love having written this actor, but I think it's the right way to go:

I don't think that using the executor's mailbox as its work-queue gives us what we want:
- AFAICT there's no way to "pause" the actor while a KairosDB request is in flight, so we would have to choose between (a) losing concurrency limits or (b) blocking the actor's thread until the response comes back;
- we'd lose some load-balancing capability;
- we wouldn't be able to do stuff like deduplicate tasks unless we write our own `Mailbox` implementation.

Akka Streams might be able to replace some of this boilerplate, but at the cost of introducing subtleties and gnarly type signatures that, after spending a day failing to write a deduping queue, I'm pessimistic about my ability to wrangle correctly.

app/global/MainModule.java
vjkoskela: Okay, please document this clearly.

app/com/arpnetworking/rollups/CollectionActor.java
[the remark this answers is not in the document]
vjkoskela: Fair enough. Thanks for looking into it.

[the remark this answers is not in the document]
BrandonArp: Why do you assume the messages need to stay in the mailbox? Its easy to require a message and put it into a LinkedHashMap to get both dedupe and queue behavior. Then you control dispatch by checking if the queue is empty and processing the next message when one finishes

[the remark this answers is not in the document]
speezepearson: Hmm, am I right in thinking that what you're proposing is _conceptually_ identical to what's currently implemented? There's a buffer that work goes into, and there's some number of workers that pull work out of that pool.

So, what you're proposing is: combine the worker-actors (currently `class ConsistencyChecker`) and the buffer-actor (currently `class CollectionActor`) into a _single_ actor, that arranges for (up to) some fixed number of worker-CompletionStages to be running at a time.

Accurate?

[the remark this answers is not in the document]
speezepearson: Prototyped it, liked it, implemented it. This class no longer exists.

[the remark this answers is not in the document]
BrandonArp: I was merely throwing it out there as an alternative.  We've used it in the past.  If you like it better, great.

app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -0,0 +1,696 @@
...
+                    .add("_period", _period)
+                    .add("_startTime", _startTime)
+                    .add("_trigger", _trigger)
+                    .toString();
+        }
+
+        /**
+         * Builder for {@link Task}.
+         */
+        public static final class Builder extends OvalBuilder<Task> {
```
vjkoskela, line 420: ThreadLocalBuilder?

PR #380, opened 2020-05-11
Alert execution repository

cwbriones:


app/com/arpnetworking/metrics/portal/scheduling/JobExecutionRepository.java
```diff
@@ -43,12 +43,24 @@
...
 
     /**
-     * Get the last successful execution, if any.
+     * Get the most recently scheduled execution, if any.
+     * <p>
+     * This could possibly return an execution that's pending completion.
      *
      * @param jobId The UUID of the job that completed.
      * @param organization The organization owning the job.
+     * @return The last successful execution.
```
cwbriones, line 52: Just noticed, will fix.

app/models/internal/impl/DefaultAlertEvaluationResult.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Default implementation for an {@code AlertEvaluationResult}.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+@JsonDeserialize(builder = DefaultAlertEvaluationResult.Builder.class)
```
cwbriones, line 36: I would think this is handled by the standard object mapper config but without this Jackson complains that it can't deserialize `ImmutableList` (since it's performing it through reflection and not through the setter below.

conf/postgresql.application.conf
```diff
@@ -94,6 +94,7 @@ hostRepository.type = "com.arpnetworking.metrics.portal.hosts.impl.DatabaseHostR
 # Alerts
 # ~~~~~
 alertRepository.type = "com.arpnetworking.metrics.portal.alerts.impl.DatabaseAlertRepository"
+alertExecutionRepository.type = "com.arpnetworking.metrics.portal.alerts.impl.DatabaseAlertExecutionRepository"
```
cwbriones, line 97: This just reminded me that I need to add `NoAlertExecutionRepository` as a default.

test/java/com/arpnetworking/metrics/portal/integration/repositories/JobExecutionRepositoryIT.java
```diff
@@ -0,0 +1,276 @@
...
+import static org.junit.Assert.fail;
+
+/**
+ * Generic integration tests for implementations of {@link JobExecutionRepository}.
+ *
+ * @param <T> The job result type.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public abstract class JobExecutionRepositoryIT<T> {
```
cwbriones, line 49: This is almost exactly what's in `ReportExecutionRepositoryIT`. You may just want to `git diff` against that.

I'm going to follow up this diff with a change to make that code use this instead, but the gist of it is `setUp` had to be fixed to be generic and the success tests needed a `newResult` callback since the result type became generic.

app/models/internal/impl/DefaultAlertEvaluationResult.java
```diff
@@ -0,0 +1,98 @@
...
+            super(DefaultAlertEvaluationResult::new);
+        }
+
+        /**
+         * Set the firing tags. Defaults to empty.
+         *
+         * @param firingTags The set of firing tags.
+         * @return This instance of {@code Builder}.
+         */
+        @JsonProperty("firingTags")
```
vjkoskela, line 92: This definitely should not be necessary. If it is, combined with the above, I suspect we may be using the wrong `ObjectMapper` instance (e.g. one that is not configured to support our usual builder pattern).

conf/db/migration/metrics_portal_ddl/V21__create_alert_execution_tables.sql
```diff
@@ -0,0 +1,54 @@
+-- Copyright 2018 Dropbox, Inc.
```
vjkoskela, line 1: ?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertExecutionRepository.java
```diff
@@ -0,0 +1,194 @@
...
+        assertIsOpen();
+        return _ebeanServer.find(AlertExecution.class)
+                .where()
+                .eq("alert_id", jobId)
+                .eq("organization.uuid", organization.getId())
+                .setMaxRows(1)
+                .orderBy()
+                .desc("scheduled")
+                .findOneOrEmpty()
+                .map(DatabaseExecutionHelper::toInternalModel);
```
vjkoskela, line 117: I don't see this class in the review, so I assume it's an existing class. Generally, not a fan of static helper classes, and while we do have them in places, it's generally better to create instances of them in order to facilitate testing and injection. If this is indeed existing code, feel free just to add a `TODO` into that class.

app/models/internal/impl/DefaultAlertEvaluationResult.java
vjkoskela: Where is this deserialized?

app/models/ebean/AlertExecution.java
```diff
@@ -0,0 +1,143 @@
...
+public final class AlertExecution extends BaseExecution<AlertEvaluationResult> {
+    @Id
+    @ManyToOne(optional = false)
+    @JoinColumn(name = "organization_id")
+    private Organization organization;
+    @Id
+    @Column(name = "alert_id")
+    private UUID alertId;
+    @Nullable
+    @DbJsonB
```
vjkoskela, line 58: This is your likely serialization culprit.

Here's the relevant thread:

https://github.com/ebean-orm/ebean/issues/542

This change related to 542 may offer some help but I have not traced it all the way down:

https://github.com/ebean-orm/ebean/commit/13f7d37e76c0712f7daf59fb689191e4bdba8d8a#diff-2a7fe1eac4aafd91195da026c9a88a9fR675

Specifically, here it looks into `ServerConfig` for an `ObjectMapper`, if we can set that before this method is called to the one we created in Metrics Portal then you would get all the goodies we are used to having with ObjectMapper usage.

If not, the fallback strategy here imo is to not use @DbJsonB and instead perform the serialization/deserialization in the repository object. The only downfall of this imo is that the schema may not be easily upgradable to @DbJsonB in the future (unless you can find a type that works today with manual serde and in the future with Ebean serde).

conf/db/migration/metrics_portal_ddl/V21__create_alert_execution_tables.sql
cwbriones: copypasta

app/com/arpnetworking/metrics/portal/scheduling/JobExecutionRepository.java
```diff
@@ -43,12 +43,24 @@
...
+     */
+    Optional<JobExecution<T>> getLastScheduled(UUID jobId, Organization organization);
+
+    /**
+     * Get the last successful execution, if any.
+     *
+     * @param jobId The UUID of the job that completed.
+     * @param organization The organization owning the job.
      * @return The last successful execution.
+     * @throws NoSuchElementException if no job has the given UUID.
```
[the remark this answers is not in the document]
cwbriones: This was preserved from the old Job Repository API. I think the intent is to distinguish between executions that do not exist and parent entities that do not exist.

1. `Optional.of(execution)` - That job was valid and that execution existed
2. `Optional.empty()` - That job was valid but no execution was scheduled for that time
3. `NoSuchElementException` - That job does not exist OR that organization does not exist

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertExecutionRepository.java
```diff
@@ -0,0 +1,194 @@
...
+            final String message = String.format(
+                    "Could not find org with organization.uuid=%s",
+                    organization.getId()
+            );
+            throw new EntityNotFoundException(message);
+        }
+        final Optional<AlertExecution> existingExecution = org.flatMap(r ->
+                _ebeanServer.createQuery(AlertExecution.class)
+                        .where()
+                        .eq("organization", org.get())
```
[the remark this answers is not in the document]
cwbriones: As far as I can tell it's identical - Ebean is smart enough to create the same query. Regardless, I'll change this to be consistent with the code below.

app/models/ebean/AlertExecution.java
cwbriones: Thanks for the pointers!

I went ahead and configured this in both MainModule and EbeanServerHelper. It looks like it did the trick.

cwbriones: Hmm. Apparently it works fine in the unit tests, but I'm actually discarding configuration when trying to use it for integration tests because I don't load the same configuration that play does using the hocon file.

I'm still trying to figure out how to override `ServerConfig` from within Play. If this takes more than an hour I'll probably go with the fallback strategy mentioned.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutionRepository.java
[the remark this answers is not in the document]
vjkoskela: If you have not already, please document this in the Javadoc for each method (I suppose in the interface).

app/models/ebean/AlertExecution.java
vjkoskela: I'm a little confused by what you mean when you say integration tests discard the configuration.

There are two styles of "integration" tests.

1) `*IT.java` code is executed in the Maven JVM against Metrics Portal JVM (e.g. via its APIs) running in Docker. The configuration loaded by Metrics Portal is the one packaged by default into the Docker image.

2) `*IT.java` code is executed in the Maven JVM and loads and runs classes from Metrics Portal against actual dependencies (e.g. Postgresql, Cassandra, KairosDb). There is no configuration in this context in the Maven JVM (nor should there be).

Instead, you will need to have the tests inject the correct `ObjectMapper`. You need to be careful because you want all Ebean tests to inject a single and consistent `ObjectMapper` instance. 

We actually have a similar problem for working with the Play APIs because they use (the same) custom `ObjectMapper`. The code for generating this for tests is here:

https://github.com/ArpNetworking/metrics-portal/blob/master/test/java/com/arpnetworking/testing/SerializationTestUtils.java

You should probably re-use it with some way for Ebean tests (e.g. BaseEbean abstract test class with `@BeforeClass` that does the injection). Shoot me a text and we can talk specifics.

app/com/arpnetworking/metrics/portal/scheduling/impl/DatabaseExecutionHelper.java
```diff
@@ -73,6 +73,8 @@ public DatabaseExecutionHelper(
      * @return An internal model for this execution
      */
     public static <T, E extends BaseExecution<T>> JobExecution<T> toInternalModel(final E beanModel) {
+        // TODO(cbriones) - This should not be static, repositories should call this from an instance.
```
vjkoskela, line 76: If this is not in your development plan to address with Alerting, please file an issue for it against the OSS project.

app/global/MetricsPortalServerConfigStartup.java
```diff
@@ -0,0 +1,59 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.inject.Inject;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import io.ebean.config.ServerConfig;
+import io.ebean.event.ServerConfigStartup;
+
+/**
+ * Plugin class to configure Ebean's {@link ServerConfig} at runtime.
+ * <p>
+ * This is necessary for configuring the Ebean server with dependencies constructed via Guice, for instance. This
```
vjkoskela, line 30: nit: try to wrap comments at 80 characters.

app/global/MetricsPortalServerConfigStartup.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public MetricsPortalServerConfigStartup() {}
+
+    @Override
+    public void onStart(final ServerConfig serverConfig) {
+        LOGGER.info().setMessage("Initializing Ebean ServerConfig").log();
+        if (gObjectMapper == null) {
+            // In some cases we manually load the ebean model classes via ServerConfig#addPackage (see EbeanServerHelper).
+            //
+            // If this class is accidentally loaded in those environments, then injection won't occur and we'll silently overwrite
```
vjkoskela, line 53: By "loaded" I assume you mean "instantiated"? (loaded imo implies more class loading which obviously must occur before instantiation, but can also occur without instantiation)

app/global/MetricsPortalServerConfigStartup.java
```diff
@@ -0,0 +1,59 @@
...
+            // the configured ObjectMapper with null. Explicitly throwing makes this error appear obvious.
```
vjkoskela, line 54: I believe in either case (setting it to `null` or leaving it alone) this mean that the `ObjectMapper` is left as its default instantiated value, such that any code relying on the custom behavior will fail in weird ways. In production this could even cause data corruption (due to serializing this incorrectly). Right?

Should we be more draconian here and say `System.exit()`?

app/global/MetricsPortalServerConfigStartup.java
cwbriones: Yes, I can see how that can be confusing though - happy to tweak the language there.

app/com/arpnetworking/metrics/portal/scheduling/impl/DatabaseExecutionHelper.java
cwbriones: Filed https://github.com/ArpNetworking/metrics-portal/issues/384

app/global/MetricsPortalServerConfigStartup.java
cwbriones: After seeing this I thought I could just set this option in IntelliJ. Unfortunately not only can you not set different margins for comments v. code, the ticket for this was opened [7 years ago](https://youtrack.jetbrains.com/issue/IDEA-114661).

In fact, another was opened [15(!!!) years ago](https://youtrack.jetbrains.com/issue/IDEABKL-1289)

I'll just do this manually...

app/global/MetricsPortalServerConfigStartup.java
cwbriones: I don't think that's necessary - this exception seems to kill the process on startup because Guice fails to instantiate the repository. The dependency chain looks like:

`DatabaseXXXRepository -> EbeanServer -> ObjectMapper -> **Boom**`

app/models/ebean/AlertExecution.java
cwbriones: These integration tests are the second flavor (only one JVM, no Play APIs, "real" external databases). I understand that there are two kinds but given the context of the PR I thought it was reasonable to omit any specifics.

What I had meant by "discard configuration" is that because of the way this is now set up to configure the `ServerConfig`, this `MetricsPortalServerConfigStartup` class needs to be loaded by ebean in order to call the callback and populate the `ObjectMapper` field. As mentioned below, it has to be called at inject time otherewise the ObjectMapper is null and everything fails.

In the tests, we load the ebean classes manually from within [EbeanServerHelper](https://github.com/ArpNetworking/metrics-portal/blob/2c9237c77c9bd1b05c4b0eb31e9f9b581f6e6407/test/java/com/arpnetworking/metrics/portal/integration/test/EbeanServerHelper.java#L89). This sets the ObjectMapper to something reasonable, but if `MetricsPortalServerConfigStartup` gets loaded here, it will clobber the ObjectMapper. This scenario is what I mean by "discard" - we overwrite any configured value. I was accidentally doing this in one of my earlier commits and it caused my tests to fail.

It sounds like then there are currently 2 (3?) locations where we create an object mapper, and they all vary independently (even though they should all be the same):
* [MainModule](https://github.com/ArpNetworking/metrics-portal/blob/master/app/global/MainModule.java#L290)
* [EbeanServerHelper](https://github.com/ArpNetworking/metrics-portal/blob/2c9237c77c9bd1b05c4b0eb31e9f9b581f6e6407/test/java/com/arpnetworking/metrics/portal/integration/test/EbeanServerHelper.java#L89)
* [SerializationTestUtils](https://github.com/ArpNetworking/metrics-portal/blob/master/test/java/com/arpnetworking/testing/SerializationTestUtils.java)

At the very least I can make the last two the same since they're both run during tests.

app/global/MetricsPortalServerConfigStartup.java
vjkoskela: ^ Please add that to the comment (or something to that effect)

app/models/ebean/AlertExecution.java
vjkoskela: Yes please! Reducing the two in tests to just one would be very helpful. The duplication and matching with prod is already called out here:

https://github.com/ArpNetworking/metrics-portal/blob/master/test/java/com/arpnetworking/testing/SerializationTestUtils.java#L40

Feel free to file an issue to refactor the Prod code so it can be used from Test as well. That would make it just one definition. Although to be clear, the risk there is someone modifying the prod definition and the tests passing _BUT_ not being backwards compatible. The dual definition should at least catch such configuration changes to ObjectMapper. 

So 2 is better than 3, but 1 may not be better than 2. ;)

PR #382, opened 2020-05-12
Update the alert internal model

cwbriones:
- Updates the fields in the Alert model
- Rips out all code related to the creation and evaluation of alerts in the previous form.

For the latter change in particular, I figured that there wasn't much value in keeping the existing data and view layers around with this change in place since the interface additions would propagate into those layers and I don't intend on supporting them now anyway.

I'm working on limited knowledge here since this repo is out in the wild - if there's any concerns about deleting all of this code I'm open to hearing them, but just from initial discussions it seemed like this was fine.


app/com/arpnetworking/metrics/portal/alerts/AlertJob.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public Schedule getSchedule() {
+        // TODO(cbriones): If the alert is enabled, this should return an interval corresponding to the smallest query period
+        // in that alert's query.
+        return NeverSchedule.getInstance();
+    }
+
+    @Override
+    public Duration getTimeout() {
+        return Duration.of(1, ChronoUnit.MINUTES);
```
vjkoskela, line 70: I assume this will eventually be specifiable as part of the alert definition.

PR #383, opened 2020-05-13
bump HTTP sink version

speezepearson:


pom.xml
```diff
@@ -104,7 +104,7 @@
     <jsr305.version>3.0.2</jsr305.version>
     <logback.steno.version>1.18.0</logback.steno.version>
     <metrics.client.version>0.10.0</metrics.client.version>
```
vjkoskela, line 106: You need to upgrade the associated dependencies as well to ensure we don't have version conflicts.

vjkoskela: You can see it in the build output:

```
[WARNING] 
Dependency convergence error for com.arpnetworking.metrics:metrics-client:0.10.0 paths to dependency are:
```

https://api.travis-ci.com/v3/job/333369089/log.txt

vjkoskela:
The new metrics client uses the V3 client protocol. Unfortunately, due to the poor design the client configurability the endpoint (including version) is the only way to configure the target (MAD) host and port:

https://github.com/ArpNetworking/metrics-portal/blob/50d523f6775e2088fe6659c9e216fd4c2bd47984/app/global/MainModule.java#L214

So we need to update this to be v3.

I would recommend testing locally with the Metrics project's `build/start.sh` (now fixed).

pom.xml
speezepearson: I'll fix that. For my edification, though -- it looks to me like there are 160 other `Dependency convergence error`s as well. Why is this one in particular worth fixing?

speezepearson: From my local build, after updating a few dependencies to their latest releases:
```
Dependency convergence error for com.arpnetworking.metrics:metrics-client:0.11.1 paths to dependency are:
+-com.arpnetworking.metrics:metrics-portal:0.9.29-SNAPSHOT
  +-com.arpnetworking.metrics:metrics-client:0.11.1
and
+-com.arpnetworking.metrics:metrics-portal:0.9.29-SNAPSHOT
  +-com.arpnetworking.metrics.extras:apache-http-sink-extra:0.11.1
    +-com.arpnetworking.metrics:metrics-client:0.11.1
and
+-com.arpnetworking.metrics:metrics-portal:0.9.29-SNAPSHOT
  +-com.arpnetworking.metrics.extras:jvm-extra:0.11.0
    +-com.arpnetworking.metrics:metrics-client:0.11.0
and
+-com.arpnetworking.metrics:metrics-portal:0.9.29-SNAPSHOT
  +-com.arpnetworking.metrics.extras:incubator-extra:0.11.0
    +-com.arpnetworking.metrics:metrics-client:0.11.0
```
I assume that this means I need to update the `pom.xml`s of `jvm-extra` and `incubator-extra`, to depend on `metrics-client:0.11.1`, and then cut new releases of those projects?

speezepearson: (assuming yes: https://github.com/ArpNetworking/metrics-jvm-extra/pull/14 and https://github.com/ArpNetworking/metrics-incubator-extra/pull/9 )

vjkoskela: > Why is this one in particular worth fixing?

They are all worth fixing. I've been doing this with MAD, and a lesser extent with CAGG. Metrics Portal most of these are more difficult to fix because they require a Play Framework upgrade. This in turn, requires either ripping out or finding a new technology for Metrics Portal's Cassandra schema management. 

> I assume that this means I need to update the pom.xmls of jvm-extra and incubator-extra

Yes, I have the same action item to go and update those dependencies myself for consistency. We need to ensure we do this once to those packages to bring consistency within them but also to other dependencies (e.g. slf4j => 1.7.30). I'll post into those some details or see if I can knock out the important ones tonight.

vjkoskela: I've updated the underlying dependencies (of these four packages) and released the packages you need:

* metrics-client
* metrics-apache-http-sink-extra
* metrics-jvm-extra
* metrics-incubator-extra

All should be version `0.11.2` and have no self-conflicts.

Ref:
https://repo.maven.apache.org/maven2/com/arpnetworking/metrics/metrics-client/0.11.2/
https://repo.maven.apache.org/maven2/com/arpnetworking/metrics/extras/incubator-extra/0.11.2/
https://repo.maven.apache.org/maven2/com/arpnetworking/metrics/extras/jvm-extra/0.11.2/
https://repo.maven.apache.org/maven2/com/arpnetworking/metrics/extras/apache-http-sink-extra/0.11.2/

speezepearson:
obsoleted by #416

PR #388, opened 2020-05-22
Run consistency-checker on newly-written rollup datapoints

speezepearson:


app/com/arpnetworking/kairos/client/KairosDbClient.java
```diff
@@ -15,57 +15,79 @@
...
+     *   (Due to KairosDB implementation issues, it's actually the set of all names/values over a couple-week period
+     *    that includes the given time.)
+     */
+    public CompletionStage<ImmutableMultimap<String, String>> queryMetricTags(final String metricName, final Instant time) {
+        return queryMetricTags(ThreadLocalBuilder.build(TagsQuery.Builder.class, tqb -> tqb
+                .setMetrics(ImmutableList.of(ThreadLocalBuilder.build(MetricTags.Builder.class, mtb -> mtb.setName(metricName))))
+                .setStartTime(time)
+                .setEndTime(time.plus(Duration.ofDays(1)))
+        )).thenApply(response -> response.getQueries().get(0).getResults().get(0).getTags());
+    }
```
speezepearson, line 92: I think that this method should also be used by the RollupGenerator, but its current structure makes that hard to do. I might perform the necessary re-factor in a separate PR.

app/com/arpnetworking/rollups/RollupExecutor.java
```diff
@@ -156,9 +158,49 @@ private void fetchRollup() {
     private void handleFinishRollup(final FinishRollupMessage message) {
         _metrics.recordCounter("rollup/executor/finish_rollup_message/received", 1);
         _rollupManager.tell(message, getSelf());
+        final RollupDefinition defn = message.getRollupDefinition();
+        if (shouldRequestConsistencyCheck(message)) {
```
orborde, line 162: This seems like it should be the responsibility of the RollupManager. Then you don't have a check for rollup failure in two places.

conf/portal.application.conf
```diff
@@ -133,6 +133,7 @@ rollup {
   generator.count = 5
   executor.count = 5
   executor.pollInterval = "5min"
+  executor.consistency_check_fraction_of_writes = 0.1
```
orborde, line 136: Can you set yourself a followup to, about a week after we roll this change to prod and therefore have some confidence in its won't-break-everything, turn this up to 1? Unless you have reason to believe that it'll be an outage-causing amount of load (which I doubt is the case).

Reasoning: 10x the coverage is valuable because you'll detect problems about 10x faster. And we'd strongly prefer to run inefficiently to detect problems earlier than later - right now, it's not a big deal for us to be inefficient.

PR #389, opened 2020-05-26
Read-Only Alert Repository

cwbriones:
Introduces a read-only repository for reading alert models that have been serialized to a file. This is essentially just an in-memory repository that hydrates itself from a File - once the repository is opened there is no subsequent File I/O.

UUIDs are computed on the fly when absent to allow for consistency between internal UUID keys and name-based keys. This is fairly straightforward using named-based UUIDs but since JDK8 doesn't include native support for v5 IDs, I've pulled in JUG for this purpose. If this name-based use case becomes more widespread in the org's repos we'll likely pull this up into commons.


app/com/arpnetworking/metrics/portal/alerts/impl/FileSystemAlertRepository.java
```diff
@@ -0,0 +1,402 @@
...
+                    reader,
+                    AlertGroup.class);
+        } catch (final IOException e) {
+            throw new RuntimeException("Could not load alerts", e);
+        }
+        final ImmutableMap.Builder<UUID, Alert> mapBuilder = ImmutableMap.builder();
+        for (final SerializedAlert fsAlert : group.getAlerts()) {
+            final UUID uuid = fsAlert.getUUID().orElseGet(() -> computeUUID(fsAlert));
+
+            // TODO(cbriones):
```
cwbriones, line 171: I've left this TODO and the one below in here since as mentioned in the description these are must-fix problems.

app/com/arpnetworking/metrics/portal/alerts/impl/FileSystemAlertRepository.java
```diff
@@ -0,0 +1,402 @@
...
+/**
+ * An alert repository that loads definitions from the filesystem.
+ *
+ * @apiNote
+ * This repository is read-only, so any additions, deletions, or updates will
+ * result in an {@link UnsupportedOperationException}.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com).
+ */
+public class FileSystemAlertRepository implements AlertRepository {
```
cwbriones, line 61: On second thought, I'll probably rename this unless it changes considerably from its current form since this name seems to imply that I'm reading each alert individually from their own file vs. all alerts in a single file.

app/com/arpnetworking/metrics/portal/alerts/impl/FileAlertRepository.java
```diff
@@ -0,0 +1,403 @@
...
+                            .build();
+            mapBuilder.put(uuid, alert);
+        }
+        return mapBuilder.build();
+    }
+
+    private UUID computeUUID(final SerializedAlert alert) {
+        // TODO(cbriones): UUID v5 rather than v3.
+        //
+        // v3 UUID identifiers use MD5 while v5 uses SHA-1.
```
speezepearson, line 199: This seems like it will become much harder to fix once we push this code for the first time, since the UUIDs will change for all the alerts that don't specify UUIDs. True? Or am I missing something?

test/java/com/arpnetworking/metrics/portal/alerts/impl/FileAlertRepositoryTest.java
```diff
@@ -0,0 +1,160 @@
...
+import static org.junit.Assert.fail;
+
+/**
+ * Unit Tests for {@link FileAlertRepository}.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public class FileAlertRepositoryTest {
+    private static final UUID METADATA_ALERT_ID = UUID.fromString("0eca730a-5f9a-49db-8711-29a49cac98ff");
+    private static final Map<String, Object> EXPECTED_METADATA = ImmutableMap.of(
```
speezepearson, line 57: Looks to me like this is only used by one test, and will only ever conceivably be used by one test. Can we make it a local variable in that test, to make its limited use clearer?

test/java/com/arpnetworking/metrics/portal/alerts/impl/FileAlertRepositoryTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+    @Test
+    public void testInvalidPathOnOpen() {
+        final Path path;
+        try {
+            path = Paths.get("valid_but_nonexistent");
+        } catch (final InvalidPathException e) {
+            fail("Test path should be valid");
+            return; // Needed to convince the compiler path is initialized.
+        }
```
speezepearson, line 92: I think you could avoid this dance by adding `... throws Exception` to the signature. (I would never suggest this for _production_ code, but it seems A-okay to me in the context of this test, since (a) nobody ever _calls_ it, and (b) throwing an exception will cause it to fail, which is what we want.)

speezepearson:
Lingering things that I'm not clear on, but assume you have a plan for:
- Do we really want queries to have type `String`? Why not have a real class for them?
- It sounds like you plan to tweak / split up the `MetricsQuery` interface, to separate out the query times from the more abstract description of the time-series to query?

cwbriones:
Perhaps this is going to get a bit wordy since it's been on my mind this week, but to give context on both those points:

> Do we really want queries to have type String ?

That's only intended for this current serialized form - after being read from the file it should be wrapped into a `MetricsQuery` instance, or whatever abstraction we end up (see next point).

> It sounds like you plan to tweak / split up the MetricsQuery interface, to separate out the query times from the more abstract description of the time-series to query?

Yes, I think that's somewhat necessary or at least there's definitely a distinction between the form of the query as specified in an alert vs. one consumable by some generic query API. The reason being, if it were a valid query on its own, it would contain a time range. The query for the alert doesn't itself contain that information as it should be inferred from the scheduled time and the smallest aggregator used within the query.

app/com/arpnetworking/metrics/portal/alerts/impl/FileAlertRepository.java
cwbriones: I was thinking that if anything ever required us to change how we created the UUID anyway we could "freeze" it into the serialized form. _but_ I will still see if I can do anything about this in this diff.

The main hiccup is that JDK8 doesn't support UUID v5 so I'll either need to copy-pasta this in or pull in another dep.

speezepearson: "freeze it" == "take all the UUIDless alert definitions that are imported into MPortal, and give them explicit v3 UUIDs"?

cwbriones: Yes, freeze would mean to take the existing UUIDs they already have and assign them to the definitions so that they don't get computed at read time. Then we would be free to change how they are computed for newer definitions. I think we would do all of this on the config side though outside of MPortal's runtime.

speezepearson:
> Perhaps this is going to get a bit wordy since it's been on my mind this week, but to give context on both those points:
> 
> > Do we really want queries to have type String ?
> 
> That's only intended for this current serialized form - after being read from the file it should be wrapped into a `MetricsQuery` instance, or whatever abstraction we end up (see next point).
> 
> > It sounds like you plan to tweak / split up the MetricsQuery interface, to separate out the query times from the more abstract description of the time-series to query?
> 
> Yes, I think that's somewhat necessary or at least there's definitely a distinction between the form of the query as specified in an alert vs. one consumable by some generic query API. The reason being, if it were a valid query on its own, it would contain a time range. The query for the alert doesn't itself contain that information as it should be inferred from the scheduled time and the smallest aggregator used within the query.

Gotcha. Okay, concerns resolved!

app/com/arpnetworking/metrics/portal/alerts/impl/FileAlertRepository.java
speezepearson: Hrrm... if it were _just us_ who needed to do that, I'd be on board, but if it's _all_ users of MPortal alerts, I feel like a backwards-incompatible change is likely to get shot down. It seems pretty easy to pull in a library that would give us UUIDv5, though -- am I mistaken?

PR #390, opened 2020-05-28
Metrics Portal Histogram Out Query Inject Merge

abbeyqy:


app/controllers/KairosDbProxyController.java
```diff
@@ -165,6 +173,12 @@ public KairosDbProxyController(
                         Results.badRequest("All queried metrics must have at least one aggregator"));
             }
 
+            if (_addMergeAggregator) {
+                return _kairosService.queryMetrics(addMergeAggregator(metricsQuery))
```
vjkoskela, line 177: nit: Should we make `metricsQuery` non-final and change it here (e.g. `metrics=addMergeAggregator(metricsQuery)`) and then let the normal flow resume?

app/controllers/KairosDbProxyController.java
```diff
@@ -173,6 +187,29 @@ public KairosDbProxyController(
         }
     }
 
+    MetricsQuery addMergeAggregator(final MetricsQuery metricsQuery) {
+        final List<Metric> newMetrics = new ArrayList<>();
+        for (Metric metric : metricsQuery.getMetrics()) {
```
vjkoskela, line 192: nit: `final Metric metric`

app/controllers/KairosDbProxyController.java
```diff
@@ -173,6 +187,29 @@ public KairosDbProxyController(
...
+            List<Aggregator> newAggregators = new ArrayList<>();
+            if (!metric.getAggregators().isEmpty() && !(metric.getAggregators().get(0).getName()).equals("merge")
```
vjkoskela, line 194: Should this condition be moved into a function? Probably make it easier to test as well.

app/controllers/KairosDbProxyController.java
```diff
@@ -220,8 +257,11 @@ public KairosDbProxyController(
     private final ObjectMapper _mapper;
     private final boolean _filterRollups;
     private final boolean _requireAggregators;
+    private final boolean _addMergeAggregator;
     private final KairosDbService _kairosService;
 
+    private static final ImmutableSet<String> NONHISTAGGREGATORS = ImmutableSet.of("sum", "count", "avg", "max", "min");
```
vjkoskela, line 263: nit: statics are capital-snakecase; e.g. `NON_HISTOGRAM_AGGREGATORS`

app/controllers/KairosDbProxyController.java
```diff
@@ -173,6 +187,29 @@ public KairosDbProxyController(
         }
     }
 
+    MetricsQuery addMergeAggregator(final MetricsQuery metricsQuery) {
```
vjkoskela, line 190: It's unclear from the method name that this is conditional add of the merge aggregator.

app/controllers/KairosDbProxyController.java
```diff
@@ -173,6 +187,29 @@ public KairosDbProxyController(
...
+        final List<Metric> newMetrics = new ArrayList<>();
+        for (Metric metric : metricsQuery.getMetrics()) {
+            List<Aggregator> newAggregators = new ArrayList<>();
```
vjkoskela, line 193: Nit: the creation of the new `ArrayList` is unnecessary for the `else` branch. Considering deferring creation into the first branch.

app/controllers/KairosDbProxyController.java
```diff
@@ -173,6 +187,29 @@ public KairosDbProxyController(
...
+                final Optional<Sampling> originalSampling = metric.getAggregators().get(0).getSampling();
+                newAggregators.add(ThreadLocalBuilder.build(Aggregator.Builder.class,
+                        b -> b.setName("merge").setSampling(originalSampling.map(
+                                sampling -> ThreadLocalBuilder.clone(sampling, Sampling.Builder.class)).orElse(null))));
+                newAggregators.addAll(metric.getAggregators());
+            } else {
+                newAggregators = metric.getAggregators();
+            }
+            final ImmutableList<Aggregator> finalNewAggregators = ImmutableList.copyOf(newAggregators);
+            newMetrics.add(ThreadLocalBuilder.clone(metric, Metric.Builder.class, b->b.setAggregators(finalNewAggregators)));
```
vjkoskela, line 206: We could avoid copying the `Metrics` object if the `newAggregators` was not modified. For example, marking `newAggregators` as `@Nullable` and initially setting it to `null` and not setting it in the `else` branch at all. Now we know whether we need to modify the metric or not.

conf/portal.application.conf
```diff
@@ -163,6 +163,7 @@ kairosdb {
 }
 kairosdb.proxy {
   requireAggregators = false
+  addMergeAggregator = true
```
vjkoskela, line 166: Should default to false for backwards compatibility.

test/java/controllers/KairosDbProxyControllerTest.java
```diff
@@ -194,4 +195,44 @@ public void testQueryRequiresAggregatorOff() {
...
+        final Result result = Helpers.invokeWithContext(request, Helpers.contextComponents(), () -> {
+            final CompletionStage<Result> completionStage = _controller.queryMetrics();
+            return completionStage.toCompletableFuture().get(10, TimeUnit.SECONDS);
+        });
+
+        assertEquals(Http.Status.OK, result.status());
+        assertEquals("{\"queries\":[]}", Helpers.contentAsString(result));
+
+        final MetricsQuery newMetricsQuery = _controller.addMergeAggregator(builder.build());
+        assertEquals("merge", newMetricsQuery.getMetrics().get(0).getAggregators().get(0).getName());
```
vjkoskela, line 234: You should also assert that the merge used the correct period from the aggregator immediately following it.

test/java/controllers/KairosDbProxyControllerTest.java
```diff
@@ -194,4 +195,44 @@ public void testQueryRequiresAggregatorOff() {
         assertEquals(Http.Status.OK, result.status());
         assertEquals("{\"queries\":[]}", Helpers.contentAsString(result));
     }
+
+    @Test
+    public void testAddMergeAggregator() {
```
vjkoskela, line 200: Should we have a test where it is not added?

app/controllers/KairosDbProxyController.java
```diff
@@ -173,6 +187,29 @@ public KairosDbProxyController(
...
     }
 
+    MetricsQuery addMergeAggregator(final MetricsQuery metricsQuery) {
+        final List<Metric> newMetrics = new ArrayList<>();
+        for (Metric metric : metricsQuery.getMetrics()) {
+            List<Aggregator> newAggregators = new ArrayList<>();
+            if (!metric.getAggregators().isEmpty() && !(metric.getAggregators().get(0).getName()).equals("merge")
+                    && metric.getAggregators().stream().noneMatch(
+                            aggregator -> NONHISTAGGREGATORS.contains(aggregator.getName()))) {
+                final Optional<Sampling> originalSampling = metric.getAggregators().get(0).getSampling();
```
speezepearson, line 197: Hmm, I think we'll want to copy the sampling from the first aggregator _that specifies a sampling_, rather than just the first aggregator.

Example where this is important: I, a desktop developer, have measured some time in milliseconds, but I want to display it in seconds on Grafana. So I query my metric and use the aggregators
```
[{"name": "div", "divisor": 1000},
 {"name": "percentile", "percentile": 0.75, "sampling": {"value": 1, "unit": "HOURS"}}
]
```
As written, the code will set the merge aggregator's `sampling` to `null`, but we actually want "one hour."

Additionally, we'll want to copy over the `align_sampling` and `align_start_time` fields from that first-aggregator-with-sampling. It's kinda hard to describe what they do ([docs](https://kairosdb.github.io/docs/build/html/restapi/QueryMetrics.html#metric-properties)), but they're conceptually part of `sampling`, and should get copied over along with that field.

app/controllers/KairosDbProxyController.java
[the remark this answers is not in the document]
vjkoskela: We are moving away from that convention as it is not prescribed by standard Oracle/Sun Java. However, it still exists in some projects. In projects that have mixed use I'm okay with either style.

app/controllers/KairosDbProxyController.java
vjkoskela: 👍

test/java/controllers/KairosDbProxyControllerTest.java
```diff
@@ -194,4 +198,50 @@ public void testQueryRequiresAggregatorOff() {
...
+                .setAggregators(ImmutableList.of(
+                        new Aggregator.Builder().setName("filter").build(),
+                        new Aggregator.Builder().setName("max").setSampling(
+                                new Sampling.Builder().setValue(1).setUnit(SamplingUnit.HOURS).build()).build()));
+        final Metric.Builder metric2Builder = new Metric.Builder()
+                .setName("metric2")
+                .setAggregators(ImmutableList.of(new Aggregator.Builder().setName("min").build()));
+        final MetricsQuery.Builder builder = new MetricsQuery.Builder()
+                .setStartTime(Instant.now())
+                .setMetrics(ImmutableList.of(metric1Builder.build(), metric2Builder.build()));
```
speezepearson, line 215: It looks to me like the only way these builders are used is by calling `.build()` on them -- instead of
`final MetricsQuery.Builder builder = (...)`, why not write
`final MetricsQuery query = (...).build()`?

(I'm guessing that you copied this pattern from another test which declared some Builders; that's a reasonable pattern if you want to create several _variations_ on an object, since that lets you say
```java
foo = builder.setName("foo").build();
bar = builder.setName("bar").build();
baz = builder.setName("baz").build();
...
```
but if you're just building a single object, it's clearer to simply build the object up-front.)

test/java/controllers/KairosDbProxyControllerTest.java
```diff
@@ -194,4 +198,50 @@ public void testQueryRequiresAggregatorOff() {
...
+        assertEquals("{\"queries\":[]}", Helpers.contentAsString(result));
+
+        final MetricsQuery newMetricsQuery = _controller.checkAndAddMergeAggregator(builder.build());
+        assertEquals("merge", newMetricsQuery.getMetrics().get(0).getAggregators().get(0).getName());
+        assertEquals("filter", newMetricsQuery.getMetrics().get(0).getAggregators().get(1).getName());
+        assertEquals("max", newMetricsQuery.getMetrics().get(0).getAggregators().get(2).getName());
+        assertEquals(newMetricsQuery.getMetrics().get(0).getAggregators().get(2).getSampling(),
+                newMetricsQuery.getMetrics().get(0).getAggregators().get(0).getSampling());
+        assertEquals("min", newMetricsQuery.getMetrics().get(1).getAggregators().get(0).getName());
+    }
```
speezepearson, line 246: This test seems to be doing two distinct things:
- testing the `checkAndAddMergeAggregator` method
- testing an HTTP request made against the controller

Could we split those into two test-methods? Having more granular tests makes it easier to figure out what's going wrong when something breaks.

test/java/controllers/KairosDbProxyControllerTest.java
abbeyqy: I see your point! But the previous test `testQueryRequiresAggregatorOff` also uses this pattern even though it also builds a single object. If I need to change this, should I change `testQueryRequiresAggregatorOff` as well?

test/java/controllers/KairosDbProxyControllerTest.java
```diff
@@ -220,23 +217,90 @@ public void testAddMergeAggregator() {
...
+                        .setQueries(ImmutableList.of()).build())
+        );
+
+        final Result result = Helpers.invokeWithContext(request, Helpers.contextComponents(), () -> {
+            final CompletionStage<Result> completionStage = controller.queryMetrics();
+            return completionStage.toCompletableFuture().get(10, TimeUnit.SECONDS);
+        });
+
+        assertEquals(Http.Status.OK, result.status());
+        assertEquals("{\"queries\":[]}", Helpers.contentAsString(result));
```
speezepearson, line 282: One last ask, sorry: we now have these 18 lines repeated almost exactly across three different test-methods, differing only in: 
- what argument we're passing to `when(_mockKairosDbClient.queryMetrics(...))`
- what `controller` is being hit by the request

Could we refactor this out into a method? (This might be a great chance to experiment with IntelliJ's "Extract method" functionality! Select a bunch of code, right-click, "Refactor", "Extract method". It makes it really easy to refactor a chunk of code out into its own method.)

app/controllers/KairosDbProxyController.java
```diff
@@ -173,6 +184,35 @@ public KairosDbProxyController(
...
     }
 
+    /* package private */ MetricsQuery checkAndAddMergeAggregator(final MetricsQuery metricsQuery) {
+        final List<Metric> newMetrics = new ArrayList<>();
+        for (final Metric metric : metricsQuery.getMetrics()) {
+            if (needMergeAggregator(metric.getAggregators())) {
+                final List<Aggregator> newAggregators = new ArrayList<>();
+                final Optional<Aggregator> aggregatorWithSampling = metric.getAggregators().stream().filter(
+                        aggregator -> aggregator.getSampling().isPresent()).findFirst();
+                newAggregators.add(aggregatorWithSampling.map(
```
vjkoskela, line 194: If the aggregator has parameters (e.g. percentile value) that don't apply to the merge aggregator, we should probably clear them. My point here, is that I suspect it is better to construct a new MergeAggregator definition than to clone from a random target one just to get its sampling.

test/java/controllers/KairosDbProxyControllerTest.java
```diff
@@ -172,26 +176,102 @@ public void testQueryRequiresAggregatorOff() {
...
+                .setMetrics(ImmutableList.of(metric1Builder.build(), metric2Builder.build()));
+
+        mockQueryMetrics(_controller, builder.build(), builder.setMetrics(ImmutableList.of(metric1Builder.build(),
+                        metric2Builder.setAggregators(ImmutableList.of(
+                                new Aggregator.Builder().setName("merge").build(),
+                                new Aggregator.Builder().setName("filter").build())).build())).build());
+    }
+
+    @Test
+    public void testAddMergeAggregatorOff() {
```
vjkoskela, line 201: Can you include a test around non-merge specific attributes on the top aggregator with sampling do not get attached to the merge aggregator.

app/controllers/KairosDbProxyController.java
```diff
@@ -192,7 +192,12 @@ public KairosDbProxyController(
                 final Optional<Aggregator> aggregatorWithSampling = metric.getAggregators().stream().filter(
                         aggregator -> aggregator.getSampling().isPresent()).findFirst();
                 newAggregators.add(aggregatorWithSampling.map(
-                        aggregator -> ThreadLocalBuilder.clone(aggregator, Aggregator.Builder.class, b -> b.setName("merge"))).
+                        aggregator -> ThreadLocalBuilder.build(Aggregator.Builder.class, b -> b
```
vjkoskela, line 195: This might be easier to read as a full lambda block:

```
ThreadLocalBuilder.build(Aggregator.Builder.class, b -> {
    b.setName("merge");
    aggregator.getSampling.ifPresent(p -> b::setSampling)
     ...
  })
```

PR #391, opened 2020-05-29
remove unused metrics query code and add query format

cwbriones:
MetricsController and the associated view model are not currently used
so I'm opting to rip them out just as with the unused alerting code.
Since the time-range won't be used until execution, we also introduce a
serialization format field to the query to allow for dispatch based on
type.

An enum is being used over a sealed hierarchy because no additional
behavior will be shared among MetricsQuery instances.

This will be immediately used by #389.


app/models/internal/impl/DefaultMetricsQuery.java
```diff
@@ -77,32 +69,19 @@ public Builder setQuery(final String value) {
...
-        public Builder setEnd(final ZonedDateTime value) {
-            _end = value;
-            return this;
-        }
-
-        /**
-         * Sets the start time. Required. Cannot be null.
+         * Sets the format. Required. Cannot be null.
          *
          * @param value the start time
```
cwbriones, line 74: Will fix

cwbriones, line 74: Will fix.

cwbriones:
I'll have to update this now that the JobRepository code landed, it depended on this but I just reverted that change and landed it.

BrandonArp:
It is also my understanding that Smartsheet is running a fork with no intention of changing/upgrading.

cwbriones:
Thanks for confirming @BrandonArp !

PR #393, opened 2020-06-02
Add support for periodCount > 1 in PeriodicSchedule

cwbriones:
Add support for `PeriodicSchedule` instances whose interval is several periods (e.g. 15 minutes instead of just a minute). This only alters the internal model, so for backwards compatibility, the periodCount is set to 1 by default.

I also took this opportunity to refactor the `unboundedNextRun` code. I removed the check to see if the run was before `getRunAtAndAfter` because this is already handled in `BaseSchedule`, and indeed the doc there specifies that subclasses do not need to make this check. This change only a single offset-related test, and I consider that change a bug fix. We were skipping a run that was otherwise valid as it fell between the specified `runAtAndAfter` and `runUntil` values.

### DIFF
--- a/test/java/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicScheduleTest.java
+++ b/test/java/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicScheduleTest.java
@@ -17,19 +17,16 @@ package com.arpnetworking.metrics.portal.scheduling.impl;
 
 import com.arpnetworking.metrics.portal.scheduling.Schedule;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Streams;
 import org.junit.Test;
 
 import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
-import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Optional;
-import java.util.stream.Stream;
 
 import static org.hamcrest.Matchers.equalTo;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


test/java/com/arpnetworking/metrics/portal/scheduling/impl/PeriodicScheduleTest.java
```diff
@@ -87,11 +91,11 @@ public void testNextRunWithUnalignedBounds() {
 
         // typical progression, from lastRun=null to lastRun>runUntil
         assertEquals(
-                Optional.of(Instant.parse("2019-01-02T12:00:00Z")),
+                Optional.of(Instant.parse("2019-01-01T12:00:00Z")),
```
cwbriones, line 94: This is the only test affected by my refactor in `unboundedNextRun`.

To see why I consider this a bug, here's a before and after for computing the initial run.

### Parameters
```
runAtAndAfter: 2019-01-01T06:00:00Z
     runUntil: 2019-01-04T00:00:00Z
       period: 1 day
       offset: 12 hours
```

### Before
1. Last run is empty, so get the value of runAtAndAfter. `2019-01-01T06:00:00Z`
2. Truncate this to the period (1 day). `2019-01-01T00:00:00Z`
3. Is this before runAtAndAfter? (yes) If so, add a period. `2019-01-02T00:00:00Z` (❗️)
4. Add the offset of 12 hours. `2019-01-02T00:12:00Z`
5. **Result** `2019-01-02T00:12:00Z`


### After
1. Last run is empty, so get the value of runAtAndAfter. `2019-01-01T06:00:00Z`
2. Truncate this to the period (1 day). `2019-01-01T00:00:00Z`
3. Add the offset of 12 hours. `2019-01-01T00:12:00Z`
4. **Result** `2019-01-01T00:12:00Z`


So by making the boundary check before adding the offset, we were skipping the run scheduled for `2019-01-01T00:12:00Z` even though it occurred between the configured time range. The change essentially amounts to adding the offset _then_ checking the boundary conditions.

cwbriones: For some more context, the reason I thought this in-depth explanation was necessary was because the test clearly expected the "Before" case and so there could be some disagreement on what the correct behavior is here.

PR #394, opened 2020-06-03
Create a JobRepository wrapper for alerts

cwbriones:
This wrapper takes an AlertRepository and wraps the returned Alert instances in order to implement the Job interface.

Specifically, a new class AlertExecutionContext is introduced (analogous to ReportExecutionContext) that contains all of the specific dependencies and behavior needed for alert evaluation. AlertJob thus acts as a binding between this context and Alert instances - it's an extremely shallow wrapper on its own.

While we could eventually infer longer execution intervals based on the specifics of the alert query, for now we just poll at a fixed interval (30s) regardless of the alert.


test/java/com/arpnetworking/metrics/portal/alerts/AlertExecutionContextTest.java
```diff
@@ -0,0 +1,72 @@
...
+                .setOrganization(organization)
+                .setEnabled(true)
+                .setName("TestAlert")
+                .setDescription("Used in a test.")
+                .setQuery(
+                        new DefaultMetricsQuery.Builder()
+                                .setQuery("Query TBD")
+                                .setStart(ZonedDateTime.now())
+                                .setEnd(ZonedDateTime.now())
+                                .build()
```
cwbriones, line 62: This query is only defined as such because #391 hasn't landed - once that's in place I can remove the `setStart` and `setEnd` calls.

test/java/com/arpnetworking/metrics/portal/alerts/impl/AlertJobRepositoryTest.java
```diff
@@ -0,0 +1,142 @@
...
+                .setOrganization(_organization)
+                .setEnabled(true)
+                .setName("TestAlert")
+                .setDescription("Used in a test.")
+                .setQuery(
+                        new DefaultMetricsQuery.Builder()
+                                .setQuery("Query TBD")
+                                .setStart(ZonedDateTime.now())
+                                .setEnd(ZonedDateTime.now())
+                                .build()
```
cwbriones, line 84: Same as the previous case - this will be changed once the MetricsQuery changes land.

app/com/arpnetworking/metrics/portal/alerts/scheduling/AlertExecutionContext.java
```diff
@@ -0,0 +1,82 @@
...
+        final CompletableFuture<AlertEvaluationResult> future = new CompletableFuture<>();
+        future.completeExceptionally(new UnsupportedOperationException("Alert execution is not implemented"));
+        return future;
+    }
+
+    /**
+     * Get an evaluation schedule for this alert.
+     * <p>
+     * This will attempt to find the largest possible schedule that will still
+     * guarantee alert evaluation will not fall behind.
```
cwbriones, line 72: I'm not sure if I should leave this language in

app/com/arpnetworking/metrics/portal/alerts/scheduling/AlertExecutionContext.java
```diff
@@ -0,0 +1,82 @@
...
+     * Evaluate an alert that was scheduled for the given instant.
+     *
+     * @param alert The alert to evaluate.
+     * @param scheduled The scheduled evaluation time.
+     * @return A completion stage containing {@code AlertEvaluationResult}.
+     */
+    public CompletionStage<AlertEvaluationResult> execute(final Alert alert, final Instant scheduled) {
+        final CompletableFuture<AlertEvaluationResult> future = new CompletableFuture<>();
+        future.completeExceptionally(new UnsupportedOperationException("Alert execution is not implemented"));
+        return future;
```
wittekm, line 65: man, jdk has come a long way since 1.6, this api is actually great

app/global/MainModule.java
```diff
@@ -207,6 +213,20 @@ protected void configure() {
                 .asEagerSingleton();
     }
 
+    @Singleton
+    @Provides
+    @SuppressFBWarnings("UPM_UNCALLED_PRIVATE_METHOD") // Invoked reflectively by Guice
```
wittekm, line 218: disclaimer: I'm unsure what the behavior here will be

would this result in "spiky" cpu usage? that is to say, if we ran every 30 minutes, would it look something like the following timeline?
```
[alert schedule starts] [5 minutes of CPU at 100%] [25 minutes of CPU at ~0%]
```

and, if so, I don't even know if that's a problem

but what i'm getting at is, maybe we want to introduce some like noise/stutter and spread this out?

again, to reiterate: I Have No Idea If This Is A Thing To Be Concerned About (tm) (copyright)

conf/portal.application.conf
```diff
@@ -205,6 +205,10 @@ reporting {
   }
 }
 
+alerting {
+  execution.defaultInterval = 30s
+}
+
```
wittekm, line 211: purely out of curiosity, do we have a mechanism to detect the case where:

we have 40 seconds of work we're trying to schedule every 30 seconds

?

test/java/com/arpnetworking/metrics/portal/alerts/AlertExecutionContextTest.java
wittekm: ack

test/java/com/arpnetworking/metrics/portal/alerts/impl/AlertJobRepositoryTest.java
```diff
@@ -0,0 +1,142 @@
...
+    @Test
+    public void testJobQueryProxiesParametersAndResults() {
+        // Arbitrary since we're just testing proxying logic
+        final long alertTotal = 42;
+        final int offset = 7;
+        final int limit = 100;
+
+        // mock AlertRepository such that for the above query params it returns
+        // a page with a single alert.
+        final ArgumentCaptor<AlertQuery> captor = ArgumentCaptor.forClass(AlertQuery.class);
```
wittekm, line 120: extremely cool class name. ArgumentCaptor.

conf/portal.application.conf
cwbriones: We don't have any proactive mechanism, no. The current use case is reporting and runs significantly less frequently than alerting, at most daily, and so it would likely never run into this.

This is an interesting question though and we might want to force a timeout on jobs to make sure this doesn't happen. I definitely intend to instrument any scheduling lag (e.g. ∆t_lag = scheduled - now < 0) if it isn't already present.

app/com/arpnetworking/metrics/portal/alerts/scheduling/AlertExecutionContext.java
cwbriones: the QoL improvements are really great, and they're even better in 11 once we can get on that!

app/global/MainModule.java
cwbriones: I definitely expect some sort of spikiness, we already see something like this with rollups when the hour/day rolls over. I don't think this will necessarily be an issue out the gate, but @vjkoskela and I had discussed this as another motivating factor for query caching which I believe is already on our roadmap.

As for any noise/jittering - I think I'd like to see this in the [JobCoordinator] instead of dealing with it at the per-job level. At the least we may want to introduce some form of queueing since I think right now we just spawn an actor for [every possible job]. This would broaden any spikes not by messing with scheduling but by limiting the number of concurrently executing jobs. For alerts this should already be handled by limits on [the query executor], but it would still be nice to have this guarantee more generally.

[JobCoordinator]: https://github.com/ArpNetworking/metrics-portal/blob/7b51cca93cfd1d9ebc0acba83b2a76942aa37862/app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
[the query executor]: https://github.com/ArpNetworking/metrics-portal/blob/e1466b6fb4ef5602caa25842248cca9793925fb6/app/global/MainModule.java#L353
[every possible job]: https://github.com/ArpNetworking/metrics-portal/blob/04921f3c7e1980759815e04c90bd9d0e3a3c0b34/app/com/arpnetworking/kairos/service/KairosDbServiceImpl.java#L66

PR #395, opened 2020-06-05
Measure job execution lag

cwbriones:
This is the time between when the job was scheduled vs. when it actually
executed. So for example, if a job was scheduled for `16:00:00` and runs at `16:00:42`, then the lag is 42 seconds.


cwbriones:
A metric like this was discussed in https://github.com/ArpNetworking/metrics-portal/pull/394#discussion_r434865947 and it was straightforward so here it is!

PR #398, opened 2020-06-08
Periodically create partitions for alert executions

cwbriones:
Introduce a DailyPartition manager actor that runs within the `DatabaseAlertExecutionRepository`, periodically attempting to create partitions for the alert executions table that was added in #380. Notably I've introduced another `EbeanServer` which uses the DDL role (`metrics_dba`), since that Role is necessary to create tables.

I've included unit tests for the actor while the DB testing is covered by the existing integration tests.


BrandonArp:
Have we setup the Akka pools for these actors?  You're updating connection pool sizes, but if the Akka actor's thread pools aren't big enough (they usually assume non-blocking), then you're gonna block before the connection open.

cwbriones:
Is the concern because I block on actor start whereas we ordinarily would let that happen asynchronously? I'm not sure I entirely understand the issue.

pom.xml
```diff
@@ -1165,7 +1165,6 @@
       <groupId>org.flywaydb</groupId>
       <artifactId>flyway-play_${scala.package.version}</artifactId>
       <version>${flyway.play.version}</version>
-      <scope>runtime</scope>
```
cwbriones, line 1168: Scope changed because I need to block on the flyway-play `PlayInitializer` from within `MainModule` to guarantee that migrations have run.

main/postgres/initdb.d/init.sql.1
```diff
@@ -27,7 +27,7 @@ ALTER ROLE metrics_app WITH NOSUPERUSER INHERIT NOCREATEDB NOCREATEROLE NOREPLIC
 
 CREATE ROLE metrics_dba LOGIN;
 ALTER ROLE metrics_dba WITH PASSWORD 'metrics_dba_password';
-ALTER ROLE metrics_dba WITH NOSUPERUSER INHERIT NOCREATEDB NOCREATEROLE NOREPLICATION CONNECTION LIMIT 6;
+ALTER ROLE metrics_dba WITH NOSUPERUSER INHERIT NOCREATEDB NOCREATEROLE NOREPLICATION CONNECTION LIMIT 15;
```
cwbriones, line 30: Why 15:

I bumped the connection pool to be the same size as what we have for `metrics_dml`, and set the size to be the same in tests. That is `5 + 5 = 10` connections. There's also an additional connection flyway creates with this role. That's a total of 11. I then just bumped it to 15.

conf/db/migration/metrics_portal_ddl/V22__replace_alert_execution_function.sql
```diff
@@ -0,0 +1,49 @@
...
+--     Example for portal.alert_executions, 8 May 2020:
+--
+--     CREATE TABLE portal.alert_executions_2020_05_08 PARTITION OF portal.alert_executions
+--            FOR VALUES FROM ('2020-05-08') TO ('2020-05-09');
+--
+-- Params:
+--    Table - Text - The name of the parent table.
+--    Start - Date - The beginning date of the time range, inclusive.
+--    End   - Date - The end date of the time range, exclusive.
+CREATE OR REPLACE FUNCTION create_daily_partition(schema TEXT, tablename TEXT, start_date DATE, end_date DATE)
```
cwbriones, line 26: Content is exactly the same as the previous function, just formatted in a nicer way and with an added schema parameter

BrandonArp:
The concern regarding thread pools is because you seem to need to adjust the connection pools.  If we're now seeing enough database traffic to warrant tuning the connection pools, we'll also need to tweak the thread pools (requests are served on Akka actors).  It hadn't been a problem up until now so we've just kinda ignored it.  But it seemed like you were hitting some limitation.  And I just wanted to call out that just bumping the connection pool size would likely be insufficient.

vjkoskela:
I *believe* the issue here is that the code is now using the DDL connection pool to manage the partition tables. Previously, this connection pool was only used by Flyway and tuned to prevent accidental usage (by limiting its size). Now that we are intentionally using this pool for dynamically extending the partitions, the default pool size needs to be increased.

cwbriones:
> Now that we are intentionally using this pool for dynamically extending the partitions, the default pool size needs to be increased.

Yes, this was my reasoning for increasing the connection pool size. It is likely overkill to have it the same size as the DML pool since the only additional traffic will be from the single partition-creating actor. The request pool was unchanged because the additional traffic is from an internal actor rather than generated via external requests.

cwbriones:
Some additional context that I provided in slack:

> The context is that when we execute jobs within metrics portal, we store the results in some execution table. For alerts this was added in #380 . Since we expect alerts to evaluate twice a minute, thats ~3000 executions per alert per day. Because of daily volume and the fact that we really only care about alert executions in the current day, the table is partitioned by day - but Postgres does not automatically create these tables when the day rolls over.
>
> So this PR adds an actor that periodically wakes up and creates the tables if they don’t exist at some scheduled time.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -0,0 +1,283 @@
...
+import java.util.concurrent.ExecutionException;
+import javax.persistence.PersistenceException;
+
+/**
+ * An actor that will periodically create table partitions.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public class DailyPartitionCreator extends AbstractActorWithTimers {
+    /* package private */ static final Object TICK = new Object();
```
vjkoskela, line 50: I assume these are messages? Typically I believe we use string constants. Not sure if the recommendation on that has changed.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -0,0 +1,283 @@
...
+    private static final Object START_TICKING = new Object();
+    private static final Object EXECUTE = new Object();
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DailyPartitionCreator.class);
+    private static final Duration TICK_INTERVAL = Duration.ofMinutes(1);
+    private static final String TICKER_NAME = "PERIODIC_TICK";
+    private final EbeanServer _ebeanServer;
```
vjkoskela, line 57: nit: blank line between statics and non-statics

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -0,0 +1,283 @@
...
+     * <p>
+     * This will execute partition creation exactly once and then periodically thereafter using the actor's props.
+     *
+     * @param ref an {@code DailyPartitionCreator}.
+     * @param timeout timeout for the operation
+     * @throws ExecutionException if an exception was thrown during execution.
+     * @throws InterruptedException if the actor does not reply within the allotted timeout, or if the actor thread was
+     * interrupted for other reasons.
+     */
+    public static void start(
```
vjkoskela, line 145: Is there a reason that Actor must be explicitly started as opposed to just starting on the Akka onStart method?

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    private void recordCounter(final String metricName, final long value) {
+        final String fullMetric = String.format("partition_creator/%s/%s", _table, metricName);
+        _periodicMetrics.recordCounter(fullMetric, value);
+    }
+
+    // Message handlers
+
+    private void startTicking() {
```
vjkoskela, line 204: Moving this to onStart would remove the statefulness of the actor entirely; e.g. if it exists then it's running

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -0,0 +1,283 @@
...
+
+    /**
+     * Create a series of daily partitions for the given parameters.
+     *
+     * @param schema the database schema
+     * @param table the parent table
+     * @param startDate the start date, inclusive.
+     * @param endDate the end date, exclusive.
+     */
+    protected void execute(final String schema,
```
vjkoskela, line 267: nit: If we wrap one, then we wrap all.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -0,0 +1,283 @@
...
+                .addData("endDate", endDate)
+                .log();
+
+        Optional<Exception> error = Optional.empty();
+        try {
+            execute(_schema, _table, startDate, endDate);
+            _lastRun = Optional.of(_clock.instant());
+        } catch (final PersistenceException e) {
+            error = Optional.of(e);
+            LOGGER.error().setMessage("Failed to create daily partitions for table")
```
vjkoskela, line 247: nit: `.setMessage` on next line.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+
+    // Wrapper to propagate any errors that occurred to the caller.
+    // This is really only useful on a call to `start`.
+    private Optional<Exception> execute() {
+        final LocalDate startDate = ZonedDateTime.ofInstant(_clock.instant(), _clock.getZone()).toLocalDate();
+        final LocalDate endDate = startDate.plusDays(_lookahead);
+
+        LOGGER.info().setMessage("Creating daily partitions for table")
```
vjkoskela, line 234: nit: .setMessage on next line.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -0,0 +1,283 @@
...
+            error = Optional.of(e);
+            LOGGER.error().setMessage("Failed to create daily partitions for table")
+                    .addData("schema", _schema)
+                    .addData("table", _table)
+                    .addData("startDate", startDate)
+                    .addData("endDate", endDate)
+                    .setThrowable(e)
+                    .log();
+        }
+        recordCounter("create", error.isPresent() ? 0 : 1);
```
vjkoskela, line 255: This will not be executed at all if there is a runtime exception.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -0,0 +1,283 @@
...
+        // it still returns a "non-empty" void result and so we can't use the
+        // ordinarily more appropriate SqlUpdate type.
+        final SqlQuery sql = _ebeanServer.createSqlQuery(
+                "select * from create_daily_partition(?::text, ?::text, ?::date, ?::date)")
+                .setParameter(1, schema)
+                .setParameter(2, table)
+                .setParameter(3, startDate)
+                .setParameter(4, endDate);
+
+        sql.findOneOrEmpty().orElseThrow(() -> new IllegalStateException("Expected a single empty result."));
```
vjkoskela, line 281: This is a top level message handler. Throwing the exception here will just result in an uncaught message handler exception in Akka. Seems like we could do better handling, logging and instrumenting this failure. Thoughts?

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertExecutionRepository.java
```diff
@@ -92,13 +124,26 @@ private AlertExecution findOrCreateAlertExecution(
...
+            throw new RuntimeException("Failed to start partition creator", e);
+        }
         _isOpen.set(true);
     }
 
     @Override
     public void close() {
         assertIsOpen();
         LOGGER.debug().setMessage("Closing DatabaseAlertExecutionRepository").log();
+        assert _partitionCreator != null : "partitionCreator should be non-null when open";
```
vjkoskela, line 140: We generally avoid the use of assert since it's conditional based on JVM arguments

main/postgres/initdb.d/init.sql.1
vjkoskela: nit: Flyway creates (for some reason) 2 connections.

cwbriones: Any change required here then? that would be 12 which is still below the 15 limit.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: You mean the signature?

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: I originally had it started via `onStart` but did it this way so that I could synchronously start when the repository was opened, then any immediate errors when attempting to create the initial partition will bubble up to `Repository#open` instead of happening in the background.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: (marking as resolved since this outcome is tied to the other comment, but feedback makes sense)

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: No preference here so I'll change it - I just saw Objects used elsewhere and took that as the convention.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: meant to put this in a finally 😄

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: This is not the top-level handler - the actual handler is the `execute` overload with no parameters. This has the actual DB logic but the other code adds logging/metrics.

It is the case though that the wrapper code will not handle this illegal state exception which would still cause the problem you're mentioning albeit for different reasons so I could probably handle that.

app/global/MainModule.java
```diff
@@ -441,6 +473,23 @@ public EbeanServer get() {
         }
     }
 
+    private static final class MetricsPortalDDLEbeanServerProvider implements Provider<EbeanServer> {
```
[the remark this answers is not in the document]
cwbriones: Yes it's purely stylistic in this case, the behavior would be identical. I wrote a SIC for this to keep it consistent with the other EBeanServer provider.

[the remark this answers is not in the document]
cwbriones: Generally though I do prefer `@Provides` methods because there's less boilerplate.

main/postgres/initdb.d/init.sql.1
vjkoskela: Nope, just trying to spread the knowledge.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -63,6 +64,8 @@
     private final Schedule _schedule;
     private Optional<Instant> _lastRun;
 
+    private Set<LocalDate> _partitionCache;
```
vjkoskela, line 67: nit: `final`

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -131,49 +135,59 @@ public static Props props(
...
-                    @SuppressWarnings("unchecked") final Optional<Throwable> o = (Optional<Throwable>) reply;
-                    final CompletableFuture<Void> future = new CompletableFuture<>();
-                    o.ifPresent(future::completeExceptionally);
-                    future.complete(null);
-                    return future;
-                })
-                .toCompletableFuture()
-                .get();
+    public static void stop(final ActorRef ref, final Duration timeout) throws ExecutionException, InterruptedException {
+        Patterns.gracefulStop(ref, timeout).toCompletableFuture().get();
```
vjkoskela, line 147: Is this worth the method? Especially, since it actually has no logic specific to this actor. Moreover, you could use `PatternCS` (assuming we're on a new enough version of Akka in MP) which should simplify the call to stop (I _believe_).

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -131,49 +135,59 @@ public static Props props(
...
+     * @param date The partition date
      * @param timeout timeout for the operation
      * @throws ExecutionException if an exception was thrown during execution.
-     * @throws InterruptedException if the actor does not stop within the allotted timeout, or if the actor thread was
+     * @throws InterruptedException if the actor does not respond within the allotted timeout, or if the actor thread was
      * interrupted for other reasons.
      */
-    public static void stop(final ActorRef ref, final Duration timeout) throws ExecutionException, InterruptedException {
-        Patterns.gracefulStop(ref, timeout).toCompletableFuture().get();
+    public static void ensurePartitionExistsForDate(
```
vjkoskela, line 161: Not entirely without merit, and not as concerned, but at least may be simpler with `PatternsCS`. I believe we also doubled up on the timeouts with this pattern in MAD and CAGG; e.g.

https://github.com/ArpNetworking/metrics-aggregator-daemon/commit/158884ec04cc53f5b7b6dc705e78cd1ffab437c4#diff-fed66f6be12dfd97b6fb62212dd0d6ebR52

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -131,49 +135,59 @@ public static Props props(
...
 
+    @Override
+    public void preStart() {
+        LOGGER.info().setMessage("Starting execution timer")
+                .addData("schema", _schema)
+                .addData("table", _table)
+                .addData("lookahead", _lookahead)
+                .log();
+        getSelf().tell(TICK, getSelf());
+        getTimers().startPeriodicTimer(TICKER_NAME, TICK, TICK_INTERVAL);
```
vjkoskela, line 187: I assume this gets cancelled by the base class `AbstractActorWithTimers`?

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -187,9 +201,13 @@ public void postStop() throws Exception {
     @Override
     public Receive createReceive() {
         return new ReceiveBuilder()
-                .matchEquals(START_TICKING, msg -> startTicking())
                 .matchEquals(TICK, msg -> tick())
-                .matchEquals(EXECUTE, msg -> execute())
+                .match(CreateForRange.class, msg -> {
+                    final Status.Status resp = execute(msg.getStart(), msg.getEnd());
+                    if (!getSender().equals(getSelf())) {
```
vjkoskela, line 207: Hrm. Is this what you were looking for?

https://doc.akka.io/docs/akka/current/typed/interaction-patterns.html

See: "Ignoring replies"

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -200,35 +218,42 @@ private void recordCounter(final String metricName, final long value) {
...
-    }
-
     private void tick() {
         recordCounter("tick", 1);
 
         final Instant now = _clock.instant();
         if (_schedule.nextRun(_lastRun).map(run -> run.isBefore(now)).orElse(true)) {
-            getSelf().tell(EXECUTE, getSelf());
+            final LocalDate startDate = ZonedDateTime.ofInstant(now, _clock.getZone()).toLocalDate();
+            final LocalDate endDate = startDate.plusDays(_lookahead);
```
vjkoskela, line 227: nit: rename `_lookahead` to `_lookAheadDays`

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -200,35 +218,42 @@ private void recordCounter(final String metricName, final long value) {
...
-        getSender().tell(executeResult, getSelf());
-    }
-
     private void tick() {
         recordCounter("tick", 1);
 
         final Instant now = _clock.instant();
         if (_schedule.nextRun(_lastRun).map(run -> run.isBefore(now)).orElse(true)) {
-            getSelf().tell(EXECUTE, getSelf());
+            final LocalDate startDate = ZonedDateTime.ofInstant(now, _clock.getZone()).toLocalDate();
```
vjkoskela, line 226: Do we really need a local timezone here? What happens if two boxes have different timezones? Would this work equally well just fixed to UTC?

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -253,9 +279,17 @@ private void tick() {
...
         } finally {
-            recordCounter("create", error.isPresent() ? 0 : 1);
+            recordCounter("create", status instanceof Status.Success ? 0 : 1);
+        }
+        return status;
+    }
+
+    private void updateCache(final LocalDate start, final LocalDate end) {
+        LocalDate date = start;
+        while (!date.equals(end)) {
```
vjkoskela, line 289: As these are local dates, I believe they won't be equal if they have different locales (zones) even if adjusted it's the same point in time. Is this what we want for comparison? (related: see comment about zones above)

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: yes they're lifecycle bound

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: This is not a local time zone, it is UTC regardless of host. I initialized the clock to `Clock.systemUTC()`. I can be explicit here and just specify UTC a second time.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: If I understand the concern, it's not applicable here because LocalDate instances do not have timezone information - it's essentially just day/month/year

cwbriones: (I'm treating all time in here as UTC which is why the zone offset information was dropped)

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: I can inline this; I had a method to be consistent with the `start` call but since that was removed there's no need for the consistency.

As far as the APIs used, PatternsCS and Patterns were merged in Akka 2.5.19 (we're running 2.5.20).

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: As mentioned above PatternsCS and Patterns are essentially the same API for this version of Akka - but I'll add the timeouts. I'm pretty sure I'm the one that asked you to double up on that code to begin with 😄

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: I just looked and MAD is running Akka 2.5.16 so it makes sense that using Patterns works here but not over there.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
cwbriones: This is exactly what I was looking for, but that's Akka `2.6.XX` using typed actors so it's not available here 😢

PR #405, opened 2020-06-12
KairosDB Query executor

cwbriones:
Implementation of `QueryExecutor` for KairosDB queries.

Part 1/2 for alert evaluation. I then need to wire this into `AlertExecutionContext` and interpret the result.


app/com/arpnetworking/metrics/portal/query/QueryExecutor.java
```diff
@@ -31,7 +34,6 @@
      *
      * @param query query to execute
      * @return {@link CompletionStage} of the result
-     * @throws QueryExecutionException on errors during query execution
      */
-    CompletionStage<MetricsQueryResult> executeQuery(MetricsQuery query) throws QueryExecutionException;
```
cwbriones, line 36: I removed this exception from the signature because it didn't make sense to me to have a function return a `CompletableFuture` and also throw an exception.

IMO that error, if it occurs, should happen asynchronously anyway and therefore should be handled via an `exceptionally` combinator from within the `CompetionStage`.

cwbriones:
Leaving this as a draft until #411 lands.

app/com/arpnetworking/metrics/portal/query/impl/KairosDbQueryExecutor.java
```diff
@@ -0,0 +1,182 @@
...
+     */
+    @Inject
+    public KairosDbQueryExecutor(final KairosDbService service, final ObjectMapper objectMapper) {
+        _service = service;
+        _objectMapper = objectMapper;
+    }
+
+    @Override
+    public CompletionStage<MetricsQueryResult> executeQuery(final BoundedMetricsQuery query) {
+        return CompletableFuture.completedFuture(query).thenCompose(this::executeQueryInner);
```
speezepearson, line 68: I'd have thought that `CompletableFuture.completedFuture(x).thenCompose(y)` was semantically identical to `y(x)` -- what's the difference?

app/com/arpnetworking/metrics/portal/query/impl/KairosDbQueryExecutor.java
```diff
@@ -0,0 +1,182 @@
...
+                .min(ChronoUnit::compareTo);
+    }
+
+    private Optional<ChronoUnit> periodHint(final Metric metric) {
+        final List<Aggregator> aggregators = metric.getAggregators();
+        if (aggregators.isEmpty()) {
+            return Optional.empty();
+        }
+        // For the purposes of periodic querying, the only aggregator size that matters
+        // is the final aggregator with a sampling parameter.
```
speezepearson, line 95: Huh, I'd have guessed it would be the _first_ sampling aggregator. Or... maybe even that coarsest of _any_ sampling aggregator?

app/com/arpnetworking/metrics/portal/query/impl/KairosDbQueryExecutor.java
cwbriones: There's a slight difference in error-handling semantics, although I'm unsure if there's a more explicit way to achieve the same effect.

The difference is that in the former, if y(x) throws an exception it is thrown "in the monad" so that the entire function returns a failed future containing said exception. If you instead called y(x) directly and it threw, that exception would be on the current thread of execution. I could get the same exception wrapping effect manually by putting a try/catch around y(x), performing the wrapping myself, but this current setup handles all of that for me behind the scenes.

i.e. the equivalent would be
```java
try {
    return executeQueryInner(query);
} catch (final Exception e) {
    final CompletableFuture<MetricsQueryResponse> future = new CompletableFuture<>();
    future.completeExceptionally(e);
    return future;
}
```

I noticed there's [CompletableFuture#supplyAsync](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#supplyAsync-java.util.function.Supplier-) but that runs it by default in the global fork-join pool. There's a separate executor argument you can specify but at that point I wouldn't know where to pull that from since ideally it would run in the same pool as the query itself.

app/com/arpnetworking/metrics/portal/query/impl/KairosDbQueryExecutor.java
cwbriones: I've been playing around with this a bit more and I think you're correct. I'll update.

app/com/arpnetworking/metrics/portal/query/impl/KairosDbQueryExecutor.java
cwbriones: Future<T, E> 
Result<Future<T, E>, E>

cwbriones: `supplyAsync` doesn't work because I want `executeQueryInner` to return a `CompletionStage`. I just rewrote it with the explicit try-catch.

speezepearson: Ahhh, I understand now! (I suspect that, if you don't throw in a comment or a test for this, somebody will accidentally simplify this feature away. But it doesn't seem like a huge deal.)

cwbriones: There's a test that catches this already!

PR #409, opened 2020-06-17
rollups: refactor out computation of time intervals to roll up

speezepearson:


app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -295,36 +295,23 @@ private void handleLastDataPointMessage(final LastDataPointsMessage message) {
...
-                    final RollupDefinition defn = rollupDefBuilder.setStartTime(startTime).build();
-                    _rollupManagerPool.tell(defn, self());
-                    LOGGER.debug()
-                            .setMessage("sent task to _rollupManagerPool")
-                            .addData("task", defn)
-                            .addData("pool", _rollupManagerPool)
-                            .log();
-                    _metrics.recordCounter("rollup/generator/task_sent", 1);
-                }
+            final SortedSet<Instant> startTimes = getRollupTimes(message, period);
```
cwbriones, line 298: (Since you're already in here)

Is the only reason this is a sorted set to guarantee iteration order? I would guess that the goal is to ensure the backfill goes from oldest -> newest but otherwise it doesn't look like the fact that it's sorted is used anywhere else.

If that's the case I would just have this return a `Stream` (using `Stream.iterate`) or `Iterable` instead of a collection, since it's not important to actually store all of the periods in memory or even to have that collection be sorted under all operations.

speezepearson: I like the no-duplication guarantee of a Set!

speezepearson: And I don't think the simultaneous-loading-into-memory is a big deal: even 1 year is only ~10k hours, and all those RollupDefinitions are going to be loaded into memory anyway (in the recipient's mailbox).

speezepearson: Realized I'm bikeshedding, went to implement this; `Stream.iterate` is neat, but I can't think how to make it work well here without a `takeWhile`.

cwbriones: > without a takeWhile.

I actually had that same problem earlier today 😢 .

But fair enough! I'm less concerned about the memory usage and more about conveying intent to the reader through narrower types. Using less-specific types makes it clear what behavior you actually depend on.

speezepearson: That's very fair! I think I'm convinced enough to do it your way... next time ;)

PR #411, opened 2020-06-18
QueryExecutor RFC: Time-bounded queries and delegated executor

cwbriones:
I've isolated these changes from the Alert execution code I've written, since there were a few things necessary to get that into a working state.

Changes Introduced

* We need a notion of time range on the query model used for the executor. This was previously in the class `MetricsQuery`, but instead I've introduced a separate `BoundedMetricsQuery` that keeps all the same fields along with a time range. This distinction is necessary because alert query definitions do not need a time range specified since it's inferred at execution time based on the scheduler and period hint (see next bullet).

* executors can now specify a "periodHint" via `QueryExecutor#periodHint`. This is *necessary* for alert evaluation to ensure we don't miss any gaps in a query. For example, an alert query can be aggregated hourly, and if we polled minutely, we will miss the last data point most of the time and think the alert has stopped firing.

* `QueryExecutorRegistry` and `DelegatingQueryExecutor` - these go hand-in-hand. I've essentially replaced arbitrary registration via names by a more static registration using queryFormat. We need this to delegate `MetricsQuery` instances to the correct executor. Example: `QueryFormat.KAIROS_DB -> KairosDbQueryExecutor`

### DIFF
--- a/app/com/arpnetworking/metrics/portal/query/impl/NoQueryExecutor.java
+++ b/app/com/arpnetworking/metrics/portal/query/impl/NoQueryExecutor.java
@@ -18,10 +18,11 @@ package com.arpnetworking.metrics.portal.query.impl;
 import com.arpnetworking.metrics.portal.query.QueryExecutionException;
 import com.arpnetworking.metrics.portal.query.QueryExecutor;
 import com.google.common.collect.ImmutableList;
-import models.internal.MetricsQuery;
+import models.internal.BoundedMetricsQuery;
 import models.internal.MetricsQueryResult;
 import models.internal.Problem;
 
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import javax.inject.Singleton;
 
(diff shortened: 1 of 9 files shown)


app/com/arpnetworking/metrics/portal/query/impl/NoQueryExecutor.java
```diff
@@ -33,8 +34,10 @@
 @Singleton
 public class NoQueryExecutor implements QueryExecutor {
     @Override
-    public CompletionStage<MetricsQueryResult> executeQuery(final MetricsQuery query) throws QueryExecutionException {
```
cwbriones, line 36: Smaller change but still relevant - I removed the checked exception because IMO it doesn't make sense to both return a completion stage and possibly throw an exception in the main thread. If there are any query errors, they should be thrown in the CompletionStage monad and handled later via an `exceptionally` combinator or `whenComplete`.

PR #412, opened 2020-06-19
Add Jackson annotations to the Query interfaces

cwbriones:
I'd like to include a bunch of testcases in a different diff as JSON files and this makes that possible. There were some annotations before but Jackson cannot resolve the interfaces without encoded type information. Since most of these types have a backing `DefaultXXX` class, I included those as the default deserialize types.

### DIFF
--- a/app/models/internal/TimeSeriesResult.java
+++ b/app/models/internal/TimeSeriesResult.java
@@ -15,7 +15,6 @@
  */
 package models.internal;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
@@ -41,7 +40,6 @@ public interface TimeSeriesResult {
      * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
      */
     interface Query {
-        @JsonProperty("sample_size")
         long getSampleSize();
 
         ImmutableList<? extends Result> getResults();
--- a/app/models/view/TimeSeriesResult.java
+++ b/app/models/view/TimeSeriesResult.java
@@ -19,7 +19,6 @@ import com.arpnetworking.commons.builder.OvalBuilder;
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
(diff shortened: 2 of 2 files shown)


app/models/internal/TimeSeriesResult.java
```diff
@@ -40,8 +53,18 @@
...
+            property = "@type"
+    )
+    @JsonSubTypes(
+            @JsonSubTypes.Type(
+                    name = "default",
+                    value = DefaultTimeSeriesResult.Query.class
+            )
+    )
     interface Query {
-        @JsonProperty("sample_size")
```
cwbriones, line 44: I believe this was as it is to deserialize a kairosDB query directly into this model.

That mapping is handled manually in #405.

app/models/internal/TimeSeriesResult.java
```diff
@@ -69,15 +103,39 @@
      *
      * @author Brandon Arp (brandon dot arp at inscopemetrics dot com)
      */
-    interface QueryGroupBy {
-    }
+    @JsonTypeInfo(
+            use = JsonTypeInfo.Id.NAME,
+            property = "name"
+    )
+    @JsonSubTypes({
```
cwbriones, line 110: These subtypes aren't as flexible as I'd like, but I can easily workaround this with a custom `Deserializer` class.

Ideally we would want to delegate each derived interface to the annotations on that interface, but Jackson doesn't resolve this recursively. This would be like a tree of `JsonTypeInfo`.

cwbriones:
The only issue I can see with this is perhaps having JSON annotations on an internal model making it a bit confusing from the perspective of the reader (since only view or data models tend to be serialized).

speezepearson:
Hmm... why do these internal models need to be deserialized directly, rather than going through the view models? It should be equivalent, no?

cwbriones:
> Hmm... why do these internal models need to be deserialized directly, rather than going through the view models? It should be equivalent, no?

There are no view models for this API since the (generic) metrics query models are all internal right now. This is not the KairosDB model despite looking mostly the same.

speezepearson:
> There are no view models for this API

Wait, there aren't?
- `MetricsQueryResult`: [view](https://github.com/ArpNetworking/metrics-portal/blob/master/app/models/view/MetricsQueryResult.java), [internal](https://github.com/ArpNetworking/metrics-portal/blob/master/app/models/internal/MetricsQueryResult.java)
- `TimeSeriesResult`: [view](https://github.com/ArpNetworking/metrics-portal/blob/master/app/models/view/TimeSeriesResult.java), [internal](https://github.com/ArpNetworking/metrics-portal/blob/master/app/models/internal/TimeSeriesResult.java)

cwbriones:
# 🤦 

You were totally right and I just missed these. I went ahead and added the necessary annotations to the view model (QueryGroupBy) and reverted the others.

PR #413, opened 2020-06-19
Return an informative error if no JSON can be found in the request body (instead of a NullPointerException).

orborde:


ddimensia:
Checkstyle errors, otherwise looks fine.

PR #421, opened 2020-06-22
reduce Ebean admin pool size

speezepearson:
Context: https://github.com/ArpNetworking/metrics-portal/pull/398 bumped up this connection-pool size from 2 to 5, probably prematurely, and this has been smashing us into a connection limit. From my DMs with @cwbriones : 
> spencerpearson  3:53 PM
> Ah... but why haven't we always been hitting [the connection limit], then?
> 3:53
> For us to not be hitting this already, we'd need at most (2+2)*3 connections.
> 
> cbriones  3:53 PM
> it used to be 2, it got bumped to 5
> 
> spencerpearson  3:53 PM
> Aha!
> 
> cbriones  3:55 PM
> honestly the bump to 5 is probably unnecessary since only one partition creator is ever running right now - I think the decision to increase it was made prematurely
> 
> spencerpearson  3:56 PM
> All righty. Mind if I bump it back down to 2?
> 
> cbriones  3:57 PM
> I would bump it back down to 2, if it causes issues then we know we really do need the extra connections and we can do the increase then
>


cwbriones:
You can likely lower the `init` script limit if the runtime pools are going to be lowered https://github.com/ArpNetworking/metrics-portal/blob/master/main/postgres/initdb.d/init.sql.1#L30

It was 6 previously, I think that should still be OK since you have, at most
* 2 flyway connections
* 2 test ebean server connections
* 2 normal ebean server connections (integration tests)
* Total: 6, as before

cwbriones:
Thanks for handling this.

PR #422, opened 2020-06-23
Implement Alert Execution

cwbriones:


app/com/arpnetworking/metrics/portal/alerts/scheduling/AlertExecutionContext.java
```diff
@@ -79,4 +90,153 @@ public AlertExecutionContext(
...
+                .map(g -> (TimeSeriesResult.QueryTagGroupBy) g)
+                .findFirst()
+                .map(TimeSeriesResult.QueryTagGroupBy::getGroup);
+    }
+
+    private BoundedMetricsQuery applyTimeRange(final MetricsQuery query,
+                                               final Instant scheduled,
+                                               final ChronoUnit period) {
+        // We must truncate to avoid dealing with partially aggregated periods.
+        final ZonedDateTime latest = ZonedDateTime.ofInstant(scheduled, ZoneOffset.UTC).truncatedTo(period);
```
speezepearson, line 222: We want this end-time to be _exclusive_, right? IIRC, KairosDB's semantics are that `end_time` is _inclusive_, so I think we want to subtract 1ns or 1ms or something from this value. We could do that here, if we want the alerting code to have the convention that endTime is inclusive; or [here](https://github.com/ArpNetworking/metrics-portal/blob/6fe9fe26d2cfced47174ba05095a699674ee3734/app/com/arpnetworking/metrics/portal/query/impl/KairosDbQueryExecutor.java#L113), if we want the convention that it's exclusive.

app/models/internal/alerts/AlertEvaluationResult.java
```diff
@@ -43,13 +44,23 @@
...
-    // XXX(cbriones): This is likely to change once we actually integrate alerts with the job scheduling code.
+    /**
+     * The series name at the time of evaluation.
+     * <p>
+     * This can be used to distinguish cases where the underlying query changed
+     * between successive evaluations.
+     *
+     * @return the series name.
+     */
+    String getSeriesName();
```
speezepearson, line 55: IIUC: we're worried about getting a result back and being confused about what it's a result _for_, because the alert changed in the meantime?

If that's the case: have you considered sticking the entire query in here, rather than just the series-name? That would seem maximally informative about what actually got run.
I don't have a strong preference here, though -- I don't expect it to matter much in practice.

cwbriones: Yes that's correct. I suppose there could be other ambiguous cases as well (what's the series name for an arithmetic query?). But like you mention I don't expect it matter too much and I was looking for some minimal solution since this value is going to be stored twice a minute in the worst case.

If this did not suffice, I'd rather store some smaller value derived from the query (like a checksum) than the actual query itself.

app/com/arpnetworking/metrics/portal/alerts/scheduling/AlertExecutionContext.java
cwbriones: This range should be inclusive like it is with the generic query executor. I don't think there's any change necessary if kept inclusive. Subtracting some duration would be essentially treating the range as exclusive, no?

speezepearson: Hrrm, but-- suppose
```
period = MINUTELY
scheduled = 2020-06-24 12:00:00
```
Then
```
latest   = 2020-06-24 12:00:00
previous = 2020-06-24 11:59:00
```
...and then we'll fire a request at Kairos like
```json
{ "...": "..." , "start_absolute": "2020-06-24 11:59:00", "end_absolute": "2020-06-24 12:00:00" }
```
...and then we'll get the datapoints for both 11:59:00 _and_ 12:00:00.
Is that what we want?

cwbriones: Yes, I think that's alright. The only timestamp we're concerned with when interpreting the result is the one belonging to the most recent datapoint. So if we got both points back, I would only look at the one corresponding to `2020-06-24 12:00:00`. It's also why I then check that the difference between scheduled and the last timestamp is within one period, so for example if the only datapoint was at `11:59:00` then it wouldn't be considered firing at this moment since that was >=1 period ago.

cwbriones: All this talk however does make me realize I need a mechanism to account for data delay, because if I'm always looking for a datapoint at the current minute it won't be present due to the 3 minute delay from aggregation and so we'd always report the alert as not firing.

speezepearson: Ahhh, indeed. Do you plan to put that in this diff or fast-follow it?

cwbriones: I put that in this diff. It was only ~80 lines in total...although it did push this diff over 1k 😬 

https://github.com/ArpNetworking/metrics-portal/pull/422/commits/ce8ceb13aa152e774853c4d4fd674bc584ce5c72

PR #426, opened 2020-06-24
rollups: fix mis-logged metric from #417

speezepearson:


app/com/arpnetworking/rollups/ConsistencyChecker.java
```diff
@@ -234,7 +234,7 @@ private void sampleCountsReceived(final SampleCounts sampleCounts) {
             final double nSamplesDropped = nOriginalSamples - nRollupSamples;
 
             metrics.incrementCounter("rollup/consistency_checker/original_samples", (long) nOriginalSamples);
-            metrics.incrementCounter("rollup/consistency_checker/dropped_samples", (long) nOriginalSamples);
+            metrics.incrementCounter("rollup/consistency_checker/dropped_samples", (long) Math.max(0, nOriginalSamples - nRollupSamples));
```
orborde, line 237: `Math.max` seems like it's glossing over the interesting corner case of "how on Earth did we wind up with _more_ rollup samples than original samples?".

i.e. if this can never happen, you don't need `Math.max`, and if it _can_ happen, maybe you wanna, I dunno, log a warning instead of silently ignoring it?

No strong feelings.

PR #427, opened 2020-06-24
make RollupManager a cluster singleton so hash-ring shuffles don't create duplicate work

speezepearson:
Consider yonder graph of the RollupManager queue size, after I turned on rollups for many new metrics, and restarted the MPortal hosts so they would pick up the change:

Screenshot 2020-06-24 12 45 54

The initial huge spike makes sense: we need to backfill rollups for all these metrics I just turned on.
The second spike is a mistake, and I'm _pretty_ sure it's due to a hash-ring shuffle:
- Notice how the green host didn't get any work in the first spike: I suspect it was late to the party for some reason, so when the work was being divided up, the hash-ring had only two nodes.
- Then, an hour later, the RollupGenerators fire off a bunch of work again. Most of the jobs are duplicates of things that have already been submitted to the RollupManagers. But because of the shuffle, there's only a 1/3 chance that the job will hit the same RollupManager that it did last time; so there's a 2/3 chance that it won't be de-duplicated!
- The logs confirm that a lot of duplicate work got done: there were 30586 distinct jobs in this time range, but 54883 executions. Most of duplication is due to the same job being executed on two different hosts.

A weak point in this explanation: the first two spikes makes sense, but why the subsequent spikes? Once the pool had all 3 nodes, no more shuffling should have occurred. I don't know the answer to this, but I'm pretty confident that the hash-ring shuffle is the heart of the issue.

Downsides to this change:
- **Homogeneity.** We lose some homogeneity in the cluster. It's nice to have all your nodes doing exactly the same thing.
- **Remote messages:** There will be more messages sent between actors on different nodes: each RollupExecutor used to talk to its _local_ RollupManager, and now (N-1)/N of them will have to send messages to another node in order to fetch work / report completion.

Why am I willing to accept those downsides?
- **Homogeneity.** We already have 4 different cluster-singleton actors, including one for rollups.
- **Remote messages:** How bad will this be? Well, it can't more than double the number of cross-node messages in the system: the RollupGenerator already sends cross-node messages when it's submitting work to the RollupManager pool, so if the RollupExecutors do that too, well, we go from 1 remote message per job to 2.

    If this turns out to be a problem, I feel strongly that batching messages (Generator->Manager: "here are 10 jobs"; Executor->Manager: "please give me 10 jobs") will be a simpler solution than dealing with a distributed system.


speezepearson:
Hrrm. I'm no longer convinced that we were right about the underlying cause. Here's another occurrence of the same symptoms:

Screenshot 2020-06-24 15 05 45

orborde:
> Hrrm. I'm no longer convinced that we were right about the underlying cause. Here's another occurrence of similar symptoms, but here, it doesn't seem plausible to blame rebalancing:
> 
> Screenshot 2020-06-24 15 05 45

Agreed - this PR is probably an improvement, but I'm skeptical that it will solve the observed weirdness.

BrandonArp:
Is there a particular reason we're not using cluster sharding, but we're essentially trying to do it ourselves?  Am I missing something here?

orborde:
@speezepearson left the company. Given that we're not even sure we correctly identified the problem this PR is trying to solve, I'm tabling it.

orborde:
Actually, let's see if I can just merge it.

orborde:
> Is there a particular reason we're not using cluster sharding, but we're essentially trying to do it ourselves? Am I missing something here?

You're talking about the state of things before this change, right?

PR #430, opened 2020-06-29
Interface for reloading config-based alert definitions

cwbriones:
FileAlertRepository, introduced in #389, allowed for reading hardcoded alert definitions from a JSON file but had no mechanism for reloading them on updates. While this PR does not actually introduce a working live-update implementation, it does break apart the configuration loading into an interface to make this reloading possible.

The existing static file logic has been extracted into its own `ConfigProvider`. A `PluggableAlertRepsitory` configured with a `StaticFileConfigProvider` would be equivalent in functionality to the previous `FileAlertRepository`.

Why the additional abstraction if we're not introducing reloading implementations right away? We'll be using this to integrate with Dropbox's live configuration system. I plan on adding a Filesystem watcher implementation later.


app/com/arpnetworking/play/configuration/ConfigurationHelper.java
```diff
@@ -119,12 +123,38 @@ public static FiniteDuration getFiniteDuration(final Config configuration, final
...
+     * configuration object.
+     *
+     * @param injector The injector to use to create the object.
+     * @param environment Play {@link Environment} instance.
+     * @param configuration The config to instantiate the object from. (Note that this should almost certainly <i>not</i> be
+     *   the entire Play {@link Config} instance; instead it is some sub-object describing the Java object to instantiate.
+     * @param <T> The type of object to instantiate.
+     * @return The instantiated object.
+     */
+    public static <T> T toInstanceMapped(
```
cwbriones, line 143: This may or may not be necessary - I might be misunderstanding how Guice and Jackson interact.

I want to be able to deserialize a class defined outside of this repo from a configuration object. Since there will be no Guice provider available (because as mentioned, it's an external class), I want to be able to just read the constructor arguments from the configuration object as-is.

The most straightforward way I could think of was to turn the config object into JSON and then deserialize it via Jackson. Jackson will then just use the constructor parameter names, and there's no fussing around with any Hocon Config bindings.

e.g. `FileWatchingConfigProvider(final String path)` would be then configured in Hocon as
```
configProvider = {
    path = "/etc/config.json"
}
```

app/com/arpnetworking/metrics/portal/config/ConfigProvider.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @see StaticFileConfigProvider
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public interface ConfigProvider {
+    /**
+     * Start this config provider with the given {@code update} function.
+     * <p>
+     * The frequency and semantics of any updates are left up to the particular implementation.
```
cwbriones, line 35: Will fix: I'll add a comment about how the stream will not be buffered.

Why? I want callers to have control over the buffering.

cwbriones:
Feel free to bikeshed! I’m not too happy about the name being so general but couldn’t think of anything else

app/com/arpnetworking/play/configuration/ConfigurationHelper.java
cwbriones: It looks like we already do this manually elsewhere, like for per-type report configuration

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
```diff
@@ -215,7 +219,9 @@ public void addOrUpdateAlert(final Alert alert, final Organization organization)
                             .build();
             mapBuilder.put(uuid, alert);
         }
-        return mapBuilder.build();
+        synchronized (this) {
+            _alerts = mapBuilder.build();
+        }
```
speezepearson, line 252: Why the `synchronized`? I'd have thought this statement was already atomic:
- `mapBuilder.build()` is pure (and therefore effectively atomic);
- member assignment is atomic (right???);
- therefore `_alerts = mapBuilder.build()` is atomic.

orborde:
Note that I have zero context on this, but, to me, the obvious question is: what prevents you from loading the config once at process startup and _restarting the whole process_ when a new config arrives? Usually this is simpler and less error-prone than plumbing in dynamic-reload logic.

I just timed restarting metrics_portal in prod, and it took a little over 2 seconds. That seems fast enough that you could get away with restarting the process.

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
cwbriones: AFAIK, yes member assignment is atomic.

I actually had omitted the synchronized initially because I had the same thoughts. I suppose at this point it's moreso a matter of thread visibility rather than safety but even then I would need to make the reads themselves synchronized.

I'm inclined to remove it for now. I'm also left wondering if this could cause issues for the pagination of alerts when we are submitting them for execution (what if it changes mid pagination?). Perhaps we just kick the coordinator?

cwbriones:
> Note that I have zero context on this, but, to me, the obvious question is: what prevents you from loading the config once at process startup and restarting the whole process when a new config arrives? Usually this is simpler and less error-prone than plumbing in dynamic-reload logic.

I don't think I'd necessarily agree that it's simpler; you would still have some sort of filesystem watcher or notifier but instead of running within the application it would run outside and control the application's process itself. I'd agree though that it would have the benefit of clearing any internal state to ensure you're not running in some mixed-configuration mode.

General thoughts: Live reloading lets you abstract away the actual source of the data at runtime in a pluggable way instead of tying it to the particular build or platform infrastructure (e.g. you could use Zookeeper instead of a filesystem). It decouples application deployment from configuration deployment. It also prevents you from killing any queries (or really any other work) that are still in-flight.

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
speezepearson: Ahhh. That's a good point. This `synchronized` isn't necessary to stop _other_ code from interfering with _this_ code, but it _does_ stop _this_ code from interfering with _other_ synchronized code.

No strong preferences on this!

PR #432, opened 2020-07-01
default ebean.properties with UTC timezone

### LINKED ISSUE #419: EBeanServer should treat dates it receives as UTC
This might only be a problem with the test `EbeanServer` but I'm opening this regardless to confirm.

In #398 I discovered an issue where the Ebean was using the local time zone to compute the partition date for execution records. The repository was using UTC. This causes an issue where the repo thinks that it's created the partition for the given record, but the insert still fails because the dates don't actually match due to the timezone discrepancy.

[I was able to fix this by manually configuring the server instance used by the integration tests](https://github.com/ArpNetworking/metrics-portal/blob/e8cf91748f383ac803df3cf00655fcd2634204f4/test/java/com/arpnetworking/metrics/portal/integration/test/EbeanServerHelper.java#L115), but this may still be an issue in production. I also am afraid that this swap-over will mess up existing repositories keyed by date, since a key that would have previously mapped to local time will be mapped to UTC after the configuration change. I believe the only existing repo that uses this is ReportExecutionRepository, but I haven't confirmed this just yet.

cwbriones:
Resolves #419.

See the issue for details. The short of it is that we already treat all unzoned times internally as UTC, mostly through the usage of `Instant`. Ebean by default however, uses the JVM timezone when it initializes its connection pool. This previously caused issues in tests, and so it was explicitly set to UTC there. This PR does the same by adding a default properties file.


cwbriones:
My thought process in adding this default is that we're making this usage explicit, and by doing it through standard configuration anyone who's deployed mportal without configuring this can continue to do so by changing the timezone in `ebean.properties`

cwbriones:
Alternatively - since this is properties file I might be mistaken in how easily overridden this is. If that's a problem I think we'll just have to resort to setting it ourselves in our deployment.

cwbriones:
Closing since I think it's another source of configuration that could actually go out of sync (rather than a single source, the JVM's timezone). Additionally the prior issue seems mostly like a test isolation problem rather than one with the system during normal operation.

PR #433, opened 2020-07-08
Add a TTL to rollup series

cwbriones:
I opted for leaving this non-optional since the documented behavior is to use whatever is configured in kairos if you set it to 0.

I suppose this means the only use case that would not be covered would be if you have a default TTL but never want your rollups to have a TTL (but this seems weird so I left as is)

### DIFF
--- a/conf/portal.application.conf
+++ b/conf/portal.application.conf
@@ -132,7 +132,7 @@ hostProvider {
 rollup {
   # TTL on the individual rollup series.
   # Specifying 0 will use the global TTL configured by KairosDB.
-  ttl = 0
+  ttl = 0s
 
   generator.count = 5
   executor.count = 5
--- a/conf/portal.application.conf
+++ b/conf/portal.application.conf
@@ -130,6 +130,10 @@ hostProvider {
 # Rollups
 # ~~~~~
 rollup {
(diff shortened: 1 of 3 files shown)


conf/portal.application.conf
```diff
@@ -130,6 +130,10 @@ hostProvider {
 # Rollups
 # ~~~~~
 rollup {
+  # TTL on the individual rollup series.
+  # Specifying 0 will use the global TTL configured by KairosDB.
```
wittekm, line 134: are we 100% sure of this? If so, accepted.

cwbriones: Based on https://kairosdb.github.io/docs/build/html/restapi/Aggregators.html#save-as

I suppose that could be completely wrong

cwbriones: Did my due-diligence, in the fork we run it defaults to 0 anyway when unset:

https://github.com/InscopeMetrics/kairosdb/blob/master/src/main/java/org/kairosdb/events/DataPointEvent.java#L36

constructed via

https://github.com/InscopeMetrics/kairosdb/blob/master/src/main/java/org/kairosdb/core/aggregator/SaveAsAggregator.java#L172

PR #439, opened 2020-07-29
Create endpoint /v1/alerts/:id

cwbriones:
Part 1/2 for controller implementation.

The paginating query endpoint is what is actually necessary for regular alert evaluation, but this endpoint introduces much of the machinery for translating / serializing alerts so I decided to split it up.

A firing alert will look like this:
```javascript
{
    "id" : "bb7bb881-7383-58d8-8c34-7482738584ae",
    "name" : "MyAlert",
    "description" : "An example alert",
    "enabled" : true,
    "additionalMetadata" : {
        "arbitrary" : 0,
        "things" : {
            "can go" : "here"
        }
    },
    // null or omitted if the alert has never been evaluated
    "firingState" : {
        "lastEvaluatedAt" : "2020-07-29T00:30:01Z",
        // empty array if there were no series firing at the time of `lastEvaluatedTs`
        "firingTags" : [
            {"foo" : "bar"}, 
            {"foo" : "baz"}
        ]
    }
}
```


app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
```diff
@@ -120,6 +121,28 @@ public PluggableAlertRepository(
...
+     *
+     * Instantiates a repo with a fixed set of alerts that will never be
+     * updated.
+     *
+     * @param objectMapper The object mapper to use for alert deserialization.
+     * @param org The organization to group the alerts under.
+     * @param alerts The set of alerts to place into this repository.
+     */
+    @VisibleForTesting
+    public PluggableAlertRepository(
```
cwbriones, line 135: I opted for exposing this constructor because
* PluggableAlertRepository is already an in-memory repository, albeit unmodifiable. Since the controller doesn't have any mutation endpoints right now we don't need those anyway.
* Alternative would be: Create an explicit `MapAlertRepository` that is just a dumber pluggable alert repository. This seemed unnecessary and we can add it once we need mutator methods on the conroller.

test/java/controllers/AlertControllerTest.java
```diff
@@ -0,0 +1,236 @@
...
+            "bf560372-a853-4a9c-948c-9fe9f0269ad7"
+    ).map(UUID::fromString).collect(ImmutableList.toImmutableList());
+
+    private static final List<UUID> NOT_FIRING_IDS = Stream.of(
+            "c6e274d9-d8c6-40e3-aea8-d1146c813c0f",
+            "20adb3d5-4c0a-4b60-a4c1-819a86798bb1",
+            "11dead9e-e5cc-427d-a3a5-381259353b4d",
+            "09d7212f-06b8-4b91-a6f9-384350190cfc",
+            "ce7347cd-9d05-44d1-ac36-f868a14b7c49"
+    ).map(UUID::fromString).collect(ImmutableList.toImmutableList());
```
cwbriones, line 87: This setup has more IDs than necessary because they will be used in the pagination test. I extracted this PR from it and I left this in to spare myself the headache of needing to rebase again later.

cwbriones:
I think I'll need to add the tag group-bys to this view as well but I can do that after the fact since it requires a bit more than just adding it to the view model.

app/com/arpnetworking/metrics/portal/alerts/impl/NullConfigProvider.java
```diff
@@ -0,0 +1,41 @@
...
+    private NullConfigProvider() {}
+
+    @Override
+    public void start(final Consumer<InputStream> update) { }
+
+    @Override
+    public void stop() { }
+
+    public static NullConfigProvider getInstance() {
+        return new NullConfigProvider();
```
vjkoskela, line 39: Do we need this to be instantiable from configuration? I'm not sure that it is via `getInstance`.

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
vjkoskela: Isn't the answer just to create a `MapConfigProvider`? This would take a map, serialize it, and then provide it to the repository. Seems like you did that pattern already with `NullConfigProvider`.

app/models/view/alerts/Alert.java
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+@Loggable
+public final class Alert {
+    private UUID _id;
+    private String _name;
+    private String _description;
+    private boolean _enabled;
+    private ImmutableMap<String, Object> _additionalMetadata;
```
vjkoskela, line 39: Hrm. This contradicts the statement above: `Play view models are mutable.` (the original did too)

app/models/view/alerts/Alert.java
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * View model for an alert.
+ * <p>
+ * Play view models are mutable.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+@Loggable
+public final class Alert {
+    private UUID _id;
```
vjkoskela, line 35: Is the pattern now to type these more explicitly. I believe the API treats these as opaque strings.

app/models/view/alerts/Alert.java
```diff
@@ -0,0 +1,117 @@
...
+    private AlertFiringState _firingState;
+
+    /**
+     * Construct a view model from its internal representation.
+     *
+     * @param internal The internal model for this alert.
+     * @param mostRecentEvaluation The most recent evaluation result for this alert.
+     * @return An alert view model.
+     */
+    public static Alert fromInternal(
```
vjkoskela, line 49: Historically, these methods were not on the view model classes to avoid the server-side rendering code (view code) from having dependencies on types in the internal model.

vjkoskela: (it's been a while since I've been in this codebase, so I'm not certain whether this paradigm changed with reporting/alerting)

app/models/view/alerts/AlertFiringState.java
```diff
@@ -0,0 +1,75 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+import java.time.Instant;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * Nested view model for an alert's firing state.
+ *
```
vjkoskela, line 29: nit: `<p>`

cwbriones:
There's two changes I already noticed I need to make to this
1. I'm not sure why I made firingState non-optional. If the alert is never evaluated, it shouldn't be present.
2. Pretty obvious now but I'm missing the serialized query on the definitions 🤦

test/java/com/arpnetworking/metrics/portal/integration/test/WebServerHelper.java
```diff
@@ -86,27 +89,40 @@ public static String readContentAsString(final CloseableHttpResponse response) t
     }
 
     /**
-     * Return the response content as a {@code byte[]}.
+     * Return the response content as a {@code JsonNode}.
```
vjkoskela, line 92: Your method descriptions are reversed.

cwbriones:
> Pretty obvious now but I'm missing the serialized query on the definitions 🤦

Actually, I just remembered why I left this out. It's because it was not needed by Dropbox's client since metrics portal knows nothing about the particular query language anyway. Still, it probably should be included.

app/models/view/alerts/Alert.java
cwbriones: It may have changed, I did it this way following what was done with reports:
https://github.com/ArpNetworking/metrics-portal/blob/master/app/models/view/reports/Report.java#L118

app/models/view/alerts/Alert.java
cwbriones: I can change this but I didn't see it as a contradiction - it's not _deeply_ mutable, but you can still swap out the map reference here with something else.

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
cwbriones: I didn't do this originally because the serialization code is all within an inner class of PluggableAlertRepository, and it would have to be exposed in order to be reused in this test class. Let me see about doing this.

app/models/view/alerts/Alert.java
cwbriones: This is another case where I went with what reports did.

It's a UUID on the [report model](https://github.com/ArpNetworking/metrics-portal/blob/master/app/models/view/reports/Report.java#L40) and [on the routes](https://github.com/ArpNetworking/metrics-portal/blob/master/conf/portal.routes#L39).

I'm not too opinionated here since IMO the versioning on the endpoint and the decoupling of the view model give us sufficient wiggle room were this ever to change.

app/models/view/alerts/Alert.java
vjkoskela: Not sure if needs changing. Like I said it's everywhere. Maybe add a comment. I don't think it's a big deal at this point.

app/models/view/alerts/Alert.java
vjkoskela: I noticed that afterwards as well. The other routes are string. Pretty sure I just missed this on the reporting stuff. Not a big deal, maybe file an issue.

app/models/view/alerts/Alert.java
vjkoskela: I'm starting to see a pattern here.

test/resources/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepositoryTest.Alerts.json
```diff
@@ -1,173 +1,49 @@
...
-                "threshold" : 10000.0
-              }
-            ]
-          }
-        ]
-      }
+      "name": "FooIsTooHigh",
+      "description": "Foo is too high.",
+      "enabled": true,
+      "query": "{\"metrics\": [{\"tags\": {\"host\": [\"host1\"]}, \"name\": \"foo\", \"limit\": 10000, \"aggregators\": [{\"name\": \"sum\", \"sampling\": {\"value\": 1, \"unit\": \"minutes\"}}, {\"name\": \"filter\", \"filter_op\": \"LT\", \"threshold\": 10000.0}]}]}"
```
cwbriones, line 8: This was an automated change.

Every query field became a string containing the JSON that was originally in that field. The PluggableAlertRepository parsed as JSON since at least with the versioning it seemed OK to assume the structure and we get some nice early validation, but that doesn't play well with serializing from an internal model (since we can't guarantee it's JSON).

Since it's now a string, this will break any existing definition loading. I think this is acceptable because this is not production-ready yet and on our end we only have a single test definition which can easily be changed.

PR #442, opened 2020-07-30
Unbounded periodic schedule

cwbriones:
Introduce an unbounded periodic schedule, motivated by the alerting use case since alerts do not need backfilling behavior.

By analogy I've also renamed the `BaseSchedule` class to `BoundedSchedule` to make it clear that it makes certain assumptions about the subtype behavior. I've also refactored `NeverSchedule` to no longer be bounded since that didn't make much sense.

EDITED: This was originally just to collect feedback but I decided to mark this as a proper PR.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportRepository.java
+++ b/app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportRepository.java
@@ -342,10 +342,9 @@ public final class DatabaseReportRepository implements ReportRepository {
             beanOneOff.setRunUntil(internalOneOff.getRunUntil().orElse(null));
             return beanOneOff;
         } else if (internalSchedule instanceof NeverSchedule) {
-            final NeverSchedule internalNever = (NeverSchedule) internalSchedule;
             final ReportSchedule beanNever = new NeverReportSchedule();
             beanNever.setRunAt(Instant.ofEpochSecond(0));
-            beanNever.setRunUntil(internalNever.getRunUntil().orElse(null));
+            beanNever.setRunUntil(null);
             return beanNever;
         }
         throw new IllegalArgumentException("Unsupported internal model: " + internalSchedule.getClass());
--- a/app/com/arpnetworking/metrics/portal/scheduling/impl/UnboundedPeriodicSchedule.java
+++ b/app/com/arpnetworking/metrics/portal/scheduling/impl/UnboundedPeriodicSchedule.java
@@ -39,49 +39,37 @@ import java.util.Optional;
  * <p>
  * <b>WARNING:</b>
  * This behavior means that jobs with this schedule will ignore previously missed runs
- * - if you need backfills you should instead use a standard {@link PeriodicSchedule}.
+ * - if you need backfills you should instead use a standard {@code PeriodicSchedule}.
  *
+ * @see PeriodicSchedule
  * @author Christian Briones (cbriones at dropbox dot com)
  */
 @Loggable
 public final class UnboundedPeriodicSchedule implements Schedule {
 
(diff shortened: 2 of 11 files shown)


app/com/arpnetworking/metrics/portal/scheduling/impl/UnboundedPeriodicSchedule.java
```diff
@@ -0,0 +1,171 @@
...
+        final Duration fullPeriod = Duration.of(_periodCount, _period);
+
+        final Instant now = _clock.instant();
+        final Instant start = lastRun
+                .filter(run -> run.compareTo(now) >= 0)
+                .map(run -> run.plus(fullPeriod)) // Avoid repeating the same period.
+                .orElse(now);
+
+        final long epochMillis = start.toEpochMilli();
+        final Instant nextRun = Instant.ofEpochMilli((epochMillis / fullPeriod.toMillis()) * fullPeriod.toMillis());
```
vjkoskela, line 70: I did the same thing with: 

D = TS % INT
NEXT = TS - D

app/com/arpnetworking/metrics/portal/scheduling/impl/UnboundedPeriodicSchedule.java
```diff
@@ -0,0 +1,171 @@
...
+        _clock = builder._clock;
+    }
+
+    @Override
+    public Optional<Instant> nextRun(final Optional<Instant> lastRun) {
+        final Duration fullPeriod = Duration.of(_periodCount, _period);
+
+        final Instant now = _clock.instant();
+        final Instant start = lastRun
+                .filter(run -> run.compareTo(now) >= 0)
```
vjkoskela, line 65: This eliminates any runs in the past, right? -- this is to protect us agains the system clock moving. e.g. backwards and repeating work units or forwards and trying to catch-up (vs skip).

My only concern on the backwards is that it would effectively create a clock-time gap when nothing is run. Effectively, I am wondering whether the difference between report and alert scheduling is cpu time vs clock time. Alerts are evaluated every 30 seconds, vs reports are delivered hourly/daily based on the clock.

I'm not 100% convinced this is the difference, although I suspect it is the source of some of these issues.

cwbriones: Correct - it's to protect against the backwards system clock and also the case where the last run was already the most recent period so we can ignore it as completed.

That might be the case. The other thing I'm noticing is that there are two distinct types of schedules - ones like this where all you care about is the job running every period but you don't really care about history, and the other case where you want to ensure that every missed period up to the current time gets evaluated (e.g. reporting backfill).

app/com/arpnetworking/metrics/portal/reports/impl/DatabaseReportRepository.java
```diff
@@ -342,10 +342,9 @@ private ReportSchedule internalModelToBean(final Schedule internalSchedule) {
             beanOneOff.setRunUntil(internalOneOff.getRunUntil().orElse(null));
             return beanOneOff;
         } else if (internalSchedule instanceof NeverSchedule) {
-            final NeverSchedule internalNever = (NeverSchedule) internalSchedule;
             final ReportSchedule beanNever = new NeverReportSchedule();
             beanNever.setRunAt(Instant.ofEpochSecond(0));
-            beanNever.setRunUntil(internalNever.getRunUntil().orElse(null));
+            beanNever.setRunUntil(null);
```
cwbriones, line 347: I made it so `NeverSchedule` is no longer a subclass of `BoundedSchedule` so it doesn't have a `getRunUntil` method anymore. This should have always been null anyway.

PR #446, opened 2020-08-12
Extract out KairosDBService configuration into a provider

cwbriones:
Straightforward refactor that moves out the KairosDB service into its own provider since it should be shared between the KairosDBQueryExecutor and the ProxyController.

I noticed this because the query executor failed with a lack of bindings during an end-to-end alert evaluation test.


test/java/com/arpnetworking/kairos/service/KairosDbServiceImplTest.java
```diff
@@ -78,6 +80,8 @@
     private Metrics _mockMetrics;
     @Mock
     private MetricsQueryConfig _mockQueryConfig;
+    @Mock
+    private Consumer<MetricsQuery> _mockRewrittenMetricsQueryConsumer;
```
cwbriones, line 84: This was added because the controller tests were using this mock, when it should have been in this test suite.

PR #447, opened 2020-08-12
Fix a bug with unbounded schedule fixed points

cwbriones:
Bug-fix for a scheduling issue caught during an end-to-end alert evaluation test.

For times that land *exactly* on a period, we would return that same period as the next evaluation time. This was because the order of the `orElse` and `Duration::plus` operations were inversed, since we always want to perform that `plus` regardless of if there was a previous run or not. The time returned by the schedule should *always* be in the future. I've included a test that caught this bug and updated the other tests since their behavior was actually incorrect.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/scheduling/impl/UnboundedPeriodicSchedule.java
+++ b/app/com/arpnetworking/metrics/portal/scheduling/impl/UnboundedPeriodicSchedule.java
@@ -17,6 +17,7 @@ package com.arpnetworking.metrics.portal.scheduling.impl;
 
 import com.arpnetworking.commons.builder.OvalBuilder;
 import com.arpnetworking.logback.annotations.Loggable;
+import com.arpnetworking.metrics.portal.scheduling.JobExecutorActor;
 import com.arpnetworking.metrics.portal.scheduling.Schedule;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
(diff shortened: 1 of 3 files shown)


app/com/arpnetworking/metrics/portal/scheduling/impl/UnboundedPeriodicSchedule.java
```diff
@@ -58,10 +58,11 @@ private UnboundedPeriodicSchedule(final Builder builder) {
...
     public Optional<Instant> nextRun(final Optional<Instant> lastRun) {
         final Instant now = _clock.instant();
+
         final Instant start = lastRun
-                .filter(run -> run.compareTo(now) >= 0)
-                .map(run -> run.plus(_fullPeriod)) // Avoid repeating the same period.
-                .orElse(now);
+                .filter(run -> run.isAfter(now))
+                .orElse(now)
+                .plus(_fullPeriod); // Avoid repeating the same period
```
vjkoskela, line 65: Won't this cause the opposite behavior now where there is always an initial delay of `_fullPeriod`?

cwbriones: The point here is that when you call `nextRun` it should always be in the future. So I suppose yes, there is a delay.

Here's the before and after, assuming a schedule of every 30 minutes:

Unchanged case, where we're not on a period:
| Field | Value |
| --- | --- |
| Now | 10:01:00 |
| Last Run |  10:00:00 |
| Next Run (before) | 10:30:00 |
| Next Run (after) |  10:30:00 |

Changed case, where we're exactly on a period:
| Field | Value |
| --- | --- |
| Now | 10:00:00 |
| Last Run |  (empty) |
| Next Run (before) | 10:00:00 |
| Next Run (after) |  10:30:00 |

If there were no delay for this, as previously, then the `nextRun` will always round down back to `lastRun` if it were exactly the previous period. e.g. times 10:01, 10:02, 10:03 would all return `nextRun = 10:00:00` unless we add the period.

An alternative would be to add an if-statement to check if we're at a fixed point and intentionally bump it, but I felt that this special-casing made it harder to reason with.

cwbriones: Happy to discuss any problems you might see with this. I had run alert evaluation locally and it displayed the expected behavior of running once per minute, starting from the upcoming minute post-startup.

app/global/MainModule.java
```diff
@@ -228,14 +228,18 @@ protected void configure() {
...
+            final PeriodicMetrics metrics
     ) {
         final FiniteDuration interval = ConfigurationHelper.getFiniteDuration(config, "alerting.execution.defaultInterval");
         final java.time.Duration queryOffset = ConfigurationHelper.getJavaDuration(config, "alerting.execution.queryOffset");
 
         final Schedule defaultAlertSchedule = new UnboundedPeriodicSchedule.Builder()
                 .setPeriod(TimeAdapters.toChronoUnit(interval.unit()))
                 .setPeriodCount(interval.length())
+                .setOverrunReporter(overrunPeriodCount -> {
+                    metrics.recordCounter("jobs/executor/overrunPeriods", overrunPeriodCount);
```
cwbriones, line 241: I couldn't encode anything about the job since this isn't within the executor framework at this point. This provider is for alerts and so I *could* mark the metric as alert-specific but I figure that could be messy since it would couple it to the `alert_job` type at a distance.

Perhaps the better thing to do here is refactor this into `AlertExecutionContext`, where the schedule is constructed per-alert, and somehow plumb this in.

vjkoskela: I would be okay with logging a warning or error, ideally with an exception

PR #448, opened 2020-08-12
missing route for alert queries

cwbriones:


cwbriones:
Of course this wasn't caught because the tests for the controller are unit tests and not integration tests. I'm thinking I could add a dumb functional test that sets up the web server to at least catch the route mismatch, even if the alerts returned will always be empty.

test/java/com/arpnetworking/metrics/portal/integration/controllers/AlertControllerIT.java
```diff
@@ -0,0 +1,70 @@
...
+import static org.hamcrest.Matchers.is;
+
+/**
+ * Integration tests for {@code AlertController}.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class AlertControllerIT {
+    @Test
+    public void testQueryAlertsWorksAndIsEmpty() throws IOException, URISyntaxException {
```
cwbriones, line 43: This test pings the endpoint, and expects:
* 200 response
* Everything to be empty or 0

cwbriones:
>  I'm thinking I could add a dumb functional test

This is done now.

PR #449, opened 2020-08-13
stop polling the schedule every tick within JobExecutorActor

cwbriones:
The executor doesn't play well with the UnboundedExecutionSchedule since it recomputes the nextRun time on every tick. I've changed it to cache the computed next run and only recompute it when the execution has been attempted (either succeeds or fails). This should not have different behavior with the scheduling granularity we currently deal with.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
@@ -303,7 +303,7 @@ public class DailyPartitionCreator extends AbstractActorWithTimers {
         // it still returns a "non-empty" void result and so we can't use the
         // ordinarily more appropriate SqlUpdate type.
         final SqlQuery sql = _ebeanServer.createSqlQuery(
-                "select * from portal.create_daily_partition(?::text, ?::text, ?::date, ?::date)")
+                "select * from create_daily_partition(?::text, ?::text, ?::date, ?::date)")
                 .setParameter(1, schema)
                 .setParameter(2, table)
                 .setParameter(3, startDate)
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
@@ -303,7 +303,7 @@ public class DailyPartitionCreator extends AbstractActorWithTimers {
         // it still returns a "non-empty" void result and so we can't use the
         // ordinarily more appropriate SqlUpdate type.
         final SqlQuery sql = _ebeanServer.createSqlQuery(
-                "select * from create_daily_partition(?::text, ?::text, ?::date, ?::date)")
+                "select * from portal.create_daily_partition(?::text, ?::text, ?::date, ?::date)")
                 .setParameter(1, schema)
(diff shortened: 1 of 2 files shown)


app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -303,7 +303,7 @@ protected void execute(
         // it still returns a "non-empty" void result and so we can't use the
         // ordinarily more appropriate SqlUpdate type.
         final SqlQuery sql = _ebeanServer.createSqlQuery(
-                "select * from create_daily_partition(?::text, ?::text, ?::date, ?::date)")
+                "select * from portal.create_daily_partition(?::text, ?::text, ?::date, ?::date)")
```
cwbriones, line 306: I'm not sure why this passes integration tests, still trying to figure that out. It's likely because the actor does not report any problems in a way the test could catch.

I manually created the partitions for the next 2 weeks to mitigate this issue until this fix is out.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -253,18 +257,22 @@ private void tick(final Tick message) throws ActorNotInitializedException {
...
                         .setMessage("attempted to start executing job, but job no longer exists in repository")
                         .addData("ref", cachedJob.getRef())
-                        .addData("scheduled", nextRun.get())
+                        .addData("scheduled", _nextRun.get())
                         .log();
                 killSelf();
+            } finally {
+                // This run was attempted so we should clear it and recompute
+                // the scheduled time on the next tick.
+                _nextRun = Optional.empty();
```
cwbriones, line 275: This is the only thing I'm unsure about - when to reset the cached value of nextRun. IMO it should either be here or within `JobCompleted` but I've placed it here in case the Job doesn't actually complete successfully. I don't think it should make much of a difference because in the case the job did complete successfully, the actor will have to process that message before any tick it receives.

vjkoskela: There is a slight inconsistency with clearing it. Based on the comments cleared implies `// First tick since clearing next run, refresh.`. Should this path actually reset it instead of clearing it?

Also please verify that `attemptExecuteAndUpdateRepository` is synchronous.

cwbriones: I think you're correct, I think refreshing is the better choice here.

`attemptExecuteAndUpdateRepository` is actually asynchronous, it executes the job and pipes the result back to the actor which then marks the job as completed.

I think if I just refresh here, there shouldn't be any issues. It would then be acting the same as before this diff. I would also be concerned that refreshing instead within JobCompleted could make the executor get stuck since if somehow the job were faulty and `attemptExecuteAndUpdateRepository` failed, it would not be able to advance `nextRun`.

cwbriones:
I'm still trying to reason about what the best thing to do with the cached nextRun is, since the tests broke after I reset it instead of setting it to empty.

cwbriones:
I've moved the daily partition creator changes to https://github.com/ArpNetworking/metrics-portal/pull/450, since the query change here didn't actually fix the problem.

PR #450, opened 2020-08-14
Use JDBC conn directly within DailyPartitionCreator

cwbriones:
Ebean's SQLQuery doesn't appear to be working, and SQLUpdate does not
support SELECT statements (which is necessary to call the function).

The existing test suite didn't catch this issue because it created the
partitions for the current date during migrations, and so the partition
creation would silently fail while the tests would still pass. I've
added a new test which creates a scheduled event in the past, which
catches this oversight.

I've verified this locally by running the container separately and then executing the test suite. The expected partitions exist in the database.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
@@ -31,8 +31,17 @@ import com.arpnetworking.steno.LoggerFactory;
 import com.google.common.collect.Sets;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import io.ebean.EbeanServer;
+import io.ebean.RawSql;
+import io.ebean.RawSqlBuilder;
 import io.ebean.SqlQuery;
+import io.ebean.SqlRow;
+import io.ebean.SqlUpdate;
+import io.ebean.Transaction;
+import io.ebean.config.dbplatform.SqlExceptionTranslator;
 
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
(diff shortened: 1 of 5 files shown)


cwbriones:
While adding that test, I noticed that `AlertEvaluationResult` and `DefaultReportResult` had stale implementations of toString, hashCode, equals, so I re-generated them.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -230,7 +239,7 @@ private void tick() {
 
         LocalDate d = startDate;
         boolean allPartitionsExist = true;
-        while (!d.equals(endDate)) {
+        while (d.compareTo(endDate) <= 0) {
```
cwbriones, line 242: I did this to match the semantics of `create_daily_partition` which includes both endpoints.

test/java/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreatorTest.java
```diff
@@ -57,6 +57,8 @@
     private static final Duration MSG_TIMEOUT = Duration.ofSeconds(1);
     private static final long TEST_LOOKAHEAD = 7;
 
+    private static final Instant CLOCK_START = Instant.parse("2020-08-13T00:00:00Z");
```
cwbriones, line 60: Added so that this test no longer relies on the current time.

test/java/com/arpnetworking/metrics/portal/integration/repositories/JobExecutionRepositoryIT.java
```diff
@@ -174,6 +175,22 @@ public Integer visit(final JobExecution.Started<T> state) {
         }).apply(lastRun.get());
     }
 
+    @Test
+    public void testJobScheduledInThePast() {
```
cwbriones, line 179: This test catches the false positives from using the current date (since a partition will exist for that regardless of if `DailyPartitionCreator` succeeds.

PR #451, opened 2020-09-04
bugfix: add alert id to where clause when inserting an execution

cwbriones:
A silly bug, but it's happened nonetheless.

## Expectation
We want to update an alert execution, so we first check it exists by querying for the most recent execution _for this particular ID_.

## Reality
We want to update an alert execution, so we first check it exists by querying for the most recent execution. 

Notice that the above never involves Alert ID.

---

Fortunately, this blows up because we only expect to ever retrieve a single result, instead of several.
The fix is straightforward: add the missing where clause. I've also included a sanity check test that I used to reproduce and identify the issue.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertExecutionRepository.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertExecutionRepository.java
@@ -115,6 +115,7 @@ public final class DatabaseAlertExecutionRepository implements AlertExecutionRep
                         .where()
                         .eq("organization.uuid", org.get().getUuid())
                         .eq("scheduled", scheduled)
+                        .eq("alert_id", jobId)
                         .findOneOrEmpty()
         );
         final AlertExecution newOrUpdatedExecution = existingExecution.orElseGet(AlertExecution::new);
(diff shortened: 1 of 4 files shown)


app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertExecutionRepository.java
```diff
@@ -115,6 +115,7 @@ private AlertExecution findOrCreateAlertExecution(
                         .where()
                         .eq("organization.uuid", org.get().getUuid())
                         .eq("scheduled", scheduled)
+                        .eq("alert_id", jobId)
```
cwbriones, line 118: This single line fixes the bug.

joeyjackson (approved):
Looks good

cwbriones:
A silly bug, but it's happened nonetheless.

Expectation
We want to update an alert execution, so we first check it exists by querying for the most recent execution for this particular ID.

Reality
We want to update an alert execution, so we first check it exists by querying for the most recent execution.

Notice that the above never involves Alert ID.

Fortunately, this blows up because we only expect to ever retrieve a single result, instead of several.
The fix is straightforward: add the missing where clause. I've also included a sanity check test that I used to reproduce and identify the issue.

PR #452, opened 2020-09-11
default to empty values for version/enabled

cwbriones:
These are not required fields, and on the serialization end they are
dropped when they are considered "zero values" e.g. 0 for long. Let's
just explicitly default to these.

Additionally we log the config error instead of completely failing to start up.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
@@ -214,7 +214,11 @@ public class PluggableAlertRepository implements AlertRepository {
         try {
             group = _objectMapper.readValue(bufferedStream, AlertGroup.class);
         } catch (final IOException e) {
-            throw new RuntimeException("Could not load alerts", e);
+            LOGGER.error()
+                .setMessage("Could not load alert definitions")
+                .setThrowable(e)
+                .log();
+            return;
         }
         final ImmutableMap.Builder<UUID, Alert> mapBuilder = ImmutableMap.builder();
         for (final SerializedAlert fsAlert : group.getAlerts()) {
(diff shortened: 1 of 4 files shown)


app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
```diff
@@ -214,7 +214,11 @@ private void reload(final InputStream stream) {
         try {
             group = _objectMapper.readValue(bufferedStream, AlertGroup.class);
         } catch (final IOException e) {
-            throw new RuntimeException("Could not load alerts", e);
+            LOGGER.error()
```
joeyjackson, line 217: What is the motivation behind this change? Wouldn't you want the failure to be more noticeable if you cant read the alerts?

cwbriones: This is to decrease the granularity of failure. If we push a bad config, we'll still become aware of it through other means e.g. we have an alert on time since most recent evaluation that would fire.

The issue I want to avoid is a bad alert config causing metrics portal to fail to start up entirely (which is what happened in this case). That prevents other unrelated jobs from running such as rollups and proxied metrics requests.

joeyjackson: Sounds good. As long as there's some way that we're notified of a bad config

joeyjackson (approved):

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
vjkoskela: Can we log a metric here too?

PR #453, opened 2020-09-24
run anti-entropy when the pluggable alert repository reloads

cwbriones:
This is intended to solve the following problem:

1. Metrics portal is operating in a clustered state, with N > 1 nodes.
2. The entire fleet is restarted, perhaps during a deployment
3. The JobCoordinator for alerts restarts and then attempts to run anti-entropy to reconcile jobs
4. The anti-entropy job runs on a node that has not yet loaded its alert definitions, and so it sees an empty repository
5. Anti-entropy "succeeds" and there are no alerts until the next tick occurs in one hour.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
@@ -91,7 +91,7 @@ public class PluggableAlertRepository implements AlertRepository {
     private final ObjectMapper _objectMapper;
     private final Organization _organization;
     private final PeriodicMetrics _periodicMetrics;
-    private final Optional<ActorRef> _alertJobCoordinator;
+    private final ActorRef _alertJobCoordinator;
     private final Duration _openTimeout;
     private ImmutableMap<UUID, Alert> _alerts = ImmutableMap.of();
 
@@ -123,7 +123,7 @@ public class PluggableAlertRepository implements AlertRepository {
                 ConfigurationHelper.toInstanceMapped(injector, environment, config.getConfig("configProvider")),
                 UUID.fromString(config.getString("organization")),
                 config.getDuration("openTimeout"),
-                Optional.of(alertJobCoordinator)
+                alertJobCoordinator
         );
     }
 
@@ -143,7 +143,7 @@ public class PluggableAlertRepository implements AlertRepository {
             final ConfigProvider configProvider,
             final UUID org,
             final Duration openTimeout,
-            final Optional<ActorRef> alertJobCoordinator
+            final ActorRef alertJobCoordinator
     ) {
         _objectMapper = objectMapper;
         _configProvider = configProvider;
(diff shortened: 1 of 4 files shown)


app/com/arpnetworking/metrics/portal/scheduling/JobCoordinator.java
```diff
@@ -237,24 +249,18 @@ private static String simpleTypeName(final Class<?> clazz) {
...
     public Receive createReceive() {
         return receiveBuilder()
-                .match(AntiEntropyTick.class, message -> {
+                .matchEquals(ANTI_ENTROPY_TICK, message -> {
 
                     LOGGER.debug()
                             .setMessage("ticking")
                             .addData("repositoryType", _repositoryType)
                             .log();
-                    if (_currentlyExecuting) {
```
cwbriones, line 246: This was never set to `true` and so it already had no effect.

I removed it because I am worried that we will still have the same problem if we ignore ticks. e.g. the first node reloads, causes a tick, but the job then runs anyway on a node that's still empty. Seems unlikely, but it's completely averted if we just run this for each tick that occurs.

If the actual load is an issue from running N anti-entropy ticks at startup, I could alternatively add a "batching period" to this where we delay the actual job from running after receiving the first tick.

BrandonArp:
What would it take to just check if we have 0 alerts, and if so, load the alerts, then kick off the job?

vjkoskela:
I think the solution has two prongs:

* Prevent Incorrect Result - We should not use an unloaded repository to sync with. This seems to defeat the point of anti-entropy. One easy fix to this is to block open on the repository until it has loaded its initial data set. I'm hoping that if anti-entropy tries to query an unopen repo it would fail, and then retry (sooner than 1hour).

* The repository's contents can change whenever. We should not be waiting for the anti-entropy to run to have them take effect. If we initiate the load process (hell even if it's just calling into anti-entropy) on data change in the repository that would have the changes take affect sooner.

vjkoskela:
@BrandonArp why would a zero sized repository be special? What all the data really was deleted?

cwbriones:
The "first prong" that Ville mentioned: https://github.com/ArpNetworking/metrics-portal/pull/454.

This would allow us to distinguish "actually empty" from "I haven't loaded anything yet"

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
```diff
@@ -102,13 +108,16 @@ public PluggableAlertRepository(
             final PeriodicMetrics periodicMetrics,
             final Injector injector,
             final Environment environment,
+            @Named("AlertJobCoordinator")
+            final ActorRef alertJobCoordinator,
```
vjkoskela, line 112: Should this then be marked `@Nullable`? 

Question: Why is it not always provided? Is this because it's cluster singleton and only runs on one node? If so, can you add that to the construct description.

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
```diff
@@ -272,6 +284,17 @@ private void reload(final InputStream stream) {
         _alerts = mapBuilder.build();
         _periodicMetrics.recordCounter(RELOAD_SUCCESS_COUNTER, 1);
         _periodicMetrics.recordGauge(RELOAD_GAUGE, _alerts.size());
+
+        // Immediately kick the coordinator to pick up any job changes so that we
+        // don't have to wait for anti-entropy to begin.
+        //
+        // NOTE: Since nodes have their own copy of this repository, this will
```
vjkoskela, line 291: Wait, so when is the coordinator actor ref not provided?

vjkoskela (approved):
I'm okay with this with minor clarifications.

@BrandonArp what do you think?

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
cwbriones: That was really just for testing. The constructor tied into Guice will always provide the actor ref.

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
cwbriones: as mentioned above, this was for testing.

app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
cwbriones: I made this non-null / non-optional to avoid confusion. It was only empty in one test case anyway.

BrandonArp:
My thought was that a 0-sized repo would be weird. And that a double check against an empty repo doesn't put much load on a system (that wouldn't have much load to begin with, since it's empty). As for loading at startup before things are marked as up is a fine way to solve the problem.  The trade-off is that we're blocking on startup to avoid the race condition (as opposed to trying to recognize the race condition). I'm good with what's implemented now.

BrandonArp (approved):

PR #454, opened 2020-09-24
wait for reload when opening the alert repository

cwbriones:
Now we won't consider a repository open until it receives at least one update from the inner `ConfigProvider`. If that provides an empty stream, we can be confident that the repository is truly empty rather than unloaded.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
@@ -53,10 +53,15 @@ import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.Charset;
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


app/com/arpnetworking/metrics/portal/alerts/impl/PluggableAlertRepository.java
```diff
@@ -119,24 +126,38 @@ public PluggableAlertRepository(
...
+        _configProvider.start(stream -> {
+            reload(stream);
+            if (!initialReload.isDone()) {
+                initialReload.complete(null);
+            }
+        });
+        try {
+            initialReload.get(_openTimeout.toMillis(), TimeUnit.MILLISECONDS);
+        } catch (final InterruptedException | ExecutionException | TimeoutException e) {
+            throw new RuntimeException("failed waiting for initial reload", e);
```
cwbriones, line 159: I'm unsure if I should re-throw here since it will likely prevent startup instead of just breaking alerting.

vjkoskela (approved):

PR #455, opened 2020-09-25
singletons: return a ref to the proxy, not the manager

cwbriones:
Several actor refs handed out in `MainModule` are incorrectly returning the cluster singleton manager, when instead they should be returning a reference to the proxy. 

The proxy **must be used** instead of the manager, otherwise the message won't be routed to the singleton actor. The rollup singletons are correctly implemented and these changes are based on those providers, e.g. ([RollupMetricsDiscoveryProvider](https://github.com/ArpNetworking/metrics-portal/compare/cluster_singleton_proxy?expand=1#diff-cca88e717fceea609a9f29e31e270390R889))

See https://doc.akka.io/docs/akka/2.5/cluster-singleton.html#an-example for details.


vjkoskela (approved):
Good find.

PR #456, opened 2020-09-30
reconstruct JobRef from JobExecutorActor name if possible

cwbriones:
When a `JobExecutorActor` is moved between nodes due to a shard move, it ends up in a sort of purgatory state making no meaningful progress since one or both of the following happen:
1. the state of the actor is not moved, so it doesn't know what job it is responsible for
2. the ticker for the actor isn't restarted, so it doesn't attempt to check if the job needs to run

In order to avoid this entirely, we would need a form of dynamic props to transfer the job ref over when the actor is started post-move.  Unfortunately, classic actor sharding [1] does not allow for this and so we're left with this hack of encoding the ref ID into the actor's name. So now at startup, the actor will attempt to infer its JobRef from its name and start ticking again if it is able to do so. 

[1]: The typed actor implementation of cluster sharding appears to allow for dynamic props, if we ever were to move to this in the future.


app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -334,6 +369,7 @@ private void jobCompleted(final JobCompleted<?> message) {
                 LOGGER.info()
                         .setMessage("marking job as successful")
                         .addData("ref", ref)
+                        .addData("actorName", getSelf().path().name())
```
cwbriones, line 372: TODO before landing: remove this

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -74,23 +74,58 @@
...
+
+        final String actorName = getSelf().path().name();
+        final Optional<JobRef<?>> ref = _refSerializer.entityIDtoJobRef(actorName);
+        if (ref.isPresent()) {
+            final JobRef<T> castRef = unsafeJobRefCast(ref.get());
+            LOGGER.info()
+                    .setMessage("inferred job ref from name, triggering reload")
+                    .addData("jobRef", castRef.toString())
+                    .log();
+            reload(new Reload.Builder<T>().setJobRef(castRef).build());
```
cwbriones, line 122: Unfortunately this adds another reload to the already excessive reload messages at startup because now the actor will send itself a reload message when it starts, if it can.

If there's some way I'm missing to avoid this and distinguish actor start from an actual reload, I'd be open to changing it.

app/com/arpnetworking/metrics/portal/scheduling/JobRefSerializer.java
```diff
@@ -0,0 +1,49 @@
...
+ * <p>
+ * This is used as somewhat of a crutch for the lack of dynamic props in Akka's
+ * classic actor clustering, and would be unnecessary if we later move to typed
+ * actors.
+ * <p>
+ * See <a href=https://stackoverflow.com/a/26524666>this SO post</a> for more details. This is Option A.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public interface JobRefSerializer {
```
cwbriones, line 32: I'm not tied to the name, open to bikeshedding.

app/com/arpnetworking/metrics/portal/scheduling/TwoWayJobRefSerializer.java
```diff
@@ -0,0 +1,122 @@
...
+ * Round-trip is only guaranteed for refs whose repository types are whitelisted by this serializer
+ * (e.g. allowed by the constructor arguments).
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public class TwoWayJobRefSerializer implements JobRefSerializer {
+    private static final Logger LOGGER = LoggerFactory.getLogger(TwoWayJobRefSerializer.class);
+
+    private final Map<String, Class<? extends JobRepository<?>>> _repositories;
+    private final Map<String, Class<? extends JobExecutionRepository<?>>> _execRepositories;
```
cwbriones, line 43: I'm not entirely sure this whitelisting is necessary, but I thought to be safe we should avoid constructing a reference to any class we haven't vetted (configured) ourselves.

BrandonArp:
I believe that the "proper" way to resolve this is to use event sourcing and generating unique and consistent actor names. Basically what you did here, but instead of parsing the info out of the actor path, you'd go to the state store and read a snapshot or event stream that would have the data you need. Not saying that is what you should do, just providing additional context.

vjkoskela:
@BrandonArp the only thing that needs to be resurrected for the actor is its identity (the repo type and job ref id). There's no need to playback the messages or take multiple snapshots. Further, it does not seem possible to just store or serialize the actor state after its initialized (e.g. create a snapshot on startup without subsequent message replay). Maybe the underlying design is not aligned with Akka's design patterns. Genuinely curious how you think this should have been built.

BrandonArp:
I would either go with the parsing the actor path as Christian as done, or I would have event sourcing, but only have a single shapshot for identity and no other persistent event messages. It could go either way. I'm not a big fan of parsing strings, but it's certainly easier.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
vjkoskela: Is it safe to call `reload` in `preStart` vs sending a reload message to itself?

Given this new scheme for job actors, why do we need to send a reload message on actor creation at all? It seems completely redundant. Right?

If we go that route, the warn below likely should be an error and kill the actor (send msg).

app/com/arpnetworking/metrics/portal/scheduling/JobMessageExtractor.java
```diff
@@ -25,18 +26,24 @@
...
         super(NUM_SHARDS);
+        _refSerializer = refSerializer;
     }
 
     @Override
     @Nullable
     public String entityId(final Object message) {
         if (message instanceof JobExecutorActor.Reload) {
-            return jobRefToUId(((JobExecutorActor.Reload) message).getJobRef());
+            return _refSerializer.jobRefToEntityID(((JobExecutorActor.Reload) message).getJobRef());
```
vjkoskela, line 46: Is there any case where a reload would change the ref? I'm starting to think that a reload should be payload-less and just trigger the logic instead of describing what to reload. Effectively the latter changes the actor's identify without changing its name. Seems dangerous.

app/com/arpnetworking/metrics/portal/scheduling/TwoWayJobRefSerializer.java
```diff
@@ -0,0 +1,122 @@
...
+                .addData("repositoryType", ref.getRepositoryType());
+        }
+        final String execRepoName = ref.getExecutionRepositoryType().getSimpleName();
+        if (!_execRepositories.containsKey(execRepoName)) {
+            LOGGER.warn()
+                    .setMessage("The execution repository on this ref is not whitelisted for serialization.")
+                    .addData("repositoryType", ref.getExecutionRepositoryType());
+        }
+        return String.join(
+                "_",
```
vjkoskela, line 82: Technically I believe underscore is allowed in class and package names in Java.

https://www.geeksforgeeks.org/character-isjavaidentifierpart-method-in-java-with-examples/
https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html

app/global/MainModule.java
```diff
@@ -393,6 +396,18 @@ private Clock provideClock() {
...
     }
 
+    @Provides
+    @SuppressFBWarnings("UPM_UNCALLED_PRIVATE_METHOD") // Invoked reflectively by Guice
+    private JobRefSerializer provideJobRefSerializer() {
+        // This must match the configuration of each JobCoordinator, or else
+        // we will not be able to guarantee that job actors will run after
+        // a shard moves within the cluster.
+        return new TwoWayJobRefSerializer(
+                ImmutableList.of(AlertJobRepository.class, ReportRepository.class),
```
vjkoskela, line 406: So this is good except `AlertJobRepository` is a repository wrapper. The serializability of the job reference / repository is determined by the repository actually wrapped by it. 

-- also, should it not come from configuration?

In general, I get where you're going with this and it's good, but I think it might be more than we want to bite off given the class structure of the repositories. Defense against bad configuration is good, but if you're going to setup clustered reporting/alerting and now have a remote repository you're in deep trouble already.

app/com/arpnetworking/metrics/portal/scheduling/JobRefSerializer.java
vjkoskela: I don't think we have an existing pattern for this in general. 

We use Jackson serde classes where needed in the case of JSON. Arguably we could use Jackson's framework here since it supports class to class conversions. I'm not a fan of this.

More generically, MAD has [Parser.java](https://github.com/ArpNetworking/metrics-aggregator-daemon/blob/master/src/main/java/com/arpnetworking/metrics/common/parsers/Parser.java) which fulfills half that contract.

AFAIK there's nothing general in Guava or the JDK.

I'd be up for seeing where this goes with: `com.arpnetworking.commons.serialization.Serializer` and `com.arpnetworking.commons.serialization.Deserializer`. Double parameterized (in & out types) with a single `serialize` and `deserialize` method each. Or more generically, a `Converter` class with a single method (serialization and deserialization are really the same process, just with different names and implementations -- if you assume a non-fixed target type).

app/com/arpnetworking/metrics/portal/scheduling/JobRefSerializer.java
```diff
@@ -0,0 +1,49 @@
...
+    String jobRefToEntityID(JobRef<?> ref);
+
+    /**
+     * Reconstruct a job ref given the entity ID, if possible. This is generally unsafe to use
+     * (due to generics) unless you can guarantee that the input id was created by the same serializer.
+     *
+     * @param id the entity ID (actor name)
+     * @return the job ref.
+     */
+    Optional<JobRef<?>> entityIDtoJobRef(String id);
```
vjkoskela, line 48: I think if it cannot deserialize it should probably throw a checked exception.

vjkoskela: (disambiguates between null/optional serde and failure)

app/com/arpnetworking/metrics/portal/scheduling/TwoWayJobRefSerializer.java
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * A {@link JobRefSerializer} that allows construction to and from entity IDs.
+ *
+ * Round-trip is only guaranteed for refs whose repository types are whitelisted by this serializer
+ * (e.g. allowed by the constructor arguments).
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public class TwoWayJobRefSerializer implements JobRefSerializer {
```
vjkoskela, line 39: I would be fine with something like `JobRefSerializer` or `JobRefConverter` if implementing either `Converter` or both `Serializer` and `Deserializer` (see comment above). Alternatively, you could have separate `JobSerializer` and `JobRefDeserializer` implementations.

app/com/arpnetworking/metrics/portal/scheduling/TwoWayJobRefSerializer.java
vjkoskela: See comment above. I think there are some hidden complexities here that do not have high ROI now.

vjkoskela: I am more concerned about ActorA moving from HostA to HostB which have independent locally cloned repositories and say one is not yet open. Such a race condition would result in actor death.

vjkoskela (changes requested):

cwbriones:
> or I would have event sourcing, but only have a single shapshot for identity and no other persistent event messages. It could go either way. I'm not a big fan of parsing strings, but it's certainly easier.

Yeah, persistence was discussed but seemed redundant given the job state persisted to the database. The only state of the actor is essentially what job it is responsible for evaluating, and that's always determined at startup by the JobRef (which is 1-1 with the entity ID). I would be more inclined to snapshot if we had other state we were maintaining throughout the actor lifetime.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
cwbriones: > Is it safe to call reload in preStart vs sending a reload message to itself?

I believe so, but I can change it to a `tell` just to be consistent.

> Given this new scheme for job actors, why do we need to send a reload message on actor creation at all? It seems completely redundant. Right?

I'm not sure if you're referring to the message from `preStart` or from the `JobCoordinator` when you refer to "send a reload message on actor creation".

We need to send it on pre-start because post-move we need to reinitialize our CachedJob and start the timer again.
As for JobCoordinator, I would agree that both reloading and starting the actor would be superfluous. Really, after these changes, all we care about is actor identity that we infer from the `JobRef`. This means externally we would ideally have two interactions:
* We only want the actor to start, we don't care about reload.
* We *actually* want the JobExecutorActor to reload its state, maybe a database update just happened or anti-entropy is running.

I can't see a way to tease these apart for anti-entropy, however, since from its POV there's no difference between restarting a stopped actor and asking it to reload.

app/com/arpnetworking/metrics/portal/scheduling/JobMessageExtractor.java
cwbriones: Commented on most of this above.

I will agree that this is brittle, it relies on the actor being set up correctly by some external code yet in a way the actor implementation understands. The identity/name correspondence should *always* be correct if used via a ClusterSharding manager but there's nothing in the type system / code to enforce that invariant.

app/com/arpnetworking/metrics/portal/scheduling/TwoWayJobRefSerializer.java
cwbriones: > I think there are some hidden complexities here that do not have high ROI now

If that's the case, I have any qualms with removing this – simpler is better.

> I am more concerned about ActorA moving from HostA to HostB which have independent locally cloned repositories and say one is not yet open

Sure, but I think that concern is orthogonal to restarts in the general sense.

app/global/MainModule.java
cwbriones: Mentioned above, I'm just stripping this out for now since I agree it's not worth the ROI due to the complexity from clustering.

app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
vjkoskela: Yeah, I'm fine with reloads happening in two cases:

* Underlying data change
* Anti-entropy

However, the reload on actor start should be removed.

vjkoskela (approved):

PR #458, opened 2020-10-01
JobExecutor: request stop from parent via passivate 

cwbriones:
Right now the actor just PoisonPills itself which causes it to get
resurrected by the ClusterSharding manager. Instead we should tell the
manager to kill the child permanently via Passivate.

This unintended resurrection has likely been around for a while, but only became
visible recently since post-resurrection actors previously did nothing meaningful (this changed in #456).


vjkoskela (approved):

cwbriones:
Right now the actor just PoisonPills itself which causes it to get
resurrected by the ClusterSharding manager. Instead we should tell the
manager to kill the child permanently via Passivate.

This unintended resurrection has likely been around for a while, but only became
visible recently since post-resurrection actors previously did nothing meaningful (this changed in #456).

PR #467, opened 2020-11-05
Batch fetch alerts from postgres

cwbriones:
When we paginate over alerts, we currently fetch their execution status individually. As you can imagine, this is not very performant as we make a round trip to the DB for every alert.

This adds a batch lookup method which fetches the executions in one go. The query itself is much cheaper per-alert (measured up to 1000 records), so including the decrease in network cost I don't expect latency to be a problem again here for a while.

### DIFF
--- a/app/controllers/AlertController.java
+++ b/app/controllers/AlertController.java
@@ -64,7 +64,7 @@ public class AlertController extends Controller {
     private static final String CONFIG_EXECUTIONS_LOOKBACK_DAYS = "alerts.executions.lookbackDays";
 
     private static final int DEFAULT_EXECUTIONS_BATCH_SIZE = 500;
-    private static final int DEFAULT_EXECUTIONS_LOOKBACK_DAYS = 7;
+    private static final int DEFAULT_EXECUTIONS_LOOKBACK_DAYS = 1;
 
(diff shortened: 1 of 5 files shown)


app/controllers/AlertController.java
```diff
@@ -49,12 +59,26 @@
  */
 public class AlertController extends Controller {
 
+    private static final String CONFIG_LIMIT = "alerts.limit";
+    private static final String CONFIG_EXECUTIONS_BATCH_SIZE = "alerts.executions.batchSize";
+    private static final String CONFIG_EXECUTIONS_LOOKBACK_DAYS = "alerts.executions.lookbackDays";
```
cwbriones, line 64: I can't actually see a case where we'd want to tune this to something other than 1, since we really only care about the most recent executions. I figure this would only make sense if you had stale jobs, but even then you'd probably optimize for getting the job to run and back up to date rather than trying to fetch old results.

vjkoskela (approved):

cwbriones:
Just to mention some discussion here that happened out of band: the reason we don't paginate alerts and the associate execution in one go is because the alerts are in memory (or at least could be, in this model) while the executions are in Postgres.

PR #468, opened 2020-11-09
add a metric for backfill age per period per namespace

cwbriones:
Includes a basic tagger impl for extracting a prefix from the metric name.

### DIFF
--- a/app/com/arpnetworking/rollups/RollupGenerator.java
+++ b/app/com/arpnetworking/rollups/RollupGenerator.java
@@ -28,6 +28,8 @@ import com.arpnetworking.kairos.client.models.MetricTags;
 import com.arpnetworking.kairos.client.models.MetricsQuery;
 import com.arpnetworking.kairos.client.models.MetricsQueryResponse;
 import com.arpnetworking.kairos.client.models.TagsQuery;
+import com.arpnetworking.metrics.Metrics;
+import com.arpnetworking.metrics.MetricsFactory;
 import com.arpnetworking.metrics.incubator.PeriodicMetrics;
 import com.arpnetworking.play.configuration.ConfigurationHelper;
 import com.arpnetworking.steno.Logger;
@@ -41,6 +43,7 @@ import com.typesafe.config.ConfigUtil;
 import scala.concurrent.duration.FiniteDuration;
 
 import java.time.Clock;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.Collections;
 import java.util.List;
@@ -65,7 +68,6 @@ import javax.inject.Named;
  *
  */
 public class RollupGenerator extends AbstractActorWithTimers {
-
     /*
      * The Generator flow is as follows:
      *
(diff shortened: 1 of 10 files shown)


app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -113,12 +116,15 @@ public RollupGenerator(
...
             final Clock clock,
-            final PeriodicMetrics metrics) {
+            final PeriodicMetrics metrics,
+            final MetricsFactory metricsFactory
+    ) {
         _metricsDiscovery = metricsDiscovery;
         _rollupManager = rollupManager;
         _kairosDbClient = kairosDbClient;
         _clock = clock;
         _metrics = metrics;
```
vjkoskela, line 126: Can you rename these to `_periodicMetrics` and `periodicMetrics` respectively to disambiguate from any `Metrics` created from `metricsFactory`?

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -301,6 +307,17 @@ private void handleLastDataPointsMessage(final LastDataPointsMessage message) {
                     period
             );
 
+            if (!startTimes.isEmpty()) {
```
vjkoskela, line 310: **OR** Why isn't this periodic metrics?

app/com/arpnetworking/rollups/RollupGenerator.java
```diff
@@ -483,13 +500,22 @@ private MetricsQuery buildLastDataPointQuery(
                 : lastCompletePeriod;
     }
 
+    private Optional<String> getNamespace(final String metricName) {
```
vjkoskela, line 503: We can't hard-code this here.

We should support one or both of:
* Tag promotion; configurable tag names to promote from context to self-instrumentation
* Metric name prefix; configurable metric name prefix 

These were both implemented in KDB. And the former in MAD and CAGG as well.

vjkoskela (changes requested):

app/com/arpnetworking/rollups/RollupGenerator.java
cwbriones: It wasn't periodic metrics because the tag needs to be different per sample observed, since it's extracted from the metric name.

app/com/arpnetworking/rollups/RollupGenerator.java
cwbriones: I just went off your recent KDB implementation, supporting only metric name (prefix) for now.

app/com/arpnetworking/rollups/RollupGenerator.java
vjkoskela: 👍

vjkoskela (approved):

vjkoskela:
Probably worth testing end to end with ArpNetworking/metrics if you have not already.

PR #470, opened 2020-11-14
Ignore partially aggregated periods in alert queries

cwbriones:
This solves two current issues with alert evaluation.

1. If we have a query period >1min, we should ignore the most recent datapoint in the results because it will only be partially aggregated, which could result in us incorrectly marking the alert as firing.
2. If the query is end-aligned, the lookback window should end at the current time (since this is what end-aligned means).

(2) isn't biting us yet since no queries are end-aligned but it will be an issue that would prevent usage.


app/com/arpnetworking/metrics/portal/query/impl/KairosDbQueryExecutor.java
```diff
@@ -120,12 +122,27 @@ public Duration lookbackPeriod(final MetricsQuery query) {
...
+                .setAlignment(getAlignment(metricsQuery))
+                .build();
+    }
+
+    private QueryAlignment getAlignment(final com.arpnetworking.kairos.client.models.MetricsQuery metricsQuery) {
+        final boolean anyEndAligned = metricsQuery.getMetrics()
+                .stream()
+                .flatMap(m -> m.getAggregators().stream())
+                .flatMap(agg -> Streams.stream(agg.getAlignEndTime()))
+                .anyMatch(endAligned -> endAligned);
```
cwbriones, line 143: If it makes more sense, I can validate the assumption here that either all aggregators are end aligned or all aggregators are period aligned.

This isn't enforced by the data model, but the windowing logic relies on this. A mixed set of alignment would require the query window to start at the same time as period aligned and end at the same time as end-aligned.

vjkoskela (approved):

PR #472, opened 2020-11-30
additional logging on job completion failure

cwbriones:
There are some partition writing errors, and while it's unlikely these are caused by the job itself we don't know this for sure without logging to tie exception <-> jobRef

### DIFF
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
@@ -46,6 +46,7 @@ import java.time.temporal.ChronoUnit;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import javax.persistence.PersistenceException;
 
 /**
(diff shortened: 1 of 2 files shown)


cwbriones:
Not sure why this build is failing, `./jdk-wrapper.sh ./mvnw verify` just passed on my machine

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -205,6 +206,11 @@ private void recordCounter(final String metricName, final long value) {
         _periodicMetrics.recordCounter(fullMetric, value);
     }
 
+    private void recordTimer(final String metricName, final Duration duration) {
+        final String fullMetric = String.format("partition_creator/%s/%s", _table, metricName);
+        _periodicMetrics.recordTimer(fullMetric, duration.toMillis(), Optional.of(TimeUnit.MILLISECONDS));
```
cwbriones, line 211: There are some timeout errors when calling `DailyPartitionCreator.ensurePartitionExists` which is done via `Patterns.ask` to the actor managing the partitions. I added this to see if the operation is really taking longer than a second, or if the actor goes missing on the host.

joeyjackson (approved):

PR #474, opened 2020-12-08
rethrow, don't killSelf()

cwbriones:
This fixes a regression introduced in #472. killSelf will permanently kill the actor, and it will not be resurrected by the supervisor. We should not do this on DB failure, so instead we rethrow to trigger this behavior.

I've also renamed the method to make it clear it will stop the actor forever.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
+++ b/app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
@@ -131,7 +131,7 @@ public final class JobExecutorActor<T> extends AbstractActorWithTimers {
                     .setThrowable(e)
                     .addData("actorName", actorName)
                     .log();
-            killSelf();
+            killSelfPermanently();
         }
     }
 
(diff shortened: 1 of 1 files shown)


app/com/arpnetworking/metrics/portal/scheduling/JobExecutorActor.java
```diff
@@ -413,8 +413,15 @@ private void jobCompleted(final JobCompleted<?> message) {
...
-            return;
+
+            // Rethrow to crash the actor.
+            //
+            // This will indirectly trigger a reload anyway (see JobExecutorActor#preStart).
+            //
+            // We do this instead of reload directly because our supervisor will rate limit
+            // the resurrection of our actor, whereas reloading directly could lead to
+            // a tight retry loop if the DB issue is not transient.
+            throw e;
```
cwbriones, line 424: Log-and-throw is usually an antipattern but I don't want to just act like nothing happened (because of unbounded retry), nor do I want to remove the log which provides context about which job is failing.

joeyjackson (approved):

PR #475, opened 2020-12-21
make the job execution repositories run in a dedicated executor

cwbriones:
The actual repository changes themselves are pretty straightforward - all previously blocking calls are wrapped in a XXXAsync call that spawns the job onto a dedicated executor. The associated tests run the same, only with some `.toCompletableFuture().get()` calls sprinkled in.

The hairier changes involved actually _using_ these futures. In particular, 
* The `JobExecutorActor` had to be reworked while the CachedJob type was removed entirely since the abstraction at this point seems unnecessary.
* The alert controller now has an explicit HttpExecutionContext argument. Play holds request state in a thread local, so if a CompletionStage is executed on a different thread it loses access to this state and the application crashes.


cwbriones:
Still seeing about breaking this up, it's a bit difficult since there's not a simple way without introducing a whole set of blocking interfaces for the repositories.

app/com/arpnetworking/metrics/portal/alerts/impl/DailyPartitionCreator.java
```diff
@@ -319,6 +316,8 @@ protected void execute(
         // while SQLUpdate does not allow for SELECT statements.
         //
         // https://ebean.io/docs/intro/queries/jdbc-query
+        //
+        // TODO(cbriones): Move DB operation off the dispatcher thread pool.
```
cwbriones, line 320: I didn't do this right away to keep this diff from growing any larger. I'm not really concerned by this operation anyway, since it's extremely infrequent.

BrandonArp (approved):
Feels a little dirty, but the code looks solid.

cwbriones:
> Feels a little dirty, but the code looks solid.

Thanks for the review!

It's definitely a sledgehammer of a PR, but besides the JobExecutor overhaul it's more or less the recommended approach by the [akka docs](https://doc.akka.io/docs/akka-http/current/handling-blocking-operations-in-akka-http-routes.html).

PR #477, opened 2020-12-24
Fix a metric being logged in JobExecutorActor

cwbriones:
Now that _cachedJob is an Optional, we're missing a call to Optional#get().


cwbriones:
This isn't causing any runtime errors, it's just aggregating all the job types together which makes this metric useless as is.

BrandonArp (approved):

PR #478, opened 2021-01-08
make repositories return the execution after writing it

cwbriones:
This will be used to cache the result once it's returned.


joeyjackson (approved):
lgtm

PR #479, opened 2021-01-08
allow for deserializing types from JSON while also injecting via Guice

cwbriones:
The config helper has trouble right now if you have nested types, because it loses all context about how to derive the target class name once you've entered one "layer" of object.

e.g. before this change this config would cause problems
```
{ type = "com.example.Foo.class", inner = { type = "com.example.Bar.class" } }
```

`ConfigHelper#toInstanceMapped` would not know to use the `type` key for `Bar`. This new change just places this on the interface with JsonTypeInfo annotations like we do elsewhere. I've also pulled in `jackson-module-guice` so that we can use Guice to inject values alongside those normally passed in as builder values.


app/global/MainModule.java
```diff
@@ -502,31 +524,6 @@ private ActorRef provideStatusActor(
         return system.actorOf(StatusActor.props(cluster, clusterStatusCache), "status");
     }
 
-    @Provides
-    @SuppressFBWarnings("UPM_UNCALLED_PRIVATE_METHOD") // Invoked reflectively by Guice
-    private DatabaseAlertExecutionRepository provideDatabaseAlertExecutionRepository(
```
cwbriones, line 507: No longer needed since Jackson just uses the Builder.

app/global/MainModule.java
```diff
@@ -956,38 +953,22 @@ public ActorRef get() {
         private final boolean _enabled;
     }
 
-    private static final class RollupMetricsDiscoveryProvider implements Provider<ActorRef> {
+    private static final class RollupMetricsDiscoveryProvider extends ClusterSingletonProvider {
```
cwbriones, line 956: Refactored this because this pattern is pretty common in MainModule.

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertExecutionRepository.java
```diff
@@ -360,4 +380,97 @@ private void assertIsOpen(final boolean expectedState) {
...
+        }
+
+        public Integer getLookahead() {
+            return _lookahead;
+        }
+
+        static final class Builder extends OvalBuilder<PartitionConfig> {
+            @NotNull
+            private Integer _lookahead = 7;
+            private scala.concurrent.duration.Duration _offset = FiniteDuration.apply(0, TimeUnit.SECONDS);
```
vjkoskela, line 448: Should this be `@NotNull` as well?

conf/postgresql.application.conf
```diff
@@ -97,7 +97,7 @@ hostRepository.type = "com.arpnetworking.metrics.portal.hosts.impl.DatabaseHostR
 # Alerts
 # ~~~~~
 alertExecutionRepository.type = "com.arpnetworking.metrics.portal.alerts.impl.DatabaseAlertExecutionRepository"
-alertExecutionRepository.partitionManager {
+alertExecutionRepository.partitionConfig {
```
vjkoskela, line 100: Do you need to reflect this change in the ArpNetworking/metrics project?

vjkoskela (approved):

app/com/arpnetworking/metrics/portal/alerts/impl/DatabaseAlertExecutionRepository.java
cwbriones: Yes, will add.

cwbriones:
[Uploading out.log…](Build output)

cwbriones:
[Build output: out.log](https://github.com/ArpNetworking/metrics-portal/files/5789894/out.log)

PR #480, opened 2021-01-11
Migrate from CI to GitHub Actions

cwbriones:


.github/workflows/build.yml
```diff
@@ -0,0 +1,41 @@
+name: build
+on:
+  pull_request:
+  push:
+    # branches:
```
vjkoskela, line 5: I assume this is temporary?

.github/workflows/build.yml
```diff
@@ -0,0 +1,41 @@
...
+          path: ~/.jdk
+          key: ${{ runner.os }}-jdk-${{ hashFiles('.jdkw', 'jdk-wrapper.sh') }}
+      - name: Cache Dependencies (Maven + SBT)
+        uses: actions/cache@0781355a23dac32fd3bac414512f4b903437991a
+        with:
+          path: |
+            ~/.ivy2
+            ~/.m2
+            ~/.sbt
+          key: ${{ runner.os }}-m2-sbt-ivy-${{ hashFiles('pom.xml') }}
```
vjkoskela, line 33: I assume this should probably include at least `build.sbt` as well; maybe other SBT definition files as well that would impact caching.

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+          path: ~/.jdk
+          key: ${{ runner.os }}-jdk-${{ hashFiles('.jdkw', 'jdk-wrapper.sh') }}
+      - name: Cache Dependencies (Maven + SBT)
+        uses: actions/cache@0781355a23dac32fd3bac414512f4b903437991a
+        with:
+          path: |
+            ~/.ivy2
+            ~/.m2
+            ~/.sbt
+          key: ${{ runner.os }}-m2-sbt-ivy-${{ hashFiles('pom.xml') }}
```
vjkoskela, line 28: Same here.

vjkoskela (changes requested):

.github/workflows/build.yml
cwbriones: Yup. I added it in "temporarily enable branch builds" 😉 

I'll just remove it now though, having a test PR is just as straightforward. Since both the branch / tag filters are negated it makes more sense to just remove `push` entirely.

vjkoskela (approved):

cwbriones:
Missed the typo before I merged, oh well.

PR #482, opened 2021-01-12
Add a write-through cache for alert executions

cwbriones:
The cache at the moment is simple, it's just an actor-guarded LRU cache (guava).

An alert execution page lookup will now first attempt to look for all ids in the cache before falling back to the inner
repository (e.g. Postgres). Any cache misses that fell back are written to the cache. Any new successes are written directly to the cache.

This does *not* do any of the following
 - Cache anti-entropy. For the time being the regular evaluation of alerts already serve this function since they happen so frequently.
 - Cache distribution. For a much larger set of alerts, it _could_ make more sense to shard members of the cache and distribute them across the network by using ClusterSharding. It's unclear if this is actually better for most cases since this would require fan-out/fan-in.
 
 I originally attempted to do this using generics to leave this open as a future building block, but backtracked in favor of a specialized implementation. Dealing with generics in Akka messages was not worth it.


app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
```diff
@@ -0,0 +1,259 @@
...
+ * {@link AlertExecutionRepository#getLastSuccessBatch} is probably the most frequently
+ * accessed piece of execution data for alerts, so it's useful to cache those results
+ * separately.
+ *
+ * @author Christian Briones (cbriones at dropbox dot com)
+ */
+public final class CachingAlertExecutionRepository implements AlertExecutionRepository {
+    private final AlertExecutionRepository _inner;
+    private final ActorRef _successCache;
+    private final Duration _cacheTimeout;
```
vjkoskela, line 61: This variable is a bit misleading; it's the cache ttl, it's the cache operation timeout.

app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    @Override
+    public CompletionStage<ImmutableMap<UUID, JobExecution.Success<AlertEvaluationResult>>> getLastSuccessBatch(
+        final List<UUID> jobIds,
+        final Organization organization,
+        final LocalDate maxLookback
+    ) throws NoSuchElementException {
+        // Attempt to get all ids from cache
+        final List<CompletionStage<Optional<JobExecution.Success<AlertEvaluationResult>>>> futures =
```
vjkoskela, line 114: This is expensive. Especially, if it's all remote messages. Should the cache actor support a multi-get?

app/com/arpnetworking/metrics/portal/alerts/impl/CacheActor.java
```diff
@@ -0,0 +1,184 @@
...
+     * @param ref The CacheActor ref.
+     * @param key The key to update.
+     * @param value The associated value.
+     * @param timeout The operation timeout.
+     * @param <K> The type of key.
+     * @param <V> The type of value.
+     * @return A completion stage to await for the write to complete.
+     */
+    public static <K, V> CompletionStage<Void> put(final ActorRef ref, final K key, final V value, final Duration timeout) {
+        return Patterns.askWithReplyTo(
```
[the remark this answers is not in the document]
cwbriones: It would not be, this API is just for interacting with a single actor. that is, I wouldn't expect to be able to pass the proxy actor ref here if we were doing more complex things like replicating the cache.

I don't need any of that here though so maybe I can just rename the class and/or remove that suggestion in the docstring?

app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
cwbriones: Straightforward so I added it.

app/com/arpnetworking/metrics/portal/alerts/impl/CacheActor.java
```diff
@@ -0,0 +1,184 @@
...
+     *
+     * @param ref The CacheActor ref.
+     * @param key The key to retrieve.
+     * @param timeout The operation timeout.
+     * @param <K> The type of key.
+     * @param <V> The type of value.
+     * @return A CompletionStage which contains the value, if any.
+     */
+    @SuppressWarnings("unchecked")
+    public static <K, V> CompletionStage<Optional<V>> get(final ActorRef ref, final K key, final Duration timeout) {
```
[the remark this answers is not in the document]
cwbriones: I might remove the generics for now entirely, it's a whole other can of worms because the messages themselves have serialization issues with being generic.

app/com/arpnetworking/metrics/portal/alerts/impl/CacheActor.java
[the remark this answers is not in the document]
vjkoskela: Yeah I'm fine with that.

A generic cache abstracting its deployment would have been really sweet. However, it's out of scope.

app/com/arpnetworking/commons/akka/JacksonSerializer.java
```diff
@@ -0,0 +1,101 @@
...
+public class JacksonSerializer extends JSerializer {
+    private static final Logger LOGGER = LoggerFactory.getLogger(JacksonSerializer.class);
+    private static @Nullable ObjectMapper gObjectMapper;
+
+    /**
+     * Constructor used by Akka upon system initialization.
+     *
+     * @param ignored the actor system
+     */
+    public JacksonSerializer(final ExtendedActorSystem ignored) {}
```
vjkoskela, line 47: Is there no way to inject the `ObjectMapper` instance into the serializer? The current mutable global static seems problematic.

app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
```diff
@@ -0,0 +1,243 @@
...
+        });
+    }
+
+    private void writeBatchToCache(
+        final Map<UUID, JobExecution.Success<AlertEvaluationResult>> batch,
+        final Organization organization
+    ) {
+        batch.values()
+            .forEach(e -> {
+//                final String key = cacheKey(e.getJobId(), organization.getId());
```
vjkoskela, line 156: ?

app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
```diff
@@ -0,0 +1,243 @@
...
+                AlertExecutionCacheActor.put(_successCache, organization, e, _cacheOperationTimeout);
```
vjkoskela, line 157: I'm confused, isn't a batch of N items still N messages to the actor?

app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
```diff
@@ -0,0 +1,243 @@
...
+        final Instant scheduled
+    ) {
+        return _inner.jobStarted(jobId, organization, scheduled);
+    }
+
+    @Override
+    public CompletionStage<JobExecution.Success<AlertEvaluationResult>> jobSucceeded(
+        final UUID jobId, final Organization organization, final Instant scheduled, final AlertEvaluationResult result
+    ) {
+//        final String key = cacheKey(jobId, organization.getId());
```
vjkoskela, line 181: ?

app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
```diff
@@ -0,0 +1,243 @@
...
+         * @param inner The inner repository
+         * @return This instance of {@code Builder} for chaining.
+         */
+        public Builder setInner(final AlertExecutionRepository inner) {
+            _inner = inner;
+            return this;
+        }
+
+        /**
+         * Sets the cache timeout. Optional. Default is 5 seconds.
```
vjkoskela, line 222: This is the cache operation timeout?

app/com/arpnetworking/metrics/portal/alerts/impl/AlertExecutionCacheActor.java
```diff
@@ -0,0 +1,475 @@
...
+                sender().tell(value, getSelf());
+            })
+            .match(CacheMultiGet.class, msg -> {
+                final ImmutableMap.Builder<UUID, SuccessfulAlertExecution> results = new ImmutableMap.Builder<>();
+                int hits = 0;
+                for (final UUID jobId : msg.getJobIds()) {
+                    final CacheKey cacheKey = new CacheKey(msg.getOrganizationId(), jobId);
+                    final Optional<JobExecution.Success<AlertEvaluationResult>> value = Optional.ofNullable(_cache.getIfPresent(cacheKey));
+                    if (value.isPresent()) {
+                        hits += 1;
```
vjkoskela, line 205: You won't get a hit rate if you publish a value >1

app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
cwbriones: Removed.

app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
cwbriones: Operation timeout. I'll rename.

app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
cwbriones: You're right, I batched the get but not the fallback put. Addressed.

app/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepository.java
cwbriones: Removed.

vjkoskela (approved):

app/com/arpnetworking/commons/akka/JacksonSerializer.java
cwbriones: While I agree and don't like it either, I've been unable to find any other method of injection. The Akka docs suggest that the only way is to add a custom section into the `.conf` file and load it from that, but I don't want to duplicate the ObjectMapper setup.

[This stack overflow user](https://stackoverflow.com/questions/57050458/akka-add-custom-serializer-programmatically) also tried and failed to find a way to do serializer registration at runtime.

cwbriones:
Screen Shot 2021-01-15 at 21 16 35
How did it come to this

PR #484, opened 2021-01-20
wrap the responses so jackson understands the type parameters

cwbriones:
This was clearly incorrect, Jackson needs to either have an enclosing type with concrete parameters or use a type token in order to deserialize the ImmutableMap.

I'm not sure how this got past the test which [forces serialization during testing](https://github.com/cwbriones/metrics-portal/blob/79f368c4aef1640d74251591aee4f8c9c91d716d/test/java/com/arpnetworking/metrics/portal/alerts/impl/CachingAlertExecutionRepositoryTest.java#L77). I did the same during local testing but it seems that the only way to guarantee it's actual serializing the messages (since the config didn't work) is to actually run a local cluster.

### DIFF
--- a/app/com/arpnetworking/metrics/portal/alerts/impl/AlertExecutionCacheActor.java
+++ b/app/com/arpnetworking/metrics/portal/alerts/impl/AlertExecutionCacheActor.java
@@ -93,7 +93,6 @@ public final class AlertExecutionCacheActor extends AbstractActor {
      * @param timeout The operation timeout.
      * @return A CompletionStage which contains the execution, if any.
      */
-    @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


cwbriones:
> I'm not sure how this got past the test which forces serialization during testing

Oddly enough when I add breakpoints to the deserializer, it's only called on messages outside the actor. None of the responses trigger the breakpoints (and the responses are what were incorrect).

vjkoskela (approved):

cwbriones:
> Oddly enough when I add breakpoints to the deserializer, it's only called on messages outside the actor. None of the responses trigger the breakpoints (and the responses are what were incorrect).

The reason was that the map wasn't registered to the Jackson serializer, since it wasn't a custom type with the correct interface. It hits the breakpoint and is serialized now that there's a wrapper.

PR #490, opened 2021-02-16
job actor should be killed/restarted when a future fails

cwbriones:
When a future inside a Patterns.pipe completes exceptionally, we don't reset any of the internal actor state so it continues to think it's still executing. We should just restart the actor entirely to get back to a known good state.


cwbriones:
needed context for this PR: Patterns.pipe sends a `Status.Failure` when the future it's awaiting completes exceptionally.

vjkoskela (approved):

PR #492, opened 2021-03-13
jenkins build

BrandonArp:


BrandonArp:
Finally got the build to pass. The key was the trailing . in the dns names so that it wouldn't look in search domains.

vjkoskela (approved):
I assume that the changes to the pom.xml still build locally since they are not Jenkins specific.

BrandonArp:
Yes, they still do build locally.

PR #501, opened 2021-04-28
update typescript sbt plugin to 0.3.6

cwbriones:


cwbriones:
I'd like to update to latest if possible, but that requires updating scala which I suspect is non-trivial.

BrandonArp (approved):

BrandonArp:
@cwbriones  We had some issues with CI and for some reason it's not picking up the other branch. Let me know when you want this merged.

cwbriones:
@BrandonArp This ended up being unrelated to the release issues we were having. However it's probably good to still get this in, so feel free to merge whenever.

PR #502, opened 2021-05-03
remove github actions workflows and migrate build to jenkins

cwbriones:
This is blocking releases at the moment since they're not working in GHA.


cwbriones:
completely missed #492 🤦🏼‍♂️

PR #504, opened 2021-05-07
Do not reload a job on completion.

vjkoskela:


BrandonArp:
Import errors

test/java/com/arpnetworking/metrics/portal/scheduling/JobExecutorActorTest.java
```diff
@@ -236,43 +236,6 @@ public void testActorRequestsStopWhenJobNoLongerExists() {
         _probe.expectTerminated(ref);
     }
 
-    @Test
-    public void testActorRequestsStopWhenJobIsDeletedBetweenRuns() {
```
cwbriones, line 240: I get why you had to remove this (since this PR prevents hitting the DB) but will this still be handled by anti-entropy?

vjkoskela: Yes, anti-entropy should force a reload of the job as well, and so should explicit job deletion via an API. 

However, the anti-entropy runs only periodically, so a job may continue executing past its deletion time until that runs. That seems expected to me. I don't believe there are any tests for this; however, I did confirm (by code inspection) that the API for reports sends the reload and so does anti-entropy.

cwbriones (approved):

PR #506, opened 2021-06-09
only promote queries to rollups when baseMetric matches

cwbriones:
This fixes a bug where you have two metric names with the same prefix,
one with a rollup.

We would incorrectly promote e.g. `my_metric` to `my_metric_suffixed_1h`.


test/java/com/arpnetworking/kairos/service/KairosDbServiceImplTest.java
```diff
@@ -402,6 +402,24 @@ public void testRollupQueryRewritingBasicRewrite() {
...
+                        .build()
+        );
+        final MetricsQuery original = simpleMetricsQuery("my_metric", aggregators);
+        final MetricsQuery rewritten = KairosDbServiceImpl.useAvailableRollups(
+                ImmutableList.of("my_metric_with_suffix_1h"),
+                original,
+                s -> ImmutableSet.of(SamplingUnit.HOURS),
+                new NoOpMetrics()
+        );
+        assertEquals(original, rewritten);
```
cwbriones, line 420: Test fails if you comment out the change.

joeyjackson (approved):

cwbriones:
@BrandonArp or @vjkoskela 

Any idea what I can do about `continuous-integration/jenkins/pr-head` being stuck at "waiting for status to be reported" ?

BrandonArp:
@cwbriones I think you're right in that it didn't run the build. My guess is that it's due to it being on an external (to the org) fork. Would you mind helping me troubleshoot this tomorrow?

